PR #3, opened 2025-11-06
Image manager

sjmiller609:
See the README in lib/images/README.md for a summary


cmd/api/config/config.go
```diff
@@ -24,14 +23,13 @@ func Load() *Config {
 	_ = godotenv.Load()
 
 	cfg := &Config{
-		Port:             getEnv("PORT", "8080"),
-		DataDir:          getEnv("DATA_DIR", "/var/lib/hypeman"),
-		BridgeName:       getEnv("BRIDGE_NAME", "vmbr0"),
-		SubnetCIDR:       getEnv("SUBNET_CIDR", "192.168.100.0/24"),
-		SubnetGateway:    getEnv("SUBNET_GATEWAY", "192.168.100.1"),
-		ContainerdSocket: getEnv("CONTAINERD_SOCKET", "/run/containerd/containerd.sock"),
```
sjmiller609, line 32: no dependency on containerd or docker

lib/images/disk.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	if err := f.Truncate(diskSizeBytes); err != nil {
+		f.Close()
+		return 0, fmt.Errorf("truncate disk file: %w", err)
+	}
+	f.Close()
+
+	// Format as ext4 with rootfs contents using mkfs.ext4
+	// This works without root when creating filesystem in a regular file
+	cmd := exec.Command("mkfs.ext4", "-d", rootfsDir, "-F", diskPath)
```
sjmiller609: Explored using go-fdisk instead of shelling out, but was getting too complicated. May revisit that

lib/images/manager.go
```diff
@@ -16,43 +21,109 @@ type Manager interface {
...
+
+	images := make([]oapi.Image, 0, len(metas))
+	for _, meta := range metas {
+		images = append(images, *meta.toOAPI())
+	}
+
+	return images, nil
 }
 
 func (m *manager) CreateImage(ctx context.Context, req oapi.CreateImageRequest) (*oapi.Image, error) {
```
sjmiller609: todo before merge: this whole process is too slow (e.g. image download) to happen within a sync HTTP call, I need to change the API to have a 201 accepted / check for progress pattern instead.

cmd/api/config/config.go
```diff
@@ -24,14 +25,14 @@ func Load() *Config {
...
-		JwtSecret:        getEnv("JWT_SECRET", ""),
-		DNSServer:        getEnv("DNS_SERVER", "1.1.1.1"),
+		Port:                getEnv("PORT", "8080"),
+		DataDir:             getEnv("DATA_DIR", "/var/lib/hypeman"),
+		BridgeName:          getEnv("BRIDGE_NAME", "vmbr0"),
+		SubnetCIDR:          getEnv("SUBNET_CIDR", "192.168.100.0/24"),
+		SubnetGateway:       getEnv("SUBNET_GATEWAY", "192.168.100.1"),
+		JwtSecret:           getEnv("JWT_SECRET", ""),
+		DNSServer:           getEnv("DNS_SERVER", "1.1.1.1"),
+		MaxConcurrentBuilds: getEnvInt("MAX_CONCURRENT_BUILDS", 1),
```
sjmiller609, line 35: Concurrency setting for how many images to process at once. Requests are queued

lib/images/errors.go
```diff
@@ -1,9 +1,20 @@
...
+	"strings"
+)
 
 var (
 	ErrNotFound      = errors.New("image not found")
 	ErrAlreadyExists = errors.New("image already exists")
 )
 
+// IsInvalidNameError checks if an error is due to invalid image name
+func IsInvalidNameError(err error) bool {
```
sjmiller609: This seems weird, consider errors before merge

lib/images/README.md
```diff
@@ -0,0 +1,95 @@
...
+- `-d` flag copies directory contents into filesystem
+
+**Options:**
+- `-b 4096` - 4KB blocks (standard, matches VM page size)
+- `-O ^has_journal` - No journal (disks mounted read-only in VMs, saves ~32MB)
+- Minimum 10MB size covers ext4 metadata (~5MB for superblock, inodes, bitmaps)
+
+**Alternative tried:** go-diskfs pure Go ext4, got too complicated. Could revisit this.
+
+**Tradeoff:** Shell command vs pure Go, but mkfs.ext4 is widely available and robust
```
rgarcia: Unikraft uses / recommends erofs since from what they say it's good for read-only rootfs's and cold start performance. Might be easy enough to add an option to `CreateImageRequest ` for filesystem type and add a method like `convertToExt4` that uses `mkfs.erofs`

lib/images/errors.go
rgarcia: plugged it into gpt-5 max thinking and they had this to suggest:

I'll search the hypeman codebase for usages of `IsInvalidNameError`, similar string checks, and how image errors are handled to propose a better pattern.

[5 tools called]

I see it’s only used to classify errors in `cmd/api/api/images.go`. Instead of exposing a string-matching helper, switch to a sentinel error with proper wrapping so callers can use `errors.Is`.

Here’s a drop-in alternative:

- Define a sentinel `ErrInvalidName` and wrap the underlying cause with a typed error that implements `Is` and `Unwrap`.
- Replace string-based checks with `errors.Is(err, images.ErrInvalidName)`.
- Wrap at the source (in `manager.go`) instead of stringifying.

Proposed edits

- hypeman/lib/images/errors.go

```go
package images

import (
	"errors"
)

var (
	ErrNotFound      = errors.New("image not found")
	ErrAlreadyExists = errors.New("image already exists")
	// ErrInvalidName indicates the provided image name is invalid.
	ErrInvalidName   = errors.New("invalid image name")
)

// invalidNameError wraps the underlying cause for invalid image names.
type invalidNameError struct {
	err error
}

func (e *invalidNameError) Error() string {
	return "invalid image name: " + e.err.Error()
}

// Unwrap exposes the underlying parser/validation error.
func (e *invalidNameError) Unwrap() error { return e.err }

// Is enables errors.Is(err, ErrInvalidName) to match.
func (e *invalidNameError) Is(target error) bool { return target == ErrInvalidName }

// invalidName constructs an invalidNameError while keeping the cause.
func invalidName(cause error) error {
	if cause == nil {
		return ErrInvalidName
	}
	return &invalidNameError{err: cause}
}
```

- hypeman/lib/images/manager.go

```go
// CreateImage
normalizedName, err := normalizeImageName(req.Name)
if err != nil {
	return nil, invalidName(err)
}

// GetImage
normalizedName, err := normalizeImageName(name)
if err != nil {
	return nil, invalidName(err)
}

// DeleteImage
normalizedName, err := normalizeImageName(name)
if err != nil {
	return invalidName(err)
}
```

- hypeman/cmd/api/api/images.go

```go
// CreateImage
switch {
case errors.Is(err, images.ErrInvalidName):
	// ...

// GetImage
switch {
case errors.Is(err, images.ErrInvalidName), errors.Is(err, images.ErrNotFound):
	// ...

// DeleteImage
switch {
case errors.Is(err, images.ErrInvalidName), errors.Is(err, images.ErrNotFound):
	// ...
}
```

Why this is better

- Consistent with the rest of the package which already uses `errors.Is` for `ErrNotFound` and `ErrAlreadyExists`.
- Avoids brittle string matching; keeps the original cause available via `Unwrap`.
- Limits surface area: no need to export a special predicate.

- Replaces `images.IsInvalidNameError(err)` with `errors.Is(err, images.ErrInvalidName)` in `cmd/api/api/images.go`.
- Wraps invalid name errors at the source in `lib/images/manager.go` using a proper typed error.
- Adds `ErrInvalidName` and an internal wrapper error in `lib/images/errors.go`; removes the string-matching helper and its `strings` import.

lib/images/manager.go
```diff
@@ -3,56 +3,227 @@ package images
...
+	StatusConverting = "converting"
+	StatusReady      = "ready"
+	StatusFailed     = "failed"
+)
+
 // Manager handles image lifecycle operations
 type Manager interface {
 	ListImages(ctx context.Context) ([]oapi.Image, error)
 	CreateImage(ctx context.Context, req oapi.CreateImageRequest) (*oapi.Image, error)
 	GetImage(ctx context.Context, id string) (*oapi.Image, error)
```
rgarcia: on second read i'm noticing that the lib/images interface is coupled to the oapi generated types... i'm not a huge fan of this although it is definitely easy / convenient. There are some advantages to decoupling and having lib/images define its own types:

- Reduces ripple effects when the API schema changes (versioning, renames, optional/required changes)
- Keeps the package independent of JSON/transport concerns

I'd consider introducing domain types in lib/images and converting at the edge (in cmd/api/... or a small lib/images/oapiadapter.go that depends on oapi).

lib/images/queue.go
```diff
@@ -0,0 +1,99 @@
...
+		q.active[imageName] = true
+		go startFn()
+		return 0
+	}
+
+	q.pending = append(q.pending, build)
+	return len(q.pending)
+}
+
+func (q *BuildQueue) MarkComplete(imageName string) {
```
rgarcia, line 70: Trusting callers to remember to `MarkComplete` might be error-prone. Since `startFn` is the synchronous worker body, could the queue can wrap it and call `MarkComplete` after it returns?

lib/images/queue.go
```diff
@@ -0,0 +1,99 @@
...
+func (q *BuildQueue) Enqueue(imageName string, req oapi.CreateImageRequest, startFn func()) int {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	build := QueuedBuild{
+		ImageName: imageName,
+		Request:   req,
+		StartFn:   startFn,
+	}
+
```
rgarcia, line 59: I'd probably add queue-level dedupe to guarantee no duplicate name starts or gets queued, regardless of caller behavior

e.g. there might be a race with two concurrent `CreateImage` requests getting queued before the first `writeMetadata` completes

lib/images/oci.go
```diff
@@ -0,0 +1,232 @@
...
+	if err := os.MkdirAll(cacheDir, 0755); err != nil {
+		return nil, fmt.Errorf("create cache dir: %w", err)
+	}
+	return &OCIClient{cacheDir: cacheDir}, nil
+}
+
+// Close closes the OCI client (no-op for now)
+func (c *OCIClient) Close() error {
+	return nil
+}
```
rgarcia: do you plan to use OCIClient directly outside `lib/images`? It's slightly awkward to have an exported type and constructor but just one `Close` method exported

lib/images/oci.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (c *OCIClient) pullAndExport(ctx context.Context, imageRef, exportDir string) (*containerMetadata, error) {
```
rgarcia: should this method call `normalizeImageName(imageRef)`?

wondering if we should define a type for normalized refs so that it's clear when we're dealing with normalized vs. not. Chat came up with something that looks pretty good: https://gist.github.com/rgarcia/7f83c25ad6a5074107441b5c7743d010

lib/images/storage.go
```diff
@@ -0,0 +1,189 @@
...
+	lastColon := strings.LastIndex(name, ":")
+	if lastColon == -1 {
+		// No tag, use "latest"
+		return filepath.Join(name, "latest")
+	}
+	
+	imagePath := name[:lastColon]
+	tag := name[lastColon+1:]
+	return filepath.Join(imagePath, tag)
+}
```
rgarcia, line 60: should this method use `github.com/distribution/reference` so it’s robust to tags/digests and consistent with normalization done elsewhere?

rgarcia (approved):
very cool stuff!

lib/images/README.md
sjmiller609: Good idea, I think erofs is just better for this use case

lib/images/manager.go
sjmiller609: Yeah good point, thanks

lib/images/storage.go
sjmiller609: Yeah great point. I think both we should type the normalized refs and storage should be done under digests with symlinks for tags, and also the code needs to deal with tags versus digests (e.g. create again for latest after build, still need old build and new build but tag updated)

cmd/api/api/images.go
```diff
@@ -3,38 +3,53 @@ package api
...
 		switch {
-		case errors.Is(err, images.ErrAlreadyExists):
+		case errors.Is(err, images.ErrInvalidName):
 			return oapi.CreateImage400JSONResponse{
-				Code:    "already_exists",
-				Message: "image already exists",
+				Code:    "invalid_name",
+				Message: err.Error(),
+			}, nil
+		case strings.Contains(err.Error(), "404") || strings.Contains(err.Error(), "not found"):
```
rgarcia: could this be `errors.Is(err, images.ErrNotFound)`?

rgarcia: reading more closely maybe this is registry 404s bubbling up. Could consider mapping registry 404 to images.ErrNotFound in the OCI client i.e. detect go-containerregistry transport errors in inspectManifest and wrap as images.ErrNotFound. Then the API can errors.Is(err, images.ErrNotFound).

cmd/api/api/images.go
```diff
@@ -93,3 +106,29 @@ func (s *ApiService) DeleteImage(ctx context.Context, request oapi.DeleteImageRe
...
+	}
+	if len(img.Env) > 0 {
+		oapiImg.Env = &img.Env
+	}
+	if img.WorkingDir != "" {
+		oapiImg.WorkingDir = &img.WorkingDir
+	}
+
+	return oapiImg
+}
```
rgarcia: configur ur editur 

(to add newlines to the ends of files :))

rgarcia (approved):

PR #5, opened 2025-11-11
Cloud Hypervisor client

sjmiller609:
Thin Go wrapper around Cloud Hypervisor's HTTP API with embedded binaries.

Supports multiple versions of the Cloud Hypervisor VMM because instances are version-locked to cloud hypervisor if they are in standby. We can switch them to the latest version if we shutdown / reboot the VM.

Embeds the binaries to make it easier to install, instead of managing multiple versions of the Cloud Hypervisor CLI externally + configuring it.

See more in lib/vmm/README.md

Also set up a self-hosted github runner with KVM installed


.github/workflows/test.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, x64, kvm]
```
sjmiller609, line 8: This is the same server I'm using for dev server

rgarcia: nice. We should do something like this for the kernel-images-private unikernel build eventually

lib/vmm/README.md
```diff
@@ -0,0 +1,138 @@
+# Cloud Hypervisor VMM Client
+
+Thin Go wrapper around Cloud Hypervisor's HTTP API with embedded binaries.
+
+Supports multiple versions of the Cloud Hypervisor VMM because instances are version-locked to cloud hypervisor if they are in standby. We can switch them to the latest version if we shutdown / reboot the VM.
+
+Embeds the binaries to make it easier to install, instead of managing multiple versions of the Cloud Hypervisor CLI externally + configuring it.
```
rgarcia, line 7: I like this approach!

lib/vmm/client.go
```diff
@@ -0,0 +1,98 @@
...
+
+	cmd := exec.CommandContext(ctx, binaryPath, "--api-socket", socketPath)
+
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("start cloud-hypervisor: %w", err)
+	}
+
+	// Wait for socket to be ready
+	return waitForSocket(ctx, socketPath, 5*time.Second)
+}
```
rgarcia, line 68: not sure how i feel about a method that starts a process and doesn't have a codepath that kills it... but I guess we're treating the `socketPath` as the true side-effect that users of the library are responsible for, which makes sense

rgarcia (approved):

PR #7, opened 2025-11-12
Network manager

sjmiller609:


lib/instances/delete.go
```diff
@@ -64,13 +84,24 @@ func (m *manager) killVMM(ctx context.Context, inst *Instance) error {
...
+			}
 			
-			// Wait for process to die (SIGKILL is guaranteed, usually instant)
-			if !WaitForProcessExit(pid, 1*time.Second) {
-				log.WarnContext(ctx, "VMM process did not exit in time", "id", inst.Id, "pid", pid)
-			} else {
-				log.DebugContext(ctx, "VMM process killed successfully", "id", inst.Id, "pid", pid)
+			// Wait for process to die and reap it to prevent zombies
+			// SIGKILL should be instant, but give it a moment
+			for i := 0; i < 50; i++ { // 50 * 100ms = 5 seconds
```
sjmiller609, line 93: investigate this

cmd/api/api/instances.go
```diff
@@ -282,6 +289,30 @@ func instanceToOAPI(inst instances.Instance) oapi.Instance {
...
+	} else {
+		networkDisabled := false
+		network = &struct {
+			Enabled *bool   `json:"enabled,omitempty"`
+			Name    *string `json:"name,omitempty"`
+		}{
+			Enabled: &networkDisabled,
+		}
+	}
+
```
rgarcia, line 324:
```go
	// Build network object
	network := &struct {
		Enabled *bool   `json:"enabled,omitempty"`
		Name    *string `json:"name,omitempty"`
	}{
		Enabled: lo.ToPtr(false),
	}
	if inst.NetworkEnabled {
		network.Enabled = lo.ToPtr(true)
		network.Name = lo.ToPtr("default")
	}
```

README.md
```diff
@@ -20,6 +27,37 @@ sudo usermod -aG kvm $USER
...
+getcap ./bin/hypeman
+```
+
+**Note:** The `i` (inheritable) flag allows child processes spawned by hypeman (like `ip` and `iptables` commands) to inherit capabilities via the ambient capability set.
+
+**Note:** These capabilities must be reapplied after each rebuild. For production deployments, set capabilities on the installed binary. For local testing, this is handled automatically in `make test`.
+
 ### Configuration
 
 #### Environment variables
```
rgarcia, line 63:
````suggestion
#### Environment variables

Hypeman can be configured using the following environment variables:

| Variable | Description | Default |
|----------|-------------|---------|
| `PORT` | HTTP server port | `8080` |
| `DATA_DIR` | Directory for storing VM images, volumes, and other data | `/var/lib/hypeman` |
| `BRIDGE_NAME` | Name of the network bridge for VM networking | `vmbr0` |
| `SUBNET_CIDR` | CIDR notation for the VM network subnet | `192.168.0.0/16` |
| `SUBNET_GATEWAY` | Gateway IP address for the VM network | `192.168.0.1` |
| `UPLINK_INTERFACE` | Host network interface to use for VM internet access | `eth0` |
| `JWT_SECRET` | Secret key for JWT authentication (required for production) | _(empty)_ |
| `DNS_SERVER` | DNS server IP address for VMs | `1.1.1.1` |
| `MAX_CONCURRENT_BUILDS` | Maximum number of concurrent image builds | `1` |
| `MAX_OVERLAY_SIZE` | Maximum size for overlay filesystem | `100GB` |

**Important: Subnet Configuration**

The default subnet `192.168.0.0/16` covers all addresses from `192.168.0.0` to `192.168.255.255`. This may conflict with your existing network if:
- Your local router uses `192.168.x.x` addresses (common for home networks)
- You're running Docker (often uses `172.17.0.0/16`)
- You have VPNs or other virtual networks configured

**Check for conflicts:**
```bash
# View your current network routes
ip route show

# Check what IP your machine has
ip addr show
```

**Recommended alternative subnets to avoid conflicts:**
- `10.100.0.0/16` - Private range, less commonly used
- `172.30.0.0/16` - Private range between common Docker (172.17.x.x) and AWS (172.31.x.x) ranges
- `192.168.250.0/24` - Smaller range in the high end of 192.168.x.x space

**Example configuration for a conflict-free setup:**
```bash
# In your .env file
SUBNET_CIDR=10.100.0.0/16
SUBNET_GATEWAY=10.100.0.1
```

**Note:** The gateway IP must be within the subnet range you specify. For example, if you use `10.100.0.0/16`, your gateway should be something like `10.100.0.1`.

**Finding the uplink interface (`UPLINK_INTERFACE`)**

`UPLINK_INTERFACE` tells Hypeman which host interface to use for routing VM traffic to the outside world (for iptables MASQUERADE rules). On many hosts this is `eth0`, but laptops and more complex setups often use Wi‑Fi or other names.

**Quick way to discover it:**
```bash
# Ask the kernel which interface is used to reach the internet
ip route get 1.1.1.1
```
Look for the `dev` field in the output, for example:
```text
1.1.1.1 via 192.168.12.1 dev wlp2s0 src 192.168.12.98
```
In this case, `wlp2s0` is the uplink interface, so you would set:
```bash
UPLINK_INTERFACE=wlp2s0
```

You can also inspect all routes:
```bash
ip route show
```
Pick the interface used by the default route (usually the line starting with `default`). Avoid using local bridges like `docker0`, `br-...`, `virbr0`, or `vmbr0` as the uplink; those are typically internal virtual networks, not your actual internet-facing interface.

**Setup:**
````

File: lib/instances/create.go [hunk 4]
```diff
@@ -116,33 +124,65 @@ func (m *manager) createInstance(
...
+	var netConfig *network.NetworkConfig
+	if networkName != "" {
+		log.DebugContext(ctx, "allocating network", "id", id, "network", networkName)
+		netConfig, err = m.networkManager.AllocateNetwork(ctx, network.AllocateRequest{
+			InstanceID:   id,
+			InstanceName: req.Name,
+		})
+		if err != nil {
+			log.ErrorContext(ctx, "failed to allocate network", "id", id, "network", networkName, "error", err)
+			m.deleteInstanceData(id) // Cleanup
```
rgarcia: consider using https://pkg.go.dev/gvisor.dev/gvisor/pkg/cleanup 

we use it in the API for times like this where we need to clean up a bunch of stuff on errors

lib/instances/network_test.go
```diff
@@ -0,0 +1,349 @@
...
+// TestCreateInstanceWithNetwork tests instance creation with network allocation
+func TestCreateInstanceWithNetwork(t *testing.T) {
+	// Require KVM access
+	requireKVMAccess(t)
+
+	manager, tmpDir := setupTestManager(t)
+	ctx := context.Background()
+
+	// Setup image manager for pulling alpine
+	imageManager, err := images.NewManager(paths.New(tmpDir), 1)
```
rgarcia: why not use the image manager attached to `manager`?

lib/instances/manager_test.go
```diff
@@ -22,12 +24,21 @@ import (
 func setupTestManager(t *testing.T) (*manager, string) {
 	tmpDir := t.TempDir()
 	
-	imageManager, err := images.NewManager(paths.New(tmpDir), 1)
+	cfg := &config.Config{
+		DataDir:       tmpDir,
+		BridgeName:    "vmbr0",
+		SubnetCIDR:    "192.168.100.0/24",
+		SubnetGateway: "192.168.100.1",
```
rgarcia: we may want to configure these in a way that's dynamic to the host setup.... or maybe when Initialize()ing the network we detect misconfigurations / conflicts

Makefile
```diff
@@ -96,17 +96,44 @@ dev: $(AIR)
...
+		done; \
+	else \
+		for test in $(BIN_DIR)/tests/*.test; do \
+			if [ -f "$$test" ]; then \
+				echo ""; \
+				echo "Running $$(basename $$test)..."; \
+				$$test -test.v -test.parallel=10 -test.timeout=60s || exit 1; \
+			fi; \
+		done; \
+	fi
```
rgarcia, line 153: does `make dev` / `.air.toml` need adjustment to set capabilities?

lib/instances/network_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+	"github.com/onkernel/hypeman/lib/images"
+	"github.com/onkernel/hypeman/lib/paths"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/vishvananda/netlink"
+)
+
+// TestCreateInstanceWithNetwork tests instance creation with network allocation
+func TestCreateInstanceWithNetwork(t *testing.T) {
```
rgarcia, line 20: FWIW this test fails for me locally and i think it's at the same point you mentioned in person (no internet access on restore from standby)

lib/network/README.md
```diff
@@ -0,0 +1,231 @@
...
+    snapshot-latest/
+      config.json      # Cloud Hypervisor's config with IP/MAC/TAP
+```
+
+### Hybrid Network Model
+
+**Standby → Restore: Network Fixed**
+- TAP device deleted on standby (VMM shutdown)
+- Snapshot `config.json` preserves IP/MAC/TAP names
+- Restore recreates TAP with same name
```
rgarcia, line 85: will be curious what the performance of this is, e.g. should we keep the tap device around to optimize standby -> restore

lib/network/bridge.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	// Get uplink interface (usually eth0, but could be different)
+	// For now, we'll use a common default
+	uplink := "eth0"
+	if m.config.UplinkInterface != "" {
+		uplink = m.config.UplinkInterface
+	}
+
+	// Add MASQUERADE rule if not exists
```
rgarcia:
```suggestion
	// Allow bridge-initiated flows to traverse the host uplink while conntrack tracks new and ongoing sessions
```

lib/network/bridge.go
```diff
@@ -0,0 +1,271 @@
...
+			"-s", subnet, "-o", uplink, "-j", "MASQUERADE")
+		addCmd.SysProcAttr = &syscall.SysProcAttr{
+			AmbientCaps: []uintptr{unix.CAP_NET_ADMIN},
+		}
+		if err := addCmd.Run(); err != nil {
+			return fmt.Errorf("add masquerade rule: %w", err)
+		}
+	}
+
+	// Add FORWARD rules for outbound connections
```
rgarcia:
```suggestion
	// Allow bridge-initiated flows to traverse the host uplink while conntrack tracks new and ongoing sessions
```

lib/network/bridge.go
```diff
@@ -0,0 +1,271 @@
...
+			"-j", "ACCEPT")
+		addForwardOut.SysProcAttr = &syscall.SysProcAttr{
+			AmbientCaps: []uintptr{unix.CAP_NET_ADMIN},
+		}
+		if err := addForwardOut.Run(); err != nil {
+			return fmt.Errorf("add forward outbound rule: %w", err)
+		}
+	}
+
+	// Add FORWARD rules for inbound responses
```
rgarcia:
```suggestion
	// Permit return traffic from the uplink back to the bridge only when conntrack marks it established or related
```

lib/network/bridge.go
```diff
@@ -0,0 +1,271 @@
...
+	bridge, err := netlink.LinkByName(bridgeName)
+	if err != nil {
+		return fmt.Errorf("get bridge: %w", err)
+	}
+
+	if err := netlink.LinkSetMaster(tapLink, bridge); err != nil {
+		return fmt.Errorf("attach TAP to bridge: %w", err)
+	}
+
+	// 5. Set isolation mode (requires kernel support and capabilities)
```
rgarcia:
```suggestion
	// 5. Enable port isolation so isolated TAPs can't directly talk to each other (requires kernel support and capabilities)

```

lib/instances/create.go
sjmiller609: excellent

lib/instances/delete.go
sjmiller609: follow up later when optimizing

lib/network/README.md
sjmiller609: good question. Let's do another pass for optimization

lib/instances/network_test.go
sjmiller609: fixed

.gitignore
```diff
@@ -20,3 +20,4 @@ lib/vmm/binaries/cloud-hypervisor/*/*/cloud-hypervisor
 cloud-hypervisor
 cloud-hypervisor/**
 lib/system/exec_agent/exec-agent
+.cursor/**
```
rgarcia: i think the idea is to commit this directory so everyone can benefit from the goodies within

lib/paths/paths.go
```diff
@@ -163,8 +163,22 @@ func (p *Paths) InstanceSnapshotLatest(id string) string {
...
 	return filepath.Join(p.dataDir, "guests")
 }
 
+// Network path methods
+
+// NetworkDir returns the network state directory.
+func (p *Paths) NetworkDir() string {
+	return filepath.Join(p.dataDir, "network")
+}
+
```
rgarcia: add to comment on top of file?

README.md
```diff
@@ -54,29 +163,32 @@ make build
...
+
 ```bash
 make test
 ```
 
+The test command compiles test binaries, grants capabilities via `sudo setcap`, then runs tests as the current user (not root). You may be prompted for your sudo password during the capability grant step.
+
+**Cleanup stale resources** (if tests were killed with Ctrl+C):
+```bash
+./scripts/cleanup-test-networks.sh
```
rgarcia: don't see this committed but sitting at 100+ network interfaces on my thinkpad so could sure use it ha

openapi.yaml
```diff
@@ -172,6 +181,28 @@ components:
...
+        ip:
+          type: string
+          description: Assigned IP address (null if no network)
+          example: "192.168.100.10"
+          nullable: true
+        mac:
+          type: string
+          description: Assigned MAC address (null if no network)
+          example: "02:00:00:ab:cd:ef"
+          nullable: true
```
rgarcia: maybe nest these in `network`?

lib/network/manager.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Manager defines the interface for network management
+type Manager interface {
+	// Lifecycle
+	Initialize(ctx context.Context, runningInstanceIDs []string) error
+
+	// Instance network operations (called by instance manager)
+	AllocateNetwork(ctx context.Context, req AllocateRequest) (*NetworkConfig, error)
+	RecreateNetwork(ctx context.Context, instanceID string) error
+	ReleaseNetwork(ctx context.Context, alloc *Allocation) error
```
rgarcia: one thing that's slightly confusing is the term "network" vs. "allocation"

wondering if these methods here could be CreateAllocation / RecreateAllocation / ReleaseAllocation?

i'm coming off of just reading types.go where Network was the host-level network setup and allocation is the VM specific part of that

lib/network/allocate.go
```diff
@@ -0,0 +1,279 @@
...
+	"strings"
+	"time"
+
+	"github.com/onkernel/hypeman/lib/logger"
+)
+
+func init() {
+	// Seed RNG with current timestamp for unique random IPs each run
+	mathrand.Seed(time.Now().UnixNano())
+}
```
rgarcia: seeing this warning

```
mathrand.Seed is deprecated: As of Go 1.20 there is no reason to call Seed with
a random value. Programs that call Seed with a known value to get
a specific sequence of results should use New(NewSource(seed)) to
obtain a local random generator.deprecateddefault
```

lib/instances/manager_test.go
```diff
@@ -22,12 +24,22 @@ import (
...
 	
-	imageManager, err := images.NewManager(paths.New(tmpDir), 1)
+	cfg := &config.Config{
+		DataDir:       tmpDir,
+		BridgeName:    "vmbr0",
+		SubnetCIDR:    "192.168.0.0/16",
+		SubnetGateway: "192.168.0.1",
+		DNSServer:     "1.1.1.1",
+	}
+
```
rgarcia, line 33: my thinkpad on the union hall network is on `192.168.12.x`... another vote for `10.100.0.0/16` or making this smart about picking a test subnet

rgarcia (approved):
awesome stuff!

README.md
sjmiller609: you'll have to manually clean up once. I changed the server to clean up automatically on start but also renamed the interfaces to "hype-" to make sure we don't clear out unrelated taps

lib/instances/manager_test.go
sjmiller609: Yeah good call

PR #9, opened 2025-11-17
Remote command execution in microVM instances via vsock

sjmiller609:
See the README in lib/exec/README.md for an overall review.


cmd/exec/main.go
```diff
@@ -0,0 +1,350 @@
+package main
```
sjmiller609, line 1: this is the poc cli

lib/exec/README.md
```diff
@@ -0,0 +1,199 @@
...
+Container (chroot /overlay/newroot)
+```
+
+## How It Works
+
+### 1. API Layer (`cmd/api/api/exec.go`)
+
+- WebSocket endpoint: `GET /instances/{id}/exec`
+- **Note**: Uses GET method because WebSocket connections MUST be initiated with GET per RFC 6455 (the WebSocket specification). Even though this is semantically a command execution (which would normally be POST), the WebSocket upgrade handshake requires GET.
+- Upgrades HTTP to WebSocket for bidirectional streaming
```
rgarcia, line 27: FWIW I have a change to the kernel-images repo in PR to extend our process creation API to add the ability to start a process with a tty, and the way I did it there was extending the existing POST request and adding an attach POST request that is hijacked on the server to get full bidirectional TCP (chatgpt said this is how docker exec does it):
- https://github.com/onkernel/kernel-images/pull/96/files#diff-ecc83eae4e0b06e935228c9d3080e8387051435e6b2062c8f9e1b58c6d382dc0R1436-R1449
- https://github.com/onkernel/kernel-images/pull/96/files#diff-9ed81b55c372cb0fdf7b8ce33205e98c8e69e10815c637820487f69a6c7fe555R595

might be applicable / worth considering something like that here to make the request to start a process explicitly typed w/ a json body

i don't feel strongly about this! I think the grpc approach actually gives us some nice optionality to add other functionality to the guest agent down the road

cmd/api/api/exec.go
```diff
@@ -0,0 +1,202 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/gorilla/websocket"
```
rgarcia, line 14: we have switched to https://github.com/coder/websocket in the API since it seems to be a little bit more modern (supports ctx) and more actively maintained

lib/system/versions.go
```diff
@@ -5,50 +5,27 @@ import "runtime"
...
 )
 
 // KernelDownloadURLs maps kernel versions and architectures to download URLs
 var KernelDownloadURLs = map[KernelVersion]map[string]string{
-	KernelCH_6_12_8_20250613: {
-		"x86_64":  "https://github.com/cloud-hypervisor/linux/releases/download/ch-release-v6.12.8-20250613/vmlinux-x86_64",
-		"aarch64": "https://github.com/cloud-hypervisor/linux/releases/download/ch-release-v6.12.8-20250613/Image-aarch64",
+	Kernel_202511182: {
+		"x86_64":  "https://github.com/onkernel/linux/releases/download/ch-6.12.8-kernel-1-202511182/vmlinux-x86_64",
+		"aarch64": "https://github.com/onkernel/linux/releases/download/ch-6.12.8-kernel-1-202511182/Image-arm64",
```
rgarcia, line 28: sick

rgarcia (approved):

cmd/api/api/exec.go
sjmiller609: ok will look into that if maybe a good follow up

lib/system/versions.go
sjmiller609: meta

PR #10, opened 2025-11-18
Exec works, needs cleanup

sjmiller609:
Opened this pr to my other pr to sort out what to keep and what to delete, got messy while fixing exec command.


cmd/api/api/exec_test.go
```diff
@@ -91,6 +103,73 @@ func TestExecInstanceNonTTY(t *testing.T) {
 	require.NotEmpty(t, actualInst.VsockSocket, "vsock socket path should be set")
 	t.Logf("vsock CID: %d, socket: %s", actualInst.VsockCID, actualInst.VsockSocket)
 
+	// Print console log for debugging
```
sjmiller609: cleanup

cmd/build-dev-initrd/main.go
```diff
@@ -0,0 +1,213 @@
+package main
```
sjmiller609: not needed, but also reconsider how locally testing should work when need to change what's inside initrd

cmd/debug-shell/main.go
```diff
@@ -0,0 +1,87 @@
+package main
```
sjmiller609: not needed

cmd/test-handshake/main.go
```diff
@@ -0,0 +1,25 @@
+package main
```
sjmiller609: not needed

lib/system/versions.go
```diff
@@ -9,45 +9,55 @@ type KernelVersion string
...
 // v2.0.0: Uses pre-built Alpine image with exec-agent from Docker Hub (multi-arch OCI manifest list)
+// v2.0.1: Uses same base but we will inject local agent
+// v2.0.2: Uses same base but with interactive shell fallback
+// v2.0.2-dev: Local dev build (built via cmd/build-dev-initrd)
 var InitrdBaseImages = map[InitrdVersion]string{
-	InitrdV2_0_0: "docker.io/onkernel/hypeman-initrd:1d4efc9-oci",
+	InitrdV2_0_0:                "docker.io/onkernel/hypeman-initrd:d0e84c2-oci",
+	InitrdV2_0_1:                "docker.io/onkernel/hypeman-initrd:d0e84c2-oci",
+	InitrdV2_0_2:                "docker.io/onkernel/hypeman-initrd:d0e84c2-oci",
+	InitrdVersion("v2.0.2-dev"): "docker.io/onkernel/hypeman-initrd:d0e84c2-oci", // Not used, already built locally
```
sjmiller609: wondering, do we need to version initrd or not? maybe not if included in memory snapshot in case of restore, and in case of reboot would be OK to update? I think probably this could be simplified, check on this

scripts/repro_vm.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
sjmiller609: not needed

PR #14, opened 2025-11-26
Log streaming

sjmiller609:
https://github.com/user-attachments/assets/40cb30af-9888-4113-bb50-d830ae9728f9

will need update on CLI to use the generated SDK and handle SSE protocol (not showing "data:" and spurious newlines)

Shells out to "tail".
- Part of POSIX standard, included in GNU coreutils. Available on virtually all Unix/Linux systems.
- Handles optimization of reading from end of file without being too complicated
- Nearly the same code for with and without streaming

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -714,7 +714,11 @@ paths:
   
   /instances/{id}/logs:
     get:
-      summary: Get instance logs (tail)
+      summary: Stream instance logs (SSE)
+      description: |
+        Streams instance console logs as Server-Sent Events.
+        Returns the last N lines (controlled by `tail` parameter), then optionally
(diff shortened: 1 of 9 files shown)


rgarcia (approved):

openapi.yaml
```diff
@@ -724,30 +728,27 @@ paths:
...
         200:
-          description: Log stream
+          description: Log stream (SSE)
           content:
             text/event-stream:
               schema:
                 type: string
-            text/plain:
-              schema:
-                type: string
```
sjmiller609, line 750: still uses SSE even if not using -f option: allows for buffering of logs instead of all in memory at once option

PR #16, opened 2025-11-26
Volumes

sjmiller609:


openapi.yaml
```diff
@@ -140,7 +140,12 @@ components:
               description: Whether to attach instance to the default network
               default: true
               example: true
-        # Future: volumes, port_mappings, timeout_seconds
+        volumes:
+          type: array
+          description: Volumes to attach to the instance at creation time
+          items:
+            $ref: "#/components/schemas/VolumeAttachment"
+        # Future: port_mappings, timeout_seconds
```
rgarcia, line 148: is there some way when creating a volume to specify a host path? This is something we've requested from ukp in the past, e.g. as a way to to give a vm access to a directory

one other feature that comes to mind: the ability to create a volume from a zip file. We use this in the user code deployment path

both of these are not must-haves but just jotting them down

lib/volumes/manager.go
```diff
@@ -2,40 +2,212 @@ package volumes
...
 }
 
+// CreateVolume creates a new volume
 func (m *manager) CreateVolume(ctx context.Context, req CreateVolumeRequest) (*Volume, error) {
-	return nil, fmt.Errorf("volume creation not yet implemented")
+	// Generate or use provided ID
+	id := cuid2.Generate()
+	if req.Id != nil && *req.Id != "" {
+		id = *req.Id
+	}
```
rgarcia, line 91: feature request: ability to attach convenient `name`s to things vs dealing with IDs (for all CRUD ops). Could see this being useful for instances and volumes

lib/volumes/README.md
```diff
@@ -0,0 +1,33 @@
...
+- `/dev/vda` - rootfs (read-only image)
+- `/dev/vdb` - overlay (writable instance storage)
+- `/dev/vdc` - config disk (read-only)
+- `/dev/vdd`, `/dev/vde`, ... - attached volumes
+
+The init process inside the guest reads the requested mount paths from the config disk and mounts each volume at its specified path.
+
+## Constraints
+
+- Volumes can only be attached at instance creation time (no hot-attach)
```
rgarcia, line 26: is this supported by ch? could be cool

lib/volumes/README.md
```diff
@@ -0,0 +1,33 @@
...
+- A volume can only be attached to one instance at a time
```
rgarcia, line 27: feature request! this would be cool. I could see some use cases:
- read-only mounted on many instances for things shared across lots of instances, e.g. chrome extensions
- read-write with some kind of overlay for things shared across lots of instances that get lightly / rarely edited, e.g. default chromium user data and launch flags

lib/instances/storage.go
```diff
@@ -83,27 +84,7 @@ func (m *manager) saveMetadata(meta *metadata) error {
 // createOverlayDisk creates a sparse overlay disk for the instance
 func (m *manager) createOverlayDisk(id string, sizeBytes int64) error {
 	overlayPath := m.paths.InstanceOverlay(id)
```
rgarcia, line 86: random thoughts as i read this:
- it would be nice to be able to configure how big this overlay is for disk-hungry workloads (ignore if this is already the case)
- we should probably do something similar to how ukp has configurable limits on both the maximum resources that an individual vm can request and also the maximum resources consumed by hypeman in aggregate

rgarcia (approved):

lib/volumes/README.md
sjmiller609: Yes, I do think that's supported

lib/volumes/manager.go
sjmiller609: good idea

lib/instances/storage.go
sjmiller609: ok good idea

cmd/api/api/instances.go
```diff
@@ -154,17 +154,29 @@ func (s *ApiService) CreateInstance(ctx context.Context, request oapi.CreateInst
 }
 
 // GetInstance gets instance details
+// The id parameter can be either an instance ID or name
```
rgarcia, line 157: should document this in the openapi.yaml spec for endpoints that support it

in the Kernel API we also make the url param something like `/profiles/{id_or_name}` in the spec to make it extra clear

rgarcia (approved):
🚀

PR #17, opened 2025-11-28
gpu passthrough

rgarcia:
> [!NOTE]
> Introduce full GPU/PCI passthrough: device management + APIs, instance integration, initrd NVIDIA support, startup reconciliation, and comprehensive tests/docs.
> 
> - **Devices & API**:
>   - Add device management (`lib/devices`): discovery, VFIO bind/unbind, persistence, reconciliation, and liveness checks.
>   - New REST endpoints (`/devices`, `/devices/{id}`, `/devices/available`) with OpenAPI and generated client updates.
> - **Instances**:
>   - Support attaching devices on create; auto-bind to VFIO and mark attached; auto-unbind on delete.
>   - Pass devices to cloud-hypervisor (`VmConfig.Devices`); include devices in metadata and config disk.
> - **System/Initrd**:
>   - Build initrd with NVIDIA kernel modules and driver libs; init script loads modules, creates device nodes, and injects libs into guest/container when `HAS_GPU=1`.
>   - Kernel/version wiring for NVIDIA assets.
> - **Startup/Reconciliation**:
>   - On API start, reconcile device state, clear orphans, and integrate instance liveness.
> - **Wiring & Providers**:
>   - Wire `DeviceManager` through app, DI, and tests; adapt paths for device metadata.
> - **Tests & Docs**:
>   - Add extensive GPU E2E, NVML, and inference tests; docs for device/GPU usage and troubleshooting.
> - **Schemas/Clients/Deps**:
>   - Update OpenAPI schemas/clients; add new paths and types; refresh dependencies needed for features.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d23e7344f35255a0aa7a3d8cf381ff4e2633373. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/devices/scripts/gpu-reset.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+#
+# gpu-reset.sh - Reset GPU state after failed passthrough tests or hangs
```
sjmiller609, line 3: Note to self, compare if any cleanup logic could go on server start.

lib/devices/discovery.go
```diff
@@ -0,0 +1,257 @@
...
+func readDeviceInfo(pciAddress string) (*AvailableDevice, error) {
+	devicePath := filepath.Join(sysfsDevicesPath, pciAddress)
+
+	vendorID, err := readSysfsFile(filepath.Join(devicePath, "vendor"))
+	if err != nil {
+		return nil, fmt.Errorf("read vendor: %w", err)
+	}
+	vendorID = strings.TrimPrefix(vendorID, "0x")
+
+	deviceID, err := readSysfsFile(filepath.Join(devicePath, "device"))
```
sjmiller609, line 79: lots of joins in this file, idk if should go in lib/paths. It's not really changeable so seems to matter less in comparison to data_dir nested files.

lib/devices/manager.go
```diff
@@ -0,0 +1,422 @@
...
+func NewManager(p *paths.Paths) Manager {
+	return &manager{
+		paths:      p,
+		vfioBinder: NewVFIOBinder(),
+	}
+}
+
+func (m *manager) ListDevices(ctx context.Context) ([]Device, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
```
sjmiller609, line 106: are locks necessary on read actions like list and get?

lib/devices/manager.go
```diff
@@ -0,0 +1,422 @@
...
+	return &device, nil
+}
+
+func (m *manager) saveDevice(device *Device) error {
+	data, err := json.MarshalIndent(device, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(m.paths.DeviceMetadata(device.Id), data, 0644)
```
sjmiller609, line 735: Note: check on data being saved. Could some information be derived instead of saved?

lib/devices/types.go
```diff
@@ -0,0 +1,56 @@
...
+	Name        string     `json:"name"`         // user-provided globally unique name
+	Type        DeviceType `json:"type"`         // gpu or pci
+	PCIAddress  string     `json:"pci_address"`  // e.g., "0000:a2:00.0"
+	VendorID    string     `json:"vendor_id"`    // e.g., "10de"
+	DeviceID    string     `json:"device_id"`    // e.g., "27b8"
+	IOMMUGroup  int        `json:"iommu_group"`  // IOMMU group number
+	BoundToVFIO bool       `json:"bound_to_vfio"` // whether device is bound to vfio-pci
+	AttachedTo  *string    `json:"attached_to"`  // instance ID if attached, nil otherwise
+	CreatedAt   time.Time  `json:"created_at"`
+}
```
sjmiller609, line 28: I think this is the information getting saved to the metadata file. Is there anything here that we should derive instead of save? Reasoning for derive versus save being if the state changed outside of the control / more reliable. Ran into that with CH states, better to just grab the state from VMM api instead of mirroring the CH state to metadata file.

lib/devices/vfio.go
```diff
@@ -0,0 +1,289 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const (
+	vfioDriverPath  = "/sys/bus/pci/drivers/vfio-pci"
+	pciDriversPath  = "/sys/bus/pci/drivers"
+	vfioDevicePath  = "/dev/vfio"
+)
```
sjmiller609, line 17: probably makes more sense here than in paths since these would not change and aren't really connected with the data dir paths.

lib/devices/vfio.go
```diff
@@ -0,0 +1,289 @@
...
+		checkCmd := exec.Command("pgrep", "nvidia-persistenced")
+		if checkCmd.Run() != nil {
+			// Process doesn't exist, that's fine
+			return nil
+		}
+		return fmt.Errorf("failed to stop nvidia-persistenced (try: sudo systemctl stop nvidia-persistenced)")
+	}
+	
+	// Give it a moment to exit
+	time.Sleep(500 * time.Millisecond)
```
sjmiller609: arbitrary sleep?

lib/devices/vfio.go
```diff
@@ -0,0 +1,289 @@
...
+	return os.WriteFile(probePath, []byte(pciAddress), 0200)
+}
+
+// stopNvidiaPersistenced stops the nvidia-persistenced service
+// This service keeps /dev/nvidia* open and blocks driver unbind
+func (v *VFIOBinder) stopNvidiaPersistenced() error {
+	slog.Debug("stopping nvidia-persistenced service")
+	
+	// Try systemctl first (works as root)
+	cmd := exec.Command("systemctl", "stop", "nvidia-persistenced")
```
sjmiller609, line 179: note: review lifecycle of gpu device. Is it a one-time setup needed for a GPU, and what's the per-VM thing we need to do? If one-time setup is what needs special permissions then could we move one time setup to a cmd/ script run by operators to hook up a GPU to the system?

lib/devices/vfio.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	classCode = strings.TrimPrefix(classCode, "0x")
+	// Class 06 = Bridge, Subclass 04 = PCI bridge
+	return len(classCode) >= 4 && classCode[:2] == "06"
+}
+
+// GetDeviceSysfsPath returns the sysfs path for a PCI device (used by cloud-hypervisor)
+func GetDeviceSysfsPath(pciAddress string) string {
+	return filepath.Join(sysfsDevicesPath, pciAddress) + "/"
```
sjmiller609, line 307: another path example, not sure if want to move or not

lib/paths/paths.go
```diff
@@ -26,6 +26,9 @@
 //	      snapshots/
 //	        snapshot-latest/
 //	          config.json
+//	  devices/
+//	    {id}/
+//	      metadata.json
```
sjmiller609: yeah I think it makes sense for just the data dir stuff in this file

openapi.yaml
```diff
@@ -973,4 +1076,175 @@ paths:
...
+      summary: Register a device for passthrough
+      operationId: createDevice
+      security:
+        - bearerAuth: []
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateDeviceRequest"
```
sjmiller609, line 1393: I'm wondering about registering devices, if that's the part with a special permission or cleanup that gets complicated, could we have it done by operator instead of creating the device via API. Since you don't really add on the fly to the host new devices. Or it could be done all on startup of the server, kinda like how the host network is initialized on server start.

Kind of related, I'm not sure which cleanup logic could go in server startup, but that has also been handy for example with tap device cleanups. Especially for per-VM resources that might get stale if "pkill cloud-h" or similar direct to VMM actions run outside of API to taint the state (e.g. host powered off)

sjmiller609 (approved):
Mostly nits, but I think worth checking on:
- Any cleanup that should happen at server start, especially for tainted states, especially for per-VM resources?
- Related, any state we save that could always be derived instead of saved?
- Do we actually want / need ability to create devices on the host? why not just always use all available devices, configurable list of devices initialized on server startup, or something along those lines?

lib/devices/scripts/gpu-reset.sh
rgarcia: Good idea - adding startup reconciliation to this PR. Will detect orphaned AttachedTo states when instances were killed outside the API.

lib/devices/discovery.go
rgarcia: Agreed - sysfs paths are system-defined, keeping them here rather than lib/paths.

lib/devices/manager.go
rgarcia: RLock protects against concurrent directory iteration during creates/deletes. Cheap read lock, will add a comment explaining.

lib/devices/manager.go
rgarcia: BoundToVFIO already derived on read (lines 85, 181, 188). Adding startup reconciliation for AttachedTo - same pattern as CH state lesson.

lib/devices/types.go
rgarcia: Exactly right. BoundToVFIO derived on read. Adding startup reconciliation to handle stale AttachedTo.

lib/devices/vfio.go
rgarcia: Agreed - kernel/hardware paths, not app-configurable. Keeping here.

lib/devices/vfio.go
rgarcia: Yeah, replacing with polling loop (pgrep check with timeout).

lib/devices/vfio.go
rgarcia: Great question. One-time setup (register+bind) needs elevated privs, per-VM just passes VFIO group. Will create follow-up for operator tooling separation.

lib/devices/vfio.go
rgarcia: Same reasoning - system path, stays here.

lib/paths/paths.go
rgarcia: Agreed - lib/paths for data_dir only.

openapi.yaml
rgarcia: Adding startup reconciliation to this PR for stale state cleanup. Will create follow-ups for operator tooling and auto-discovery options.

lib/instances/create.go
```diff
@@ -191,16 +192,42 @@ func (m *manager) createInstance(
...
+				log.InfoContext(ctx, "auto-binding device to VFIO", "device", deviceRef, "pci_address", device.PCIAddress)
+				if err := m.deviceManager.BindToVFIO(ctx, device.Id); err != nil {
+					log.ErrorContext(ctx, "failed to bind device to VFIO", "device", deviceRef, "error", err)
+					return nil, fmt.Errorf("bind device %s to VFIO: %w", deviceRef, err)
+				}
+			}
+			resolvedDeviceIDs = append(resolvedDeviceIDs, device.Id)
+		}
+		log.DebugContext(ctx, "validated devices for passthrough", "id", id, "devices", resolvedDeviceIDs)
+	}
```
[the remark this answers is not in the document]
rgarcia: cc0efea

lib/instances/configdisk.go
```diff
@@ -105,6 +105,13 @@ GUEST_DNS="%s"
 `, netConfig.IP, cidr, netConfig.Gateway, netConfig.DNS)
 	}
 
+	// GPU passthrough configuration
+	// When devices are attached, set HAS_GPU=1 to trigger NVIDIA module loading in init
+	gpuSection := ""
+	if len(inst.Devices) > 0 {
+		gpuSection = "\n# GPU passthrough\nHAS_GPU=1\n"
+	}
```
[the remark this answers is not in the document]
rgarcia: true

lib/devices/README.md
```diff
@@ -0,0 +1,451 @@
...
+## Example: Full Workflow
+
+```bash
+# 1. Discover available devices
+curl localhost:8080/devices/available
+# → [{"pci_address": "0000:a2:00.0", "vendor_name": "NVIDIA Corporation", ...}]
+
+# 2. Register the GPU
+curl -X POST localhost:8080/devices \
+  -d '{"name": "l4-gpu", "pci_address": "0000:a2:00.0"}'
```
sjmiller609, line 54: What is the point of the register step?

Why not....
`GET /devices` = list all the available devices
`GET /devices/{id}` = get one device information

No other routes for devices

`POST /instances` (including device ID in the payload) attaches the device

sjmiller609: maybe just because it gets an ID and name similar to other resource types, seems good enough reason 👍

lib/devices/README.md
```diff
@@ -0,0 +1,451 @@
...
+}
+```
+
+This is not currently exposed through the hypeman API but could be added for HPC workloads.
+
+## Constraints and Limitations
+
+### IOMMU Requirements
+
+- **IOMMU must be enabled** in BIOS and kernel (`intel_iommu=on` or `amd_iommu=on`)
```
sjmiller609, line 165: check + log warning if not available on startup maybe

lib/devices/README.md
```diff
@@ -0,0 +1,451 @@
...
+- All devices in an IOMMU group must be passed through together
+- Some motherboards place many devices in the same group (ACS override may help)
+
+### VFIO Module Requirements
+
+The following kernel modules must be loaded:
+```bash
+modprobe vfio_pci
+modprobe vfio_iommu_type1
+```
```
sjmiller609, line 175: Same with this (regarding startup warning for debuggability)

lib/devices/GPU.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Performance Tuning
+
+### Huge Pages
+
+For best GPU performance, enable huge pages on the host:
+
+```bash
+echo 1024 > /proc/sys/vm/nr_hugepages
+```
```
sjmiller609, line 162: add to warning startup log and / or install script?

lib/devices/discovery.go
```diff
@@ -0,0 +1,279 @@
...
+		// 0300 = VGA controller, 0302 = 3D controller
+		if classPrefix == "0300" || classPrefix == "0302" {
+			return true
+		}
+	}
+
+	// Also include NVIDIA devices by vendor ID
+	if device.VendorID == "10de" {
+		return true
+	}
```
sjmiller609, line 185: > The fallback to include all NVIDIA devices by vendor ID (line 183) might catch non-passthrough devices like audio controllers attached to GPUs. Consider filtering by class in addition to vendor, or documenting that these will appear in the discovery list.

I think this is OK and also another justification for why to have "register" feature

lib/devices/manager.go
```diff
@@ -0,0 +1,802 @@
...
+		log.WarnContext(ctx, "failed to save device after reset-lite",
+			"device_id", device.Id,
+			"error", err,
+		)
+	}
+}
+
+// detectSuspiciousVMMProcesses logs warnings about cloud-hypervisor processes
+// that don't match known instances. This is log-only (no killing).
+func (m *manager) detectSuspiciousVMMProcesses(ctx context.Context, stats *reconcileStats) {
```
sjmiller609: This seems like it's not related to devices and so should probably live in a different module, like the instance module maybe.

lib/system/initrd.go
```diff
@@ -49,12 +53,18 @@ func (m *manager) buildInitrd(ctx context.Context, arch string) (string, error)
...
+
 	agentPath := filepath.Join(binDir, "exec-agent")
 	if err := os.WriteFile(agentPath, ExecAgentBinary, 0755); err != nil {
 		return "", fmt.Errorf("write exec-agent: %w", err)
 	}
 
+	// Add NVIDIA kernel modules (for GPU passthrough support)
+	if err := m.addNvidiaModules(ctx, rootfsDir, arch); err != nil {
+		// Log but don't fail - NVIDIA modules are optional (not available on all architectures)
+		fmt.Printf("initrd: skipping NVIDIA modules: %v\n", err)
```
sjmiller609: should this use context logger?

lib/system/initrd.go
```diff
@@ -135,10 +145,165 @@ func (m *manager) isInitrdStale(initrdPath string) bool {
...
+
+	// Extract tarball directly into rootfs
+	if err := extractTarGz(resp.Body, rootfsDir); err != nil {
+		return fmt.Errorf("extract nvidia modules: %w", err)
+	}
+
+	// Add userspace driver libraries (libcuda.so, libnvidia-ml.so, nvidia-smi, etc.)
+	// These are injected into containers at boot time - see lib/devices/GPU.md
+	if err := m.addNvidiaDriverLibs(ctx, rootfsDir, arch); err != nil {
+		fmt.Printf("initrd: warning: could not add nvidia driver libs: %v\n", err)
```
sjmiller609: context logger?

lib/system/initrd.go
```diff
@@ -135,10 +145,165 @@ func (m *manager) isInitrdStale(initrdPath string) bool {
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("download failed with status %d", resp.StatusCode)
+	}
+
+	// Extract tarball directly into rootfs
+	if err := extractTarGz(resp.Body, rootfsDir); err != nil {
+		return fmt.Errorf("extract nvidia driver libs: %w", err)
+	}
+
+	fmt.Printf("initrd: added NVIDIA driver libraries from %s\n", url)
```
sjmiller609: context logger?

sjmiller609: btw when using context logger, should show up automatically with
```
hypeman logs --source=hypeman <VM id>
```

lib/instances/configdisk.go
```diff
@@ -105,6 +107,17 @@ GUEST_DNS="%s"
 `, netConfig.IP, cidr, netConfig.Gateway, netConfig.DNS)
 	}
 
+	// GPU passthrough configuration
+	// Only set HAS_GPU=1 if at least one attached device is actually a GPU
+	gpuSection := ""
+	for _, deviceID := range inst.Devices {
+		device, err := m.deviceManager.GetDevice(ctx, deviceID)
+		if err == nil && device.Type == devices.DeviceTypeGPU {
```
sjmiller609, line 115: Looks like you already noticed but should probably not load gpu drivers if not gpu device

sjmiller609 (approved):
Looks great!

rgarcia:
hey steven, addressed your feedback:

**startup warnings for gpu prerequisites** (README.md:165, README.md:175, GPU.md:162)
- good call, added startup warnings for iommu and vfio modules. also added a hint for huge pages when devices are registered. a260091

**detectSuspiciousVMMProcesses location** (manager.go:656)
- yeah you're right, moved this to `lib/instances/liveness.go` where it belongs. also fixes the false positive bug from cursor bot since it now properly checks all instances, not just ones with devices. 8d610e9

**context loggers in initrd** (initrd.go:65, 208, 255)
- switched all the fmt.Printf calls to context loggers. since this is server-level logging (not per-instance), it'll just go to the main hypeman log but with proper structured format. 05fa5a4

**HAS_GPU flag** (configdisk.go:115)
- already fixed in cc0efea, now only sets HAS_GPU=1 when there's actually a gpu device attached 👍

PR #19, opened 2025-12-02
feat(registry): Add OCI Distribution registry for docker push support

rgarcia:
## Summary

Implements OCI Distribution Spec compliant registry endpoints that enable `docker push`-like functionality for uploading images directly to hypeman.

- **New `/v2` registry endpoints** - Full OCI Distribution Spec support for blob and manifest operations
- **Content-addressable blob storage** - Layers stored by digest with automatic deduplication
- **Layer caching** - Subsequent pushes skip already-stored blobs (verified by tests)
- **Async conversion** - Manifest push triggers background conversion to ext4 disk format
- **JWT authentication** - Registry endpoints use same auth as rest of API

### New Files
- `lib/registry/blob_store.go` - Filesystem-backed blob storage
- `lib/registry/registry.go` - Registry handler wrapping go-containerregistry
- `lib/registry/README.md` - Architecture documentation with Mermaid diagrams
- `cmd/api/api/registry_test.go` - Comprehensive tests

### Modified Files
- `lib/images/manager.go` - Added `ImportLocalImage()` for local push handling
- `lib/paths/paths.go` - Added OCI cache path helpers
- `cmd/api/main.go` - Mount `/v2` endpoints with auth

## Test Plan

- [x] `TestRegistryVersionCheck` - Basic /v2/ endpoint
- [x] `TestRegistryPushAndConvert` - Push image → convert to disk
- [x] `TestRegistryPushAndCreateInstance` - Full e2e: push → convert → start VM
- [x] `TestRegistryLayerCaching` - Verify second push skips blob uploads
- [x] `TestRegistrySharedLayerCaching` - Verify layer deduplication across images

All tests pass locally with KVM.

## Related

- CLI changes: onkernel/hypeman-cli (companion PR)


lib/registry/README.md
```diff
@@ -0,0 +1,151 @@
...
+    2d35eb...          # Layer blob (shared across all images)
+    706db5...          # Config blob
+    85f2b7...          # Manifest blob
+```
+
+## CLI Usage
+
+```bash
+# Push from local Docker daemon
+hypeman push myimage:latest
```
sjmiller609, line 141: does docker push work also, i.e. any image pusher would be compatible or no?

lib/registry/README.md
```diff
@@ -0,0 +1,151 @@
...
+## Limitations
+
+- **Docker v2 manifests**: Images from local Docker daemon use Docker v2 manifest format which may need additional handling for conversion. Images pulled directly from registries (OCI format) work seamlessly.
+- **Digest references only**: Conversion is triggered only for digest references (`@sha256:...`), not tag references (`:latest`). The CLI automatically converts to digest references.
+
+## Design Decisions
+
+### Why wrap go-containerregistry/pkg/registry?
+
+**What:** Use the existing registry implementation from go-containerregistry with custom blob storage.
```
sjmiller609, line 161: 👍

lib/registry/README.md
```diff
@@ -0,0 +1,151 @@
...
+- Enables content-addressable manifest storage consistent with layers
+
+### Why detect mediaType from manifest content?
+
+**What:** Parse manifest JSON to extract actual mediaType rather than hardcoding.
+
+**Why:**
+- Docker v2 (`application/vnd.docker.distribution.manifest.v2+json`) vs OCI (`application/vnd.oci.image.manifest.v1+json`)
+- Wrong mediaType in index.json causes "malicious manifest" errors during conversion
+- Content-based detection ensures compatibility with all sources
```
sjmiller609: Can our OCI to disk code (umoci) convert either kind of manifest to VM image?

cmd/api/api/registry_test.go
```diff
@@ -0,0 +1,422 @@
...
+	// Push to our test registry using digest reference (required for conversion trigger)
+	targetRef := serverHost + "/test/alpine@" + digest.String()
+	t.Logf("Pushing to %s...", targetRef)
+
+	dstRef, err := name.ParseReference(targetRef, name.Insecure)
+	require.NoError(t, err)
+
+	err = remote.Write(dstRef, img)
+	require.NoError(t, err)
+	t.Log("Push successful!")
```
sjmiller609, line 72: IIUC this uses `"github.com/google/go-containerregistry/pkg/v1/remote"` to read from a registry and write to our registry in one go (on the Write step, I think).

I think we might want to test other upload scenarios, especially the most-default one like this:

```
# Check the default docker build manifest type works to push + run when architecture matches
docker build -t test .
# does docker CLI work to push? If not then use hypeman CLI
docker push ...
# then do the disk conversion, boot check
```
Above, I'm assuming is similar to what a user would do.

For the above procedure, I think in testing it would be fine to just shell out to docker.

Also in the README it's mentioned about converting between manifest types. I think we should be testing each manifest type we expect to work. I think I remember an issue with wrong manifest type not being able to be converted to VM disk.

In summary, testing may benefit from, ranked most to least important:
- push / boot steps for each supported manifest type
- push test for multi-platform image, check expected behavior (only push desired architecture? only use desired architecture? not sure)
- push test for unsupported platform / architecture, check expected behavior (deny push?)
- should other "pushers" like docker work to push to our registry? If yes then add in test too

lib/registry/README.md
rgarcia: No, docker push won't work currently due to auth flow mismatch:

Docker CLI expects the v2 registry token auth flow (GET /v2/ returns 401 + WWW-Authenticate header, client calls token endpoint, retries with JWT). We use direct JWT bearer auth. The hypeman CLI uses go-containerregistry's RegistryToken which sends the JWT directly, bypassing this flow.

For docker push support we'd need to implement a token endpoint. Could be a follow-up if there's demand, but the hypeman CLI is the supported path for now.

lib/registry/README.md
rgarcia: Yes, both work. go-containerregistry normalizes Docker v2 manifests to OCI format when storing to the OCI layout, so umoci always sees OCI manifests regardless of the source format.

The Limitations section is overly cautious - I'll update it to clarify that both formats are supported.

cmd/api/api/registry_test.go
rgarcia: Great feedback on testing.

Re: docker push - it won't work because Docker CLI expects the v2 registry token auth flow (WWW-Authenticate challenge -> token endpoint -> retry). We use direct JWT auth.

**For this PR:**
- Adding tag reference support (so clients don't need to compute digests)
- Adding a test for tag-based push
- Simplifying hypeman-cli to push by tag instead of computing digest

**Follow-up items:**
- Multi-platform image handling  
- Docker registry token auth flow (if needed)

On manifest types: go-containerregistry normalizes everything to OCI format in the layout. Both Docker v2 and OCI formats work - the alpine tests already exercise Docker v2.

sjmiller609 (approved):

PR #20, opened 2025-12-02
Volume multi-attach

sjmiller609:


openapi.yaml
```diff
@@ -326,6 +334,23 @@ components:
           description: Size in gigabytes
           example: 10
     
+    VolumeAttachmentInfo:
```
sjmiller609, line 337: this is in a nested type because these are properties of being attached and in a list because might be attached to multiple vms

lib/system/init_script.go
```diff
@@ -82,13 +83,37 @@ if [ -n "${VOLUME_MOUNTS:-}" ]; then
     # Create mount point in overlay
     mkdir -p "/overlay/newroot${path}"
     
-    # Mount with appropriate options
-    if [ "$mode" = "ro" ]; then
-      mount -t ext4 -o ro "$device" "/overlay/newroot${path}"
+    if [ "$mode" = "overlay" ]; then
```
sjmiller609, line 86: btw it is possible to use go for the init script with some shelling out

lib/system/init_script.go
```diff
@@ -82,13 +83,37 @@ if [ -n "${VOLUME_MOUNTS:-}" ]; then
...
+      # Overlay mode: mount base read-only, create overlayfs with per-instance writable layer
+      overlay_device=$(echo "$vol" | cut -d: -f4)
+      
+      # Create temp mount points for base and overlay disk
+      base_mount="/mnt/vol-base-$(basename "$path")"
+      overlay_mount="/mnt/vol-overlay-$(basename "$path")"
+      mkdir -p "$base_mount" "$overlay_mount"
```
rgarcia, line 96: Are these temp mount points (`/mnt/vol-base-*`, `/mnt/vol-overlay-*`) cleaned up anywhere? They persist inside the guest VM for the lifetime of the instance, but since the guest rootfs is ephemeral this should be fine. Just want to confirm this is intentional and not a leak.

lib/instances/storage.go
```diff
@@ -87,6 +87,18 @@ func (m *manager) createOverlayDisk(id string, sizeBytes int64) error {
...
+// createVolumeOverlayDisk creates a sparse overlay disk for a volume attachment
+func (m *manager) createVolumeOverlayDisk(instanceID, volumeID string, sizeBytes int64) error {
+	// Ensure vol-overlays directory exists
+	overlaysDir := m.paths.InstanceVolumeOverlaysDir(instanceID)
+	if err := os.MkdirAll(overlaysDir, 0755); err != nil {
+		return fmt.Errorf("create vol-overlays directory: %w", err)
+	}
+
+	overlayPath := m.paths.InstanceVolumeOverlay(instanceID, volumeID)
+	return images.CreateEmptyExt4Disk(overlayPath, sizeBytes)
```
rgarcia, line 102: This creates the overlay disk, but what happens if instance creation fails after this point? I see the cleanup stack in create.go adds volume detach, but the overlay disk itself (in vol-overlays/) would remain orphaned until deleteInstanceData is called. Is this handled by the existing cleanup stack that eventually calls deleteInstanceData, or could we leak overlay disks on partial failure?

lib/instances/volumes_test.go
```diff
@@ -0,0 +1,230 @@
...
+		output, code, err = execCommand(ctx, vsockSocket, command...)
+		if err == nil {
+			return output, code, nil
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+	return output, code, err
+}
+
+// TestVolumeMultiAttachReadOnly tests that a volume can be:
```
rgarcia, line 47: Nice integration test! A couple of missing test scenarios to consider:

1. **Overlay disk cleanup on instance deletion** - verify the vol-overlays/ directory is removed when an overlay-attached instance is deleted
2. **Error paths in validation** - test that invalid overlay configs (e.g., overlay=true with readonly=false, or missing overlay_size) return proper errors
3. **Concurrent multi-attach** - what happens if two instances try to attach the same volume simultaneously?

rgarcia (commented):
Added some review comments - mostly questions about cleanup paths and test coverage. The multi-attach and overlay implementation looks solid overall.

rgarcia (approved):

lib/system/init_script.go
sjmiller609: Yes, I think it's OK because they disappear when the VM is turned off and init runs only at boot

PR #22, opened 2025-12-02
Initialize volume with data

sjmiller609:


lib/volumes/archive.go
```diff
@@ -0,0 +1,189 @@
...
+
+// ExtractTarGz extracts a tar.gz archive to destDir, aborting if the extracted
+// content exceeds maxBytes. Returns the total extracted bytes on success.
+//
+// Safety measures against adversarial archives:
+// - Rejects archives containing path traversal attempts or absolute paths
+// - Tracks cumulative extracted size, aborts immediately if limit exceeded
+// - Uses securejoin for safe path joining (defense in depth)
+// - Uses io.LimitReader as secondary protection when copying files
+func ExtractTarGz(r io.Reader, destDir string, maxBytes int64) (int64, error) {
```
sjmiller609, line 57: This is very security sensitive especially since running with elevated privileges. Large file upload and tarball extraction common issue area, so want to look into a good dependency for this perhaps. It has to extract a tar into a subdirectory and limit to a specific size securely. Or maybe there's other solution like straight from tar into filesystem but not sure if that's doable. Anyways worth another look before merge.

cmd/api/api/volumes.go
```diff
@@ -157,3 +160,92 @@ func volumeToOAPI(vol volumes.Volume) oapi.Volume {
 	return oapiVol
 }
 
+// CreateVolumeFromArchiveHandler handles POST /volumes/from-archive
+// This is a custom endpoint (outside OpenAPI spec) that accepts multipart form data
+// with a tar.gz file to create a pre-populated volume.
```
rgarcia: why not put it in the openapi spec? We do this in the main API and it works fine / the generated SDKs are good https://github.com/onkernel/kernel/blob/26fbf0e3558bf55709d4fd52f8ae5438ebf765e5/packages/api/openapi.yaml#L2945-L3009

lib/volumes/archive.go
rgarcia: agree... https://github.com/cyphar/filepath-securejoin is the only thing i could find

containerd or moby might have code we can borrow from

rgarcia (approved):

cmd/api/api/volumes.go
sjmiller609: you're right, this was a hallucinated problem

PR #24, opened 2025-12-04
Ingress

sjmiller609:


openapi.yaml
```diff
@@ -399,6 +399,82 @@ components:
...
           example: ok
+    
+    IngressMatch:
+      type: object
+      required: [hostname]
+      properties:
+        hostname:
+          type: string
+          description: Hostname to match (exact match on Host header)
+          example: api.example.com
```
rgarcia, line 410: this will be slightly coupled with the envoy config, e.g. the SSL cert it uses would need to support this hostname. not sure how to think about this constraint at the moment but it's something we deal with in unikraft land. Unikraft on startup provisions (and re-issues if expiring soon) an SSL wildcard cert on a specified domain, and then all instance FQDNs are subdomains on this

openapi.yaml
```diff
@@ -399,6 +399,82 @@ components:
...
+        port:
+          type: integer
+          description: Host port to listen on for this rule (default 80)
+          default: 80
+          example: 8080
```
rgarcia, line 415: should default be 443?

lib/ingress/README.md
```diff
@@ -0,0 +1,147 @@
...
+                                    | route -> my-api:8080    |
+                                    +------------------------>|
+                                         10.100.x.y:8080
+```
+
+## How It Works
+
+### Envoy Daemon
+
+- Envoy binary is embedded in hypeman (like Cloud Hypervisor)
```
rgarcia, line 21: woah neat

lib/ingress/README.md
```diff
@@ -0,0 +1,147 @@
...
+- Extracted to `/var/lib/hypeman/system/binaries/envoy/{version}/{arch}/envoy` on first use
+- Runs as a daemon process that survives hypeman restarts
+- Listens on `0.0.0.0:80` (configurable via `ENVOY_LISTEN_ADDRESS` and `ENVOY_LISTEN_PORT`)
```
rgarcia, line 24: feels like the default should be something that is natural for SSL connections like 443

lib/ingress/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Config Updates
+1. Regenerate `config.yaml` from all ingresses
+2. Perform hot restart by starting a new Envoy process with an incremented restart epoch
+3. New Envoy process coordinates with the old one to take over without dropping connections
+
+### Shutdown
+- By default (`ENVOY_STOP_ON_SHUTDOWN=false`), Envoy continues running when hypeman exits
+- Set `ENVOY_STOP_ON_SHUTDOWN=true` to stop Envoy with hypeman
+- Envoy can be manually stopped via admin API (`/quitquitquit`) or SIGTERM
```
rgarcia, line 151: lol is this really the endpoint

lib/ingress/README.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+Tests use:
+- Mock instance resolver (no real VMs needed)
+- Temporary directories for filesystem operations
+- Non-privileged ports to avoid permission issues
+
+## Future Improvements
+
+- TLS termination with ACME/Let's Encrypt
```
rgarcia, line 167: i think for platform integration we could punt on this and rely on continuing to use metro-api as the TLS terminator

however we would have to change metro-api since it piggybacks on the ACME/Let's Encrypt cert that unikraft's daemon mints and manages...a daemon that won't necessarily exist on hypeman machines. In which case we would need to implement the cert stuff in metro-api or hypeman

anyway i think for platform integration we do need to figure out some way to make https / ssl connections work since this is a requirement for live view and CDP URLs we pass back to users

rgarcia (commented):
just looked at openapi spec and lib/ingress/README.md

lib/ingress/README.md
sjmiller609: yep

lib/ingress/README.md
sjmiller609: we discussed in another context taking TLS in a separate pr

rgarcia (commented):
Looking gooooood. Added some questions / comments.

lib/ingress/storage.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			continue
+		}
+
+		var stored storedIngress
+		if err := json.Unmarshal(data, &stored); err != nil {
+			continue
+		}
+
+		if stored.Name == name {
```
rgarcia: Bug: Line 180 declares `name := entry.Name()` which shadows the function parameter `name`. Then line 197 compares `stored.Name == name` against the filename instead of the function parameter. Should rename the loop variable (e.g., `fileName`).

lib/providers/providers.go
```diff
@@ -119,3 +120,55 @@ func ProvideVolumeManager(p *paths.Paths, cfg *config.Config) (volumes.Manager,
...
+	// Check if instance has network enabled
+	if !inst.NetworkEnabled {
+		return "", fmt.Errorf("instance %s has no network configured", nameOrID)
+	}
+
+	// Check if instance has an IP assigned
+	if inst.IP == "" {
+		return "", fmt.Errorf("instance %s has no IP assigned", nameOrID)
+	}
+
```
rgarcia: Maybe move `instanceResolverAdapter` to `lib/ingress/resolver.go`? Seems a little off to put functional logic in providers.go, and might make more sense to colocate the default implementation of an interface with the interface.

lib/ingress/daemon.go
```diff
@@ -0,0 +1,429 @@
...
+	// Clean up shared memory
+	d.cleanupStaleSharedMemory()
+
+	return nil
+}
+
+// ReloadConfig performs a hot restart to reload Envoy's configuration.
+// This spawns a new Envoy process that coordinates with the old one to take over
+// without dropping connections during the drain period.
+func (d *EnvoyDaemon) ReloadConfig() error {
```
rgarcia, line 175: When we do zero-downtime hypeman deployments, there might be a short period when two hypeman processes are running. Not sure if there are implications for locking around config generation, epoch management, etc., but something to think about.

lib/ingress/config.go
```diff
@@ -0,0 +1,442 @@
...
+				log.WarnContext(ctx, "skipping cluster: cannot resolve instance IP",
+					"ingress_id", ingress.ID,
+					"instance", rule.Target.Instance,
+					"error", err)
+				continue
+			}
+
+			cluster := map[string]interface{}{
+				"name":            clusterName,
+				"connect_timeout": "5s",
```
rgarcia, line 475: Consider whether Envoy clusters should include health checks. Connect timeout is set (line 305), but active health checking could help detect unhealthy backends.

rgarcia:
Question: Will an instance IP ever change after ingress creation? (e.g., after restore from standby?) If so, would the ingress need to be regenerated?

sjmiller609:
> Question: Will an instance IP ever change after ingress creation? (e.g., after restore from standby?) If so, would the ingress need to be regenerated?

No it doesn’t but also just realized there should probably be a way to stop then reboot. Right now we only provide standby and restore.

lib/ingress/daemon.go
sjmiller609: Just realized, should be using dynamic resources instead of static resources, this lets envoy update itself without this outside orchestration.

https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/configuration-dynamic-filesystem

lib/ingress/config.go
sjmiller609: I think we shouldn't add health checks because this is mostly applicable if we are load balancing between multiple upstream servers from the ingress (e.g. horizontal scaling). Also we would have to know which HTTP endpoint works like a health check endpoint. So, I think we would want to include healthcheck as configurable on the ingress resource in a future change.

lib/ingress/README.md
```diff
@@ -51,8 +51,8 @@ An Ingress is a configuration object that defines how external traffic should be
 1. User creates an ingress via API
 2. Manager validates the ingress (name, instance exists, hostname unique)
 3. Ingress is persisted to `/var/lib/hypeman/ingresses/{id}.json`
-4. Envoy config is regenerated from all ingresses
-5. Envoy performs a hot restart to pick up the new config
+4. Envoy xDS config files (LDS/CDS) are regenerated from all ingresses
+5. Envoy automatically detects the file changes and reloads (no restart needed)
```
rgarcia, line 55: oh that's really nice

lib/instances/manager_test.go
```diff
@@ -406,7 +406,7 @@ func TestBasicEndToEnd(t *testing.T) {
 	client := &http.Client{Timeout: 2 * time.Second}
 	var resp *http.Response
 	var lastErr error
-	deadline := time.Now().Add(5 * time.Second)
+	deadline := time.Now().Add(30 * time.Second)
```
rgarcia: this seems like a long time for it to take to come online?

lib/paths/paths.go
```diff
@@ -255,9 +255,19 @@ func (p *Paths) EnvoyBinary(version, arch string) string {
...
+}
+
+// EnvoyLDS returns the path to the Listener Discovery Service config file.
+func (p *Paths) EnvoyLDS() string {
+	return filepath.Join(p.EnvoyDir(), "lds.yaml")
+}
+
+// EnvoyCDS returns the path to the Cluster Discovery Service config file.
+func (p *Paths) EnvoyCDS() string {
+	return filepath.Join(p.EnvoyDir(), "cds.yaml")
```
rgarcia, line 235: worried that the comment at the top of this file is going out of date. maybe we should get rid of it and move it to a readme instead and have one place where we document the directory structure?

(could even have a cursor agent in ci ensure it's always up to date)

rgarcia (approved):

lib/instances/manager_test.go
sjmiller609: I don't think it's actually taking that long usually. Not sure why it was flaking with less time.
```
2025-12-05T16:21:27.0596808Z         2025/12/05 16:21:26 [notice] 751#751: OS: Linux 6.12.8+
2025-12-05T16:21:27.0597290Z         2025/12/05 16:21:26 [notice] 751#751: getrlimit(RLIMIT_NOFILE): 1024:4096
2025-12-05T16:21:27.0597762Z         2025/12/05 16:21:26 [notice] 751#751: start worker processes
2025-12-05T16:21:27.0598217Z         2025/12/05 16:21:26 [notice] 751#751: start worker process 765
2025-12-05T16:21:27.0598648Z     manager_test.go:332: Testing ingress routing to nginx...
2025-12-05T16:21:27.0599114Z     manager_test.go:345: Using random ports: ingress=41993, admin=43235
2025-12-05T16:21:27.0599530Z     manager_test.go:358: Instance IP: 10.100.249.139
2025-12-05T16:21:27.0599877Z     manager_test.go:368: Starting Envoy...
2025-12-05T16:21:27.4378280Z     manager_test.go:381: Creating ingress rule...
2025-12-05T16:21:27.5019265Z     manager_test.go:400: Ingress created: i7powqb2p2v0fgrvveherh82
2025-12-05T16:21:27.5019867Z     manager_test.go:405: Making HTTP request through Envoy to nginx...
2025-12-05T16:21:27.5151213Z     manager_test.go:435: Got response from nginx through Envoy: 615 bytes
2025-12-05T16:21:27.5669359Z     manager_test.go:440: Ingress deleted
```

sjmiller609: it's flaking somehow. let me check on this

sjmiller609: the problem was we need to update cds.yaml before updating lds.yaml because clusters must exist before listeners. but it was working most of the time because they were updated very close in time.

PR #27, opened 2025-12-07
Support TLS for ingress

sjmiller609:


lib/instances/manager_test.go
```diff
@@ -418,28 +420,180 @@ func TestBasicEndToEnd(t *testing.T) {
...
+		_ = envPath // silence unused warning
+	}
+
+	acmeEmail := os.Getenv("ACME_EMAIL")
+	acmeDNSProvider := os.Getenv("ACME_DNS_PROVIDER")
+	cloudflareToken := os.Getenv("CLOUDFLARE_API_TOKEN")
+	tlsTestDomain := os.Getenv("TLS_TEST_DOMAIN")
+	acmeCA := os.Getenv("ACME_CA")
+
+	if acmeEmail != "" && acmeDNSProvider == "cloudflare" && cloudflareToken != "" && tlsTestDomain != "" {
```
sjmiller609, line 462: integration test with actual ACME takes about 20s.
```
Mon, 08 Dec 2025 20:38:39 GMT
    manager_test.go:535: TLS Ingress created: zm2v94iz7pkwqmnqfaegsz8b
Mon, 08 Dec 2025 20:38:39 GMT
    manager_test.go:539: Making HTTPS request (certificate will be obtained on first request)...
Mon, 08 Dec 2025 20:38:39 GMT
    manager_test.go:572: TLS request attempt failed: Get "https://127.0.0.1:34857/": remote error: tls: internal error (retrying...)
Mon, 08 Dec 2025 20:38:41 GMT
   ...
    manager_test.go:572: TLS request attempt failed: Get "https://127.0.0.1:34857/": remote error: tls: internal error (retrying...)
Mon, 08 Dec 2025 20:38:53 GMT
    manager_test.go:587: Got HTTPS response from nginx through Caddy: 615 bytes
Mon, 08 Dec 2025 20:38:53 GMT
    manager_test.go:592: TLS Ingress deleted
```

sjmiller609: disable by removing envs from test.yml

lib/ingress/config.go
```diff
@@ -2,356 +2,176 @@ package ingress
...
-		validator:     validator,
-		otel:          otel,
+// ParseDNSProvider parses a string into a DNSProvider, returning an error for unknown values.
+func ParseDNSProvider(s string) (DNSProvider, error) {
+	switch s {
+	case "":
+		return DNSProviderNone, nil
+	case "cloudflare":
+		return DNSProviderCloudflare, nil
+	case "route53":
```
sjmiller609: did not actually test route53. should delete it?

cmd/api/api/ingress.go
```diff
@@ -75,11 +85,27 @@ func (s *ApiService) CreateIngress(ctx context.Context, request oapi.CreateIngre
...
+			return oapi.CreateIngress400JSONResponse{
+				Code:    "domain_not_allowed",
+				Message: err.Error(),
+			}, nil
+		case errors.Is(err, ingress.ErrConfigValidationFailed):
+			log.ErrorContext(ctx, "failed to create ingress", "error", err, "name", request.Body.Name)
+			return oapi.CreateIngress400JSONResponse{
+				Code:    "config_validation_failed",
+				Message: err.Error(),
+			}, nil
```
rgarcia, line 108: i wonder if we could make a cursor agent workflow that runs on all PRs and examines / double checks the exhaustiveness of these switch statements

cmd/api/config/config.go
```diff
@@ -133,13 +144,22 @@ func Load() *Config {
...
-		EnvoyListenAddress: getEnv("ENVOY_LISTEN_ADDRESS", "0.0.0.0"),
-		EnvoyAdminAddress:  getEnv("ENVOY_ADMIN_ADDRESS", "127.0.0.1"),
-		EnvoyAdminPort:     getEnvInt("ENVOY_ADMIN_PORT", 9901),
-		// For production, set to false
-		// allows for updating hypeman without restarting envoy
-		EnvoyStopOnShutdown: getEnvBool("ENVOY_STOP_ON_SHUTDOWN", true),
+		// Caddy / Ingress configuration
+		CaddyListenAddress:  getEnv("CADDY_LISTEN_ADDRESS", "0.0.0.0"),
+		CaddyAdminAddress:   getEnv("CADDY_ADMIN_ADDRESS", "127.0.0.1"),
+		CaddyAdminPort:      getEnvInt("CADDY_ADMIN_PORT", 2019),
```
rgarcia: should this be randomly assigned so that there's no conflict on shared dev machines?

lib/dns/server.go
```diff
@@ -0,0 +1,212 @@
...
+	"net"
+	"strings"
+	"sync"
+
+	"github.com/miekg/dns"
+)
+
+const (
+	// DefaultPort is the default port for the local DNS server.
+	DefaultPort = 5353
```
rgarcia: same here--wondering about shared dev machine conflicts

lib/dns/server.go
```diff
@@ -0,0 +1,212 @@
...
+	mux.HandleFunc(Suffix+".", s.handleQuery)
+
+	// Bind to UDP socket first to get actual port (important when port is 0)
+	addr := fmt.Sprintf("127.0.0.1:%d", s.port)
+	conn, err := net.ListenPacket("udp", addr)
+	if err != nil {
+		return fmt.Errorf("bind DNS server: %w", err)
+	}
+
+	// Update port to actual assigned port (useful when s.port was 0)
```
rgarcia, line 90: ah maybe default port for stuff like this should be 0 for random assignment?

lib/dns/server.go
```diff
@@ -0,0 +1,212 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.running {
+		return nil
+	}
+
+	// Store context for use in resolver calls
+	// This allows DNS resolution to respect cancellation from the parent context
+	s.ctx = ctx
```
rgarcia: Technically, I think this means that DNS queries could fail if the server is still running but shutting down. Would it be better to use context.Background and pass a context per dns query?

lib/ingress/README.md
```diff
@@ -35,29 +38,106 @@ An Ingress is a configuration object that defines how external traffic should be
...
+**Example configuration:**
+```bash
+# Allow TLS for any subdomain of example.com plus the apex
+TLS_ALLOWED_DOMAINS=example.com,*.example.com
+
+# Allow TLS for specific subdomains only
+TLS_ALLOWED_DOMAINS=api.example.com,www.example.com
+
+# Allow TLS for any domain (not recommended for production)
+TLS_ALLOWED_DOMAINS=*
```
rgarcia, line 125: is there any situation in which this should be allowed?

lib/ingress/README.md
```diff
@@ -35,29 +38,106 @@ An Ingress is a configuration object that defines how external traffic should be
...
+```
+
+#### Warning Scenarios
+
+The ingress manager logs warnings in these situations:
+
+- **TLS ingresses exist but ACME not configured**: If existing ingresses have `tls: true` but `ACME_EMAIL` or `ACME_DNS_PROVIDER` is not set, a warning is logged at startup. TLS will not work until ACME is configured.
+
+- **Domain not in allowed list**: Creating an ingress with `tls: true` for a hostname not in `TLS_ALLOWED_DOMAINS` will fail with error `domain_not_allowed`.
```
rgarcia, line 134: should there be a warning if hypeman starts up and sees ingresses that were previously created that are not in the allowed domain list?

lib/ingress/README.md
```diff
@@ -93,62 +172,64 @@ DELETE /ingresses/{id} - Delete ingress
...
-- Request rates and error counts (aggregate, not per-request)
-- Upstream health (backend availability, retries)
-- Listener and cluster statistics
-- Memory and resource usage
+| Variable | Description |
+|----------|-------------|
+| `CLOUDFLARE_API_TOKEN` | Cloudflare API token with DNS edit permissions |
 
-**Note:** Per-request tracing is intentionally disabled to protect tenant privacy. Only aggregate operational metrics are exported.
+**Note on Ports:** Each ingress rule can specify a `port` in the match criteria to listen on a specific host port. If not specified, defaults to port 80. Caddy dynamically listens on all unique ports across all ingresses.
```
rgarcia, line 201: starting to think that we each will need our own dev machine ha... port collisions seem likely. But maybe there's a way to detect whether a port is in use before creating an ingress for it?

lib/ingress/README.md
```diff
@@ -164,7 +245,7 @@ Tests use:
 
 ## Future Improvements
 
-- TLS termination with ACME/Let's Encrypt
 - Path-based L7 routing
 - Health checks for backends
-- Connection draining for graceful config updates
\ No newline at end of file
+- Rate limiting
+- Custom error pages
```
rgarcia, line 251: Other DNS providers? Not sure what most common other ones are. Maybe R53

lib/ingress/config.go
```diff
@@ -2,611 +2,494 @@ package ingress
...
-func (g *EnvoyConfigGenerator) atomicWrite(path string, data []byte) error {
-	dir := filepath.Dir(path)
-	tempFile, err := os.CreateTemp(dir, "envoy-*.yaml")
-	if err != nil {
-		return fmt.Errorf("create temp file: %w", err)
-	}
-	tempPath := tempFile.Name()
+	// Build routes from ingresses
+	routes := []interface{}{}
+	redirectRoutes := []interface{}{}
```
rgarcia, line 180: Does Caddy export any types we could use here?

lib/ingress/logs.go
```diff
@@ -0,0 +1,139 @@
...
+	"log/slog"
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/onkernel/hypeman/lib/paths"
+)
+
+// CaddyLogForwarder tails Caddy's system log and forwards to OTEL.
```
rgarcia, line 16: This is neat. At first I was skeptical that this should be at the application level, but I think it makes sense to do this

.env.example
```diff
@@ -1,2 +1,101 @@
...
+# -----------------------------------------------------------------------------
+# Cloudflare DNS Provider (ACME_DNS_PROVIDER=cloudflare)
+# -----------------------------------------------------------------------------
+# CLOUDFLARE_API_TOKEN=your-api-token
+# Token needs Zone:DNS:Edit permissions for the domains you want certificates for
+
+# -----------------------------------------------------------------------------
+# AWS Route53 DNS Provider (ACME_DNS_PROVIDER=route53)
+# -----------------------------------------------------------------------------
+# Route53 supports three authentication methods:
```
rgarcia: Remove these until we actually support it?

lib/ingress/daemon.go
```diff
@@ -16,190 +17,243 @@ import (
...
+	pidPath := d.paths.CaddyPIDFile()
 	if err := os.WriteFile(pidPath, []byte(strconv.Itoa(pid)), 0644); err != nil {
-		// Non-fatal, log but continue
 		log := logger.FromContext(ctx)
 		log.WarnContext(ctx, "failed to write PID file", "error", err)
 	}
 
 	// Wait for admin API to be ready
-	waitCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	waitCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
rgarcia, line 137: nice - using `context.Background()` here ensures startup wait isn't cancelled if the parent context times out. worth a brief comment explaining the intentional choice.

lib/instances/manager_test.go
```diff
@@ -418,28 +421,181 @@ func TestBasicEndToEnd(t *testing.T) {
...
+				},
+			},
+		}
+
+		tlsIng, err := tlsIngressManager.Create(ctx, tlsIngressReq)
+		require.NoError(t, err)
+		require.NotNil(t, tlsIng)
+		t.Logf("TLS Ingress created: %s", tlsIng.ID)
+
+		// Wait for certificate to be issued (this can take 10-60 seconds with DNS-01)
```
rgarcia, line 535: nit: the envPath variable and the `if true` construct is a bit confusing - looks like leftover refactoring. could simplify to just the directory walk loop.

lib/ingress/types.go
```diff
@@ -45,6 +58,106 @@ type IngressMatch struct {
...
+
+	// Split hostname into parts
+	parts := strings.Split(m.Hostname, ".")
+	if len(parts) < 2 {
+		return nil, fmt.Errorf("hostname pattern %q must have at least two parts", m.Hostname)
+	}
+
+	captures := []string{}
+	caddyLabels := make(map[string]string)
+	wildcardParts := make([]string, len(parts))
```
rgarcia, line 106: the Caddy label indexing (right-to-left) is well documented here. nice detailed explanation of how `foo.bar.example.com` maps to labels.0=com, labels.1=example, etc.

rgarcia (approved):
solid migration from envoy to caddy with TLS support. nice work on the pattern hostname feature and the internal DNS resolver for dynamic upstreams.

lib/ingress/README.md
sjmiller609: if you have many domains in your cloudflare and don't want to configure them all, but it's rare

lib/ingress/config.go
sjmiller609: Leaving this unaddressed. One point is that POSTing the change to the caddy API does validate it, so we at least have that but it's not build time validation to be fair

PR #28, opened 2025-12-08
Fix unauthenticated dockerhub errors in tests

sjmiller609:


sjmiller609:
problem was using sudo in makefile for "make test" need to adjust how docker login works

PR #29, opened 2025-12-09
Setup Github Release automation

sjmiller609:


.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.25'
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: '~> v2'
+          args: release --clean
```
sjmiller609, line 31:
```suggestion
        with:
          # gorelease-pro is not required
          distribution: goreleaser
          version: '~> v2'
          args: release --clean
```

sjmiller609:
test failing on envoy for some reason, but that's getting replaced in another PR so let's not address that

rgarcia (approved):

PR #41, opened 2025-12-20
Move vmm calls behind hypervisor interface

sjmiller609:
We will later introduce QEMU to support vGPU passthrough, then we should move cloud hypervisor behind a strategy interface. This change moves CH behind that abstraction, in preparation to later add QEMU. See lib/hypervisor/README.md

> [!NOTE]
> Introduce a hypervisor interface with a Cloud Hypervisor implementation and migrate instance lifecycle, metadata, paths, networking, and tests from direct VMM usage to the new abstraction.
> 
> - **Hypervisor Layer**:
>   - Add `lib/hypervisor` abstraction (`Hypervisor`, `ProcessManager`, `VMConfig`, `Capabilities`, `VMInfo`, `Type`).
>   - Implement Cloud Hypervisor adapter in `lib/hypervisor/cloudhypervisor/*` (API mapping, process manager, capabilities).
> - **Instances**:
>   - Migrate lifecycle ops (`create/start/stop/standby/restore/delete/query`) to use `hypervisor.Hypervisor` instead of `vmm`.
>   - Update VM config builder to output `hypervisor.VMConfig`.
>   - Metadata changes: replace `CHVersion`/`CHPID` with `HypervisorVersion`/`HypervisorPID` and add `HypervisorType`; state derivation via hypervisor.
>   - Graceful shutdown/kill helpers renamed to hypervisor equivalents; snapshot/restore via interface.
> - **Paths/Networking**:
>   - Change `paths.InstanceSocket(id, socketName)`; register socket name per hypervisor; update network allocation derivation to use `SocketNameForType`.
> - **Tests/Docs**:
>   - Update tests to new metadata fields and hypervisor client usage; rename orphaned-process cleanup logs; add READMEs explaining the abstraction and filesystem changes.
> - **Misc**:
>   - Improve VMM startup error reporting by including `vmm.log` when socket creation fails.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8358088e5c0c17a90cd39578ee67391b27221d44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/instances/README.md
+++ b/lib/instances/README.md
@@ -40,7 +40,7 @@ Manages VM instance lifecycle using Cloud Hypervisor.
       metadata.json             # State, versions, timestamps
       overlay.raw               # 50GB sparse writable overlay
       config.erofs              # Compressed config disk
-      cloud-hypervisor.sock     # Hypervisor API socket (named after hypervisor type)
+      ch.sock                   # Hypervisor API socket (abbreviated for SUN_LEN limit)
       logs/
         app.log                 # Guest application log (serial console output)
         vmm.log                 # Hypervisor log (stdout+stderr)
--- a/lib/instances/README.md
+++ b/lib/instances/README.md
@@ -40,10 +40,11 @@ Manages VM instance lifecycle using Cloud Hypervisor.
       metadata.json             # State, versions, timestamps
       overlay.raw               # 50GB sparse writable overlay
       config.erofs              # Compressed config disk
-      ch.sock                   # Cloud Hypervisor API socket
-      ch-stdout.log             # CH process output
+      cloud-hypervisor.sock     # Hypervisor API socket (named after hypervisor type)
       logs/
-        console.log             # Serial console (VM output)
+        app.log                 # Guest application log (serial console output)
+        vmm.log                 # Hypervisor log (stdout+stderr)
+        hypeman.log             # Hypeman operations log
       snapshots/
         snapshot-latest/        # Snapshot directory
           config.json           # VM configuration
--- a/lib/instances/types.go
+++ b/lib/instances/types.go
@@ -3,7 +3,7 @@ package instances
 import (
 	"time"
 
-	"github.com/onkernel/hypeman/lib/vmm"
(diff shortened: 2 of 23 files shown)


lib/instances/README.md
```diff
@@ -40,10 +40,11 @@ Manages VM instance lifecycle using Cloud Hypervisor.
       metadata.json             # State, versions, timestamps
       overlay.raw               # 50GB sparse writable overlay
       config.erofs              # Compressed config disk
-      ch.sock                   # Cloud Hypervisor API socket
-      ch-stdout.log             # CH process output
+      ch.sock                   # Hypervisor API socket (abbreviated for SUN_LEN limit)
```
rgarcia, line 43: rename to something generic or keep "ch"? also what is SUN_LEN?

lib/instances/types.go
```diff
@@ -56,9 +56,12 @@ type StoredMetadata struct {
 	StoppedAt *time.Time // Last time VM was stopped
 
 	// Versions
-	KernelVersion string        // Kernel version (e.g., "ch-v6.12.9")
-	CHVersion     vmm.CHVersion // Cloud Hypervisor version
-	CHPID         *int          // Cloud Hypervisor process ID (may be stale after host restart)
+	KernelVersion string // Kernel version (e.g., "ch-v6.12.9")
```
rgarcia, line 59: how custom is the ch-forked kernel? Will QEMU need a separate set of kernel versions?

rgarcia (approved):
🚀

lib/instances/README.md
sjmiller609: Set it up so that the socket is named after the VMM in use, helps see what's going on more easily with "ls". SUN_LEN is a limit on the length of socket files, which is why it's not named "cloud-hypervisor.sh" (that shows up in an error if you make it too long).

lib/instances/types.go
sjmiller609: Good question. I think it should work, but we will find out if not. I'll keep an eye out for that.

PR #50, opened 2025-12-24
feat: add systemd mode for full VM experience

rgarcia:
## Summary

This PR adds support for running systemd-based OCI images, enabling a VM experience where systemd is PID 1 and manages the full system.

## Motivation

Previously, hypeman only supported "exec mode" where the Go init binary runs as PID 1 and executes the container entrypoint directly. This works great for Docker-style single-process containers, but doesn't support images designed to run a full init system.

With this change, you can now run images like `jrei/systemd-ubuntu:22.04` and get a full Linux system experience:
- `systemctl` works
- `journalctl` works  
- SSH and other system services can run
- The VM feels like an EC2 instance

### Auto-detection
The mode is auto-detected from the image's CMD:
- If CMD is `/sbin/init`, `/lib/systemd/systemd`, or similar → **systemd mode**
- Otherwise → **exec mode** (existing behavior)

### Boot Flow
```
Kernel → init.sh (mount /proc /sys /dev) → Go init binary
                                              ↓
                                    ┌─────────┴─────────┐
                                    ↓                   ↓
                              Exec Mode           Systemd Mode
                                    ↓                   ↓
                            Run entrypoint      chroot + exec /sbin/init
                            as child process    (systemd becomes PID 1)
```

### Guest Agent
In systemd mode, the guest-agent is installed as a systemd service (`hypeman-agent.service`) that starts automatically. This enables `hypeman exec`, `hypeman cp`, and other remote operations.

## Key Changes

| File | Description |
|------|-------------|
| `lib/system/init/*.go` | New Go-based init binary with modular boot phases |
| `lib/system/init/init.sh` | Shell wrapper (Go runtime needs /proc /sys /dev before main()) |
| `lib/images/systemd.go` | `IsSystemdImage()` auto-detection from CMD |
| `lib/instances/configdisk.go` | Passes `INIT_MODE` to guest via config disk |
| `integration/systemd_test.go` | Full E2E test |

## Testing

```bash
# Run the systemd integration test
make test TEST=TestSystemdMode
```

The test:
1. Pulls `jrei/systemd-ubuntu:22.04`
2. Verifies `IsSystemdImage()` detects it correctly
3. Boots a VM
4. Waits for guest-agent to be ready
5. Verifies:
   - PID 1 is `systemd`
   - `/opt/hypeman/guest-agent` exists
   - `hypeman-agent.service` is active
   - `journalctl -u hypeman-agent` works

## Demo

```bash
# Run a systemd VM
hypeman run --name demo jrei/systemd-ubuntu:22.04

# Wait a few seconds for boot, then:
hypeman exec demo cat /proc/1/comm
# → systemd

hypeman exec demo systemctl status hypeman-agent
# → active (running)

hypeman exec demo journalctl -u hypeman-agent --no-pager -n 5
# → shows agent logs
```

---

> [!NOTE]
> Enables full systemd-based VMs and modernizes guest boot/init flow.
> 
> - New Go init (`lib/system/init/*`) with structured logging and dual modes: exec (default) and systemd (auto-detected via `images.IsSystemdImage`) with service injection for `hypeman-agent`
> - Replaces shell init script with embedded binaries (`InitBinary`, wrapper `init.sh`); initrd build updated to include these and NVIDIA modules; staleness hash logic adjusted
> - Host→guest configuration switched to JSON (`lib/vmconfig/`); config disk now writes `config.json` and includes network, volumes, GPU, and `init_mode`
> - Exec API adds `wait_for_agent` and guest client adds retryable vsock dialing (`AgentVSockDialError`) and Unavailable handling
> - Makefile builds embedded `guest-agent` and new `init`; build/dev/test/release-prep depend on these; clean targets updated; .gitignore includes built init
> - Tests: new `integration/systemd_test.go`; unit tests for `IsSystemdImage`; exec tests use `WaitForAgent` and updated log assertions
> - Docs: `lib/system/README.md` updated to describe Go init, storage/versioning, and workflows
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0dd309854ce601f64537779d3a7cd1b60b445864. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.cursor/plans/systemd_vm_with_go_init_c0a9c010.plan.md
```diff
@@ -0,0 +1,418 @@
+---
```
rgarcia: will remove before merge

lib/system/init/main.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Info("mode", "entering systemd mode")
+		runSystemdMode(log, cfg)
+	} else {
+		log.Info("mode", "entering exec mode")
+		runExecMode(log, cfg)
+	}
+}
+
+// dropToShell drops to an interactive shell for debugging when boot fails
+func dropToShell() {
```
rgarcia, line 88: not sure if this is actually useful... opus tooks some liberties

lib/images/systemd.go
```diff
@@ -0,0 +1,43 @@
...
+
+// IsSystemdImage checks if the image's CMD indicates it wants systemd as init.
+// Detection is based on the effective command (entrypoint + cmd), not whether
+// systemd is installed in the image.
+//
+// Returns true if the image's command is:
+//   - /sbin/init
+//   - /lib/systemd/systemd
+//   - /usr/lib/systemd/systemd
+//   - Any path ending in /init
```
sjmiller609: Maybe any path ending in `/init` is too broad to decide it's a systemd image. I think it's not that uncommon to name an entrypoint script like that

lib/system/init/config.go
```diff
@@ -0,0 +1,194 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Config holds the parsed configuration from the config disk.
+type Config struct {
```
sjmiller609: should this be imported from somewhere? seems like this type is already defined wherever we serialize it into the file.

lib/system/init/config.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, fmt.Errorf("parse config: %w", err)
+	}
+	log.Info("config", "parsed configuration")
+
+	return cfg, nil
+}
+
+// parseConfigFile parses a shell-style configuration file.
+// It handles simple KEY=VALUE and KEY="VALUE" assignments.
+func parseConfigFile(path string) (*Config, error) {
```
sjmiller609: would it be simpler to just change to writing json to the file instead of env list? Since maybe it's only an env list as a left over artifact from being used as "source" but now if it's go writing and go reading, might be cleaner to just define the config struct in one place and use normal json serialization / deserialization.

lib/system/init/config.go
```diff
@@ -0,0 +1,194 @@
...
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+// parseVolumeMounts parses the VOLUME_MOUNTS string.
+// Format: "device:path:mode[:overlay_device] device:path:mode ..."
+func parseVolumeMounts(s string) []VolumeMount {
```
sjmiller609: Yeah, this could be way simplified I would think if we go with json serialization, define types more cleanly assuming they are being json serialized instead of kinda jamming it into env list then pulling back out from that.

lib/system/init/init.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+# Minimal init wrapper that sets up environment before running Go init
+# The Go runtime needs /proc and /dev to exist during initialization
```
sjmiller609, line 3: 👍

lib/system/init/main.go
sjmiller609: Maybe should log something, just so it's clear what's going on from `hypeman logs ...`

lib/system/init/mode_systemd.go
```diff
@@ -0,0 +1,90 @@
...
+		log.Error("systemd", "chdir / failed", err)
+		dropToShell()
+	}
+
+	// Exec systemd - this replaces the current process
+	log.Info("systemd", "exec /sbin/init")
+
+	// syscall.Exec replaces the current process with the new one
+	// /sbin/init is typically a symlink to /lib/systemd/systemd
+	err := syscall.Exec("/sbin/init", []string{"/sbin/init"}, os.Environ())
```
sjmiller609: is it always `/sbin/init`? shouldn't we run the user's CMD anyways instead of override it to `/sbin/init`?

sjmiller609 (approved):
Looks great. One primary simplification tip I think would be nice is the serialize / deserialize config as json simplification, shared single type definition in host and guest code (imported in guest code)

PR #51, opened 2025-12-26
Resource accounting

sjmiller609:
> [!NOTE]
> Introduces host resource accounting and per-VM rate limiting across CPU, memory, disk, and network.
> 
> - Adds `lib/resources` (CPU/memory/disk/network discovery, oversubscription, capacity/allocations) and new `GET /resources` endpoint and types in `oapi`/`openapi.yaml`
> - Extends instance API/types to accept and report `network.bandwidth_{download,upload}` and `disk_io_bps`; stores in metadata and includes in responses
> - Computes default `disk_io_bps` and symmetric network limits from CPU share via `ResourceManager`
> - Implements bidirectional network shaping: TBF on TAP egress (download) and HTB classes on bridge (upload) with setup/cleanup; `NetworkManager.SetupHTB`, class derivation, orphan cleanup, and restore paths updated
> - Enables disk I/O throttling in hypervisors (Cloud Hypervisor/QEMU) and applies limits to all VM disks
> - Adds config for oversubscription ratios, burst multipliers, and hard capacity limits with validation; providers/wiring initialize `ResourceManager`
> - Images/volumes expose total sizes for disk accounting; broad test and API fixture updates; minor Makefile test verbosity flag
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 366e0f767cc506ef2c13da3ecddb3f47f17cd796. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cmd/api/api/instances.go
```diff
@@ -144,18 +190,36 @@ func (s *ApiService) CreateInstance(ctx context.Context, request oapi.CreateInst
...
+	}
+	if networkBandwidthDownload == 0 || networkBandwidthUpload == 0 {
+		defaultDown, defaultUp := s.ResourceManager.DefaultNetworkBandwidth(vcpus)
+		if networkBandwidthDownload == 0 {
+			networkBandwidthDownload = defaultDown
+		}
+		if networkBandwidthUpload == 0 {
+			networkBandwidthUpload = defaultUp
+		}
+	}
```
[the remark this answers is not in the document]
sjmiller609: never is nil

lib/resources/README.md
```diff
@@ -0,0 +1,132 @@
...
+- Smooth traffic shaping (queues packets, doesn't drop)
+
+**Uploads (VM → external):**
+- HTB (Hierarchical Token Bucket) on bridge egress
+- Per-VM classes with guaranteed rate and burst ceiling
+- Fair sharing when VMs contend for bandwidth
+- fq_codel leaf qdisc for low latency under load
+
+**Default limits:**
+- Proportional to CPU: `(vcpus / cpu_capacity) * network_capacity`
```
rgarcia, line 60: so if i have 1 vcpu vm on a 256 core machine, with 2.0 oversub and a 10Gbps NIC:

Limit = 10Gbps × 2.0 × (1/256) = ~78 Mbps download and upload

seems low? Maybe we just raise oversub

lib/resources/resource.go
```diff
@@ -0,0 +1,498 @@
...
+
+// ResourceStatus represents the current state of a resource type.
+type ResourceStatus struct {
+	Type           ResourceType `json:"type"`
+	Capacity       int64        `json:"capacity"`         // Raw host capacity
+	EffectiveLimit int64        `json:"effective_limit"`  // Capacity * oversubscription ratio
+	Allocated      int64        `json:"allocated"`        // Currently allocated
+	Available      int64        `json:"available"`        // EffectiveLimit - Allocated
+	OversubRatio   float64      `json:"oversub_ratio"`    // Oversubscription ratio applied
+	Source         string       `json:"source,omitempty"` // How capacity was determined (e.g., "detected", "configured")
```
rgarcia: enum for this similar to ResourceType?

lib/resources/network.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("parse network limit: %w", err)
+		}
+		capacity = parsed
+	} else {
+		// Auto-detect from uplink interface
+		uplink, err := getUplinkInterface(cfg.UplinkInterface)
+		if err != nil {
+			// No uplink found - network limiting disabled
+			capacity = 0
```
rgarcia, line 41: log a warning in these cases where we disable network limiting?

rgarcia (approved):

lib/resources/README.md
sjmiller609: Right, and yes I agree / ran into same concept here. But it seems to me that it's sensible to give a low limit by default if the CPU is super high - presumably this user would be planning to deploy 100 VMs on the machine. Also, I think in practice we will not use the default, it seems possibly too high, because the NIC speed is going to be much higher than the actual bandwidth

PR #56, opened 2026-01-12
refactor: remove build queue for immediate build execution

hiroTamada:
## Summary

- Remove the build queue - builds now start immediately when created
- Delete `lib/builds/queue.go` and `lib/builds/queue_test.go`
- Add `ErrResourcesExhausted` error and 503 response type for resource exhaustion
- Update OpenAPI spec with 503 response and deprecated `queue_position` field

## Motivation

The build queue added complexity without providing significant benefits:
- Builder VMs use similar resources to regular VMs
- Queueing delays builds when resources may be available
- Simpler architecture without queue state management

## Key Changes

| Aspect | Before | After |
|--------|--------|-------|
| Build start | Queued, starts when slot available | Starts immediately |
| Capacity exceeded | Queued with position | Build fails with resource error |
| Cancel queued | Removes from queue | N/A (no queue) |
| Recovery on restart | Re-enqueue pending | Start pending immediately |

## Files Changed

- **Deleted**: `lib/builds/queue.go`, `lib/builds/queue_test.go`
- **Modified**: `manager.go`, `types.go`, `errors.go`, `metrics.go`, `storage.go`
- **API**: Added 503 response to `openapi.yaml`, updated `cmd/api/api/builds.go`
- **Tests**: Updated `manager_test.go` to remove queue tests
- **Docs**: Updated `README.md` to reflect new architecture

## Test plan

- [x] `go build ./...` - project compiles
- [x] `go test ./lib/builds/...` - all builds package tests pass
- [ ] Manual testing: create a build and verify it starts immediately with `building` status
- [ ] Manual testing: verify build completion and logs streaming work correctly

---

> [!NOTE]
> Modernizes the build flow to execute immediately and surface capacity errors early.
> 
> - Remove build queue: delete `lib/builds/queue.go` and tests; initial status is `building`; recovery restarts pending builds directly
> - Add preflight resource check via `instances.CheckResourceAvailability`; introduce `ErrResourcesExhausted`; map to `503 Service Unavailable` with `Retry-After` in API (`cmd/api/...`, `oapi`); surface from manager and instance layer
> - OpenAPI/types: drop `queued` status, deprecate `queue_position`, change 202 description to “started”, and add 503 response; regenerate client/server code
> - Improve manager: avoid overwriting terminal states during updates/cancel; start builds in goroutine; persist builder instance ID; adjust cancel to terminate VM safely
> - Metrics: keep `hypeman_build_duration_seconds` and `hypeman_builds_total`; remove queue/active gauges
> - Docs/tests/tooling: update `README.md`, revise tests to expect immediate `building`, remove queue tests; enhance e2e script to use WebSocket exec (`websocat`); minor dep tidy (`go.mod`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 587b767ee57eae1988b2098807af63f460643e0b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/builds/manager.go
```diff
@@ -85,7 +85,6 @@ func DefaultConfig() Config {
 type manager struct {
 	config          Config
```
rgarcia, line 87: nit: `MaxConcurrentBuilds` in the config is now unused since the queue is removed. might be worth removing or deprecating it to avoid confusion for operators who set it expecting it to limit concurrency.

lib/builds/manager.go
```diff
@@ -386,6 +379,11 @@ func (m *manager) executeBuild(ctx context.Context, id string, req CreateBuildRe
 		},
 	})
 	if err != nil {
+		// Check if this is a resource exhaustion error
+		errStr := err.Error()
+		if strings.Contains(errStr, "exceeds") && strings.Contains(errStr, "limit") {
```
rgarcia, line 410: unclear how this `ErrResourcesExhausted` error makes it back to the API handler - since `executeBuild` is called from `runBuild` which runs in a background goroutine, the caller has already received a 202 by the time this runs. the 503 path in the handler seems unreachable?

rgarcia (approved):
lgtm

lib/builds/manager.go
hiroTamada: Good call - I've left `MaxConcurrentBuilds` in the config for now since it doesn't hurt anything, but agree it should be cleaned up. I'll add a follow-up to either remove it or repurpose it for the new preflight resource checking. Thanks!

lib/builds/manager.go
hiroTamada: You're absolutely right - that was exactly the issue. I've fixed this by adding a **synchronous preflight resource check** in `CreateBuild` that runs *before* spawning the async goroutine:

```go
// Preflight check: verify resources are available before accepting the build
builderMemory := int64(policy.MemoryMB) * 1024 * 1024
if err := m.instanceManager.CheckResourceAvailability(ctx, policy.CPUs, builderMemory); err != nil {
    if errors.Is(err, instances.ErrResourcesExhausted) {
        return nil, fmt.Errorf("%w: %v", ErrResourcesExhausted, err)
    }
    return nil, fmt.Errorf("check resource availability: %w", err)
}
```

Added a new `CheckResourceAvailability()` method to `instances.Manager` that checks per-instance and aggregate limits without actually creating an instance. Now the 503 path is reachable when resources are truly exhausted at build creation time.

Also added proper sentinel errors (`ErrResourcesExhausted`) to both packages so we can use `errors.Is()` instead of brittle string matching.

PR #58, opened 2026-01-14
feat: add script to build from source and support binary path with install.sh

tnsardesai:
This is used by 

---

> [!NOTE]
> Adds tooling for building and installing Hypeman with more flexible paths.
> 
> - **New `scripts/build-from-source.sh`**: Builds `hypeman-api` and `hypeman-token`, writes logs to `build.log`, validates absolute `OUTPUT_DIR`, and copies `.env.example` to the output dir
> - **Installer enhancements (`scripts/install.sh`)**: Adds `BINARY_DIR` mode to install from prebuilt binaries (with validation and exec perms), makes `BRANCH`, `VERSION`, and `BINARY_DIR` mutually exclusive with checks, and treats `CLI_VERSION=latest` as auto-resolve; keeps build-from-source and release-download paths intact
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ebef7e87aee0761bbc3cc5cd66cf9ec543011758. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/scripts/build-from-source.sh
+++ b/scripts/build-from-source.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+#
+# Hypeman Build from Source Script
+#
+# Usage:
+#   ./scripts/build-from-source.sh
+#
+# Options (via environment variables:
+#   OUTPUT_DIR   - Full path of directory to place built binaries (optional, default: $(pwd)/bin)
+#
+
+set -e
+
+# Default values
+BINARY_NAME="hypeman-api"
+
+# Colors for output (true color)
+RED='\033[38;2;255;110;110m'
+GREEN='\033[38;2;92;190;83m'
+YELLOW='\033[0;33m'
+PURPLE='\033[38;2;172;134;249m'
+NC='\033[0m' # No Color
+
+info() { echo -e "${GREEN}[INFO]${NC} $1"; }
+warn() { echo -e "${YELLOW}[WARN]${NC} $1"; }
+error() { echo -e "${RED}[ERROR]${NC} $1"; exit 1; }
+
+# =============================================================================
+# Pre-flight checks - verify all requirements before doing anything
+# =============================================================================
+
+# Check for required commands
+command -v go >/dev/null 2>&1 || error "go is required but not installed"
+command -v make >/dev/null 2>&1 || error "make is required but not installed"
+
+# =============================================================================
+# Setup directories
+# =============================================================================
+
+# Get the directory where this script is located
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SOURCE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [ -z "$OUTPUT_DIR" ]; then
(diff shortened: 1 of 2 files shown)


scripts/build-from-source.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+#
+# Hypeman Build from Source Script
+#
+# Usage:
+#   ./scripts/build-from-source.sh
+#
+# Options (via environment variables:
```
Sayan-: nit: missing closing parenthesis - should be `(via environment variables):`

scripts/build-from-source.sh
```diff
@@ -0,0 +1,83 @@
...
+#   OUTPUT_DIR   - Full path of directory to place built binaries (optional, default: $(pwd)/bin)
```
Sayan-: nit: comment says `$(pwd)/bin` but the actual default is `${SOURCE_DIR}/bin` (the repo root's bin dir, not cwd). might want to clarify.

scripts/install.sh
```diff
@@ -99,13 +100,30 @@ command -v systemctl >/dev/null 2>&1 || error "systemctl is required but not ins
 command -v setcap >/dev/null 2>&1 || error "setcap is required but not installed (install libcap2-bin)"
 command -v openssl >/dev/null 2>&1 || error "openssl is required but not installed"
 
+# Count how many of BRANCH, VERSION, BINARY_DIR are set
+count=0
+[ -n "$BRANCH" ] && ((count++)) || true
```
Sayan-, line 105: heads up: if you adopt the bot's suggestion for `set -u`, these `[ -n "$BRANCH" ]` checks will fail on unset vars. would need `${BRANCH:-}` syntax.

Sayan- (commented):
bot comments seem relevant - particularly the file validation before `cp` in the BINARY_DIR path.

separately, what's the relation between `install.sh` and `build-from-source.sh`? the build logic (make build + go build gen-jwt + copy .env.example) now exists in both scripts. wondering if BRANCH mode in install.sh could delegate to build-from-source.sh to reduce duplication.

tnsardesai:
>separately, what's the relation between install.sh and build-from-source.sh? the build logic (make build + go build gen-jwt + copy .env.example) now exists in both scripts. wondering if BRANCH mode in install.sh could delegate to build-from-source.sh to reduce duplication.

that is actually how I had it originally but then noticed that we suggest `curl -fsSL https://raw.githubusercontent.com/kernel/hypeman/main/scripts/install.sh | bash`. If we delegate to build-from-source.sh then we are breaking curl-pipe mode to run the installer. Also the `BRANCH` mode clones the repo but the build-from-source.sh is meant to be run in the clone repo and not via curl-pipe. 

I was thinking if `build-from-source.sh` should just be a make command 🤔. what do you think? @Sayan-

scripts/build-from-source.sh
tnsardesai: :nod: yep I changed the default but forgot to update comment

Sayan- (approved):
:ship-it:

PR #63, opened 2026-01-20
fix(caddy): isolate routes by listen port to prevent wildcard conflicts

hiroTamada:
## Summary

Fixes wildcard ingress conflicts when multiple wildcard ingresses match the same hostname pattern on different ports.

### Problem

Previously, all ingress routes were placed in a single Caddy server with multiple listen addresses. When two wildcard ingresses had the same hostname pattern but different ports (e.g., `*.host.kernel.sh:443` → port 80 and `*.host.kernel.sh:3000` → port 3000), Caddy would always match the first route regardless of which port the request came in on.

**Example of the bug:**
- Request to `https://my-app.host.kernel.sh:3000/` would match the first wildcard route (port 443 → 80) instead of the correct route (port 3000 → 3000)
- Result: 502 error because the app is listening on port 3000, not port 80

### Solution

Create separate Caddy servers for each unique listen port, isolating routes so they only match requests on their intended port.

**Before (single server):**
```json
{
  "servers": {
    "ingress": {
      "listen": [":443", ":3000"],
      "routes": [
        { "match": "*.host.kernel.sh", "target_port": 80 },   // Always wins
        { "match": "*.host.kernel.sh", "target_port": 3000 }  // Never reached
      ]
    }
  }
}
```

**After (port-isolated servers):**
```json
{
  "servers": {
    "ingress-443": {
      "listen": [":443"],
      "routes": [{ "match": "*.host.kernel.sh", "target_port": 80 }]
    },
    "ingress-3000": {
      "listen": [":3000"],
      "routes": [{ "match": "*.host.kernel.sh", "target_port": 3000 }]
    }
  }
}
```

## Changes

- Refactored `buildConfig()` to group routes by listen port
- Each unique listen port gets its own Caddy server
- HTTP redirect routes go to the port 80 server
- Added `TestGenerateConfig_PortIsolation` test
- Updated existing tests for new multi-server architecture

## Test plan

- [x] All existing tests pass
- [x] New `TestGenerateConfig_PortIsolation` test verifies isolated routing
- [ ] Manual testing on staging with wildcard ingresses on different ports

---

> [!NOTE]
> Implements port-isolated Caddy servers to eliminate wildcard hostname conflicts across different listen ports.
> 
> - Refactors `buildConfig()` to group routes by port and build one server per port (`servers` → `ingress-<port>`), with per-port `automatic_https` (disable redirects on TLS ports; fully disable on non-TLS ports)
> - Sends HTTP redirect routes to the port `80` server; prepends API ingress/redirect to the correct server based on TLS
> - Adds per-server catch-all 404 route; disables access logs in each server
> - Tracks TLS hostnames per rule and deduplicates subjects for TLS automation when ACME is configured
> - Updates existing tests for multi-server output and adds `TestGenerateConfig_PortIsolation` to verify isolated routing
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 76fd373740b985fc866bdd409d68fbc4bb7c8af9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/ingress/config.go
+++ b/lib/ingress/config.go
@@ -205,6 +205,7 @@ func (g *CaddyConfigGenerator) buildConfig(ctx context.Context, ingresses []Ingr
 	routesByPort := map[int][]interface{}{}
 	tlsHostnames := []string{}
 	tlsPortsByHostname := map[string][]int{} // Track which ports need TLS for each hostname
+	tlsEnabledPorts := map[int]bool{}        // Track which ports have at least one TLS route
 
 	for _, ingress := range ingresses {
 		for _, rule := range ingress.Rules {
@@ -269,6 +270,7 @@ func (g *CaddyConfigGenerator) buildConfig(ctx context.Context, ingresses []Ingr
 			if rule.TLS {
 				tlsHostnames = append(tlsHostnames, hostnameMatch)
 				tlsPortsByHostname[hostnameMatch] = append(tlsPortsByHostname[hostnameMatch], port)
+				tlsEnabledPorts[port] = true
 
 				// Add HTTP redirect route if requested
 				// These go to port 80 server
@@ -326,6 +328,7 @@ func (g *CaddyConfigGenerator) buildConfig(ctx context.Context, ingresses []Ingr
 			apiListenPort = 443
 			tlsHostnames = append(tlsHostnames, g.apiIngress.Hostname)
 			tlsPortsByHostname[g.apiIngress.Hostname] = append(tlsPortsByHostname[g.apiIngress.Hostname], 443)
+			tlsEnabledPorts[443] = true
 
 			// Add HTTP to HTTPS redirect for API hostname
 			if g.apiIngress.RedirectHTTP {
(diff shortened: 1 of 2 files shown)


hiroTamada:
## Deployment Notes

**No migration required.** This change is safe to deploy directly.

### Why no migration is needed:

1. **Data model unchanged** - The ingress rules stored in the database remain the same. We're only changing how the Caddy config is *generated* from that data.

2. **Automatic config regeneration** - When hypeman restarts, it regenerates the Caddy config from existing ingresses. The new code will produce port-isolated servers automatically.

3. **Valid Caddy config** - The new format (multiple servers instead of one) is valid Caddy JSON that Caddy will accept on reload.

### What changes on deploy:

```
Before restart:
  servers: { "ingress": { listen: [":443", ":3000", ":80"] } }

After restart:
  servers: { 
    "ingress-443": { listen: [":443"] },
    "ingress-3000": { listen: [":3000"] },
    "ingress-80": { listen: [":80"] }
  }
```

### Pre-deployment checklist:

- [ ] Verify no stale/conflicting wildcard ingresses exist on the target server:
  ```bash
  export HYPEMAN_API_KEY=$(/opt/hypeman/bin/hypeman-token)
  /opt/hypeman/bin/hypeman-cli ingress list
  ```
- [ ] If conflicting ingresses exist (e.g., old `wildcard-ingress`), delete them first

hiroTamada:
## Response to Review Comments

### 1. Automatic HTTPS not disabled for non-TLS ports

**Fixed in commit 76fd373.** The second commit addresses this by tracking `tlsEnabledPorts` and setting `disable: true` for any port without TLS routes, not just port 80.

### 2. Redirect routes no longer guaranteed to precede regular routes

This is a valid observation, but I believe it's **low risk** in practice:

1. **Redirects have a more specific matcher** - They include `protocol: http`, so they only match HTTP requests
2. **Hostname conflicts are validated** - The manager prevents creating ingresses with the same hostname+port, so you can't have both a TLS redirect and a non-TLS route for the exact same hostname on port 80
3. **Wildcard edge case** - If you have `*.example.com` with TLS redirect AND `api.example.com` on HTTP port 80, the order matters. But this seems like a deliberate configuration choice.

If we want stricter guarantees, I can refactor to track redirect routes separately per port and prepend them when building the final routes. Let me know if you'd like that change.

sjmiller609 (approved):
clutch

PR #64, opened 2026-01-21
Fix stop / start instance with network

sjmiller609:
> [!NOTE]
> Enables stop/start of instances with networking by refining name-uniqueness checks and aligning tests with network initialization.
> 
> - Updates `Manager` interface and implementation: `NameExists(ctx, name, excludeInstanceID)`; skips the provided instance ID during checks
> - Modifies `CreateAllocation` to use the new signature so restarts don't conflict with the instance’s own name
> - Test harness now configures network (`BridgeName`, `SubnetCIDR`, `DNSServer`) and calls `NetworkManager.Initialize(...)`; `ApiService` includes `NetworkManager`
> - Lifecycle test runs with `network.enabled=true` to cover stop/start with networking
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cbf891d79f88064b5674104dc64e9e6cff14a3cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/api/api/instances_test.go
+++ b/cmd/api/api/instances_test.go
@@ -148,7 +148,7 @@ func TestInstanceLifecycle_StopStart(t *testing.T) {
 
 	// 1. Create instance
 	t.Log("Creating instance...")
-	networkEnabled := false
+	networkEnabled := true
 	createResp, err := svc.CreateInstance(ctx(), oapi.CreateInstanceRequestObject{
 		Body: &oapi.CreateInstanceRequest{
 			Name:  "test-lifecycle",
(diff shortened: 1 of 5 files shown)


cmd/api/api/instances_test.go
```diff
@@ -148,7 +148,7 @@ func TestInstanceLifecycle_StopStart(t *testing.T) {
 
 	// 1. Create instance
 	t.Log("Creating instance...")
-	networkEnabled := false
+	networkEnabled := true
```
sjmiller609, line 151: stop / start was failing with network enabled, so enabling network in the test so it would catch that

hiroTamada (approved):
clean fix for the stop/start name collision issue. excluding the instance from its own uniqueness check is the right approach.

PR #71, opened 2026-01-27
Fix image name parsing

sjmiller609:
I was having an issue with "hypeman push" when run from remote server, which I wanted to use for testing the gpu related changes.

> [!NOTE]
> Strengthens image/repo name handling and registry auth.
> 
> - Replace regex-based repo extraction with docker/distribution `v2.Router()` in `oapi_auth.go` to correctly parse `/v2/{name}/...` where `name` may contain slashes; update `extractRepoFromPath` accordingly
> - Add comprehensive tests: URL-decoding behavior of `oapi-codegen` runtime, Chi routing with URL-encoded slashes, and repo extraction across manifests/blobs/uploads paths; expand auth tests to reject registry-scoped tokens and edge cases
> - Update dependencies (`docker/distribution`, `gorilla/mux`, `opencontainers/go-digest`, `google/uuid`) to direct requirements
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2fc0043261f1cdd108aebef93b4c5f88249d7ad1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/middleware/oapi_auth.go
```diff
@@ -201,10 +203,21 @@ func isInternalVMRequest(r *http.Request) bool {
...
+	req, err := http.NewRequest(http.MethodGet, path, nil)
+	if err != nil {
+		return ""
+	}
+
+	var match mux.RouteMatch
+	if registryRouter.Match(req, &match) {
+		if name, ok := match.Vars["name"]; ok {
+			return name
+		}
```
sjmiller609, line 220: This is using the dependency to parse instead of regex. This problem could also be fixed by patching the regex but I figured to just use the dependency even though this seems a bit weird to use it by constructing a request then using methods on the request type, but that seems ok to me.

hiroTamada (approved):
looks good - using docker/distribution's router to parse registry paths is a solid approach that handles nested repo names correctly. good test coverage.

PR #80, opened 2026-01-30
feat(api): add POST /volumes/from-archive endpoint for multipart uploads

hiroTamada:
## Summary

Split volume creation into two separate endpoints so Stainless can generate proper SDK methods for both use cases:

- `POST /volumes` (JSON body) - creates an empty volume of specified size
- `POST /volumes/from-archive` (multipart/form-data) - creates a volume pre-populated from a tar.gz archive

### Problem

The previous OpenAPI spec defined `POST /volumes` with both `application/json` and `multipart/form-data` content types. When Stainless generates SDKs from specs with multiple content types on the same endpoint, it picks one (typically JSON) and ignores the others. This resulted in:

- `VolumeNewParams` only having JSON fields (`name`, `size_gb`, `id`)
- No `Content io.Reader` field for the archive binary
- No `MarshalMultipart()` method generated

This forced client code (like Kernel API) to manually construct multipart requests for "create volume from archive" functionality.

### Solution

By splitting into two endpoints, Stainless will generate:

**Go SDK:**
- `client.Volumes.New()` - JSON body, for empty volumes
- `client.Volumes.NewFromArchive()` - multipart with `Content io.Reader` field

This mirrors how `client.Builds.New()` already works with its `Source io.Reader` field.

### Changes

1. **openapi.yaml**: 
   - Simplified `POST /volumes` to JSON-only
   - Added new `POST /volumes/from-archive` endpoint with multipart-only schema

2. **stainless.yaml**:
   - Added `create_from_archive: post /volumes/from-archive` method mapping

## Server-side changes needed

The Hypeman server will need to add a handler for `POST /volumes/from-archive` that performs the same logic as the current multipart handling in `POST /volumes`. The existing `POST /volumes` can be simplified to only handle JSON bodies.

## Test plan

- [x] Verify Stainless regenerates SDKs with the new endpoint
- [x] Confirm Go SDK has `VolumeService.NewFromArchive()` with `Content io.Reader`
- [ ] Update Hypeman server to handle new route
- [ ] Update Kernel API to use new SDK method instead of manual multipart construction

---

> [!NOTE]
> **Medium Risk**
> Changes the public API surface (request/route split) and regenerates large portions of OpenAPI-generated server/client code, which can introduce subtle compatibility or routing/parsing regressions despite limited underlying business logic changes.
> 
> **Overview**
> **Splits volume creation into two endpoints**: `POST /volumes` now accepts *only* JSON to create an empty volume, and a new `POST /volumes/from-archive` streams a tar.gz request body (with `name`, `size_gb`, optional `id` as query params) to create a pre-populated volume.
> 
> Updates the API implementation and generated `lib/oapi` server/client bindings to remove multipart handling from `CreateVolume`, add `CreateVolumeFromArchive` request/response types, routing, and client helpers/parsers, and adjusts unit tests to the new request shape.
> 
> Refreshes `openapi.yaml` and embedded swagger and updates `stainless.yaml` repo/package settings plus a new `create_from_archive` method mapping for SDK generation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2466fb79299de9aeffe11e27189ca04e5bf5e181. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/api/api/volumes.go
+++ b/cmd/api/api/volumes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"mime/multipart"
 	"strconv"
 
 	"github.com/kernel/hypeman/lib/logger"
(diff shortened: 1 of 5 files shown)


cmd/api/api/volumes.go
```diff
@@ -13,6 +14,112 @@ import (
...
+			if form.ContentFile != nil {
+				form.Close()
+				return nil, &formError{Code: "invalid_form", Message: "duplicate content field"}
+			}
+			// Buffer content to a temp file to handle any field order
+			tempFile, err := os.CreateTemp("", "volume-archive-*.tar.gz")
+			if err != nil {
+				form.Close()
+				return nil, &formError{Code: "internal_error", Message: "failed to create temp file"}
+			}
```
sjmiller609: Before this change, we have the whole archive being streamed directly into the volume. But this introduces an intermediate write to disk. So it instead works like stream into a local file, then copy from the file into the volume. That is slower in comparison to streaming from the HTTP request into the volume without an intermediate step.

sjmiller609: The problem with the original approach is that it assumed an order of the fields arriving but multi part HTTP uploads don't necessarily always come in the same field order. So I think part of this fix is to get all the fields before deciding what to do next. However, I think this approach is susceptible to arbitrarily large uploads so we need to think about that.

sjmiller609: I think it's a better approach to put name and size_gb in the query parameters for the post request. 

```
/volumes/from-archive:
  post:
    parameters:
      - name: name
        in: query
        required: true
        schema:
          type: string
      - name: size_gb
        in: query
        required: true
        schema:
          type: integer
    requestBody:
      required: true
      content:
        application/gzip:
          schema:
            type: string
            format: binary
```

this allows the information for size_gb and name to exist when it's needed. we can still stream the data into the volume directly from HTTP request.

Also I think we should add a new environment variable MAX_VOLUME_SIZE used to control the max permitted size for any individual volume, and respect this when creating volumes, also respecting the existing MAX_TOTAL_VOLUME_STORAGE, which is the aggregated total volume storage

cmd/api/api/volumes.go
```diff
@@ -13,6 +14,112 @@ import (
...
+	if f.ContentFile != nil {
+		f.ContentFile.Close()
+		os.Remove(f.ContentFile.Name())
+	}
+}
+
+// parseVolumeMultipartForm parses a multipart form for volume creation.
+// It buffers the content field to a temp file to handle any field order.
+// Caller must call form.Close() to clean up temp files.
+func parseVolumeMultipartForm(multipartReader *multipart.Reader) (*volumeMultipartForm, error) {
```
sjmiller609: I think somewhat complicated logic should always live down in the services layer instead of in handler code files. handler code files should just be for mapping from API to domain types and error mapping, and things like that.

cmd/api/api/volumes.go
```diff
@@ -77,125 +185,168 @@ func (s *ApiService) CreateVolume(ctx context.Context, request oapi.CreateVolume
 	}, nil
 }
 
-// createVolumeFromMultipart handles creating a volume from multipart form data with archive content
-func (s *ApiService) createVolumeFromMultipart(ctx context.Context, multipartReader *multipart.Reader) (oapi.CreateVolumeResponseObject, error) {
+// createVolumeFromMultipartDeprecated handles the deprecated multipart form on POST /volumes
+// New integrations should use POST /volumes/from-archive instead
+func (s *ApiService) createVolumeFromMultipartDeprecated(ctx context.Context, multipartReader *multipart.Reader) (oapi.CreateVolumeResponseObject, error) {
```
sjmiller609: Instead of deprecating it, just delete it I think.

openapi.yaml
```diff
@@ -1809,7 +1809,8 @@ paths:
       description: |
         Creates a new volume. Supports two modes:
         - JSON body: Creates an empty volume of the specified size
-        - Multipart form: Creates a volume pre-populated with content from a tar.gz archive
+        - Multipart form (DEPRECATED): Creates a volume from a tar.gz archive.
+          Use POST /volumes/from-archive instead for new integrations.
```
sjmiller609: we should just delete it instead of deprecating it

sjmiller609 (changes requested):
I think we can do this without an intermediate file copy, and also I realized there is a pre-existing issue with allowing arbitrarily large volumes to be created but we probably want to limit that.

cmd/api/api/volumes.go
hiroTamada: it is technically a breaking change in staging and prod (for kernel org) if we simply delete this

hiroTamada: but I guess that is fine

sjmiller609 (approved):

PR #81, opened 2026-01-30
fix: resource limits for starting instances

sjmiller609:
- Use the lib/resources module for checking if a starting or creating VM should be accepted
- Clean up mdev for stopped instances
- Fail on standby request if vGPU is enabled
- include disk io in resources API, accidentally omitted previously

---

> [!NOTE]
> **Medium Risk**
> Touches core instance lifecycle (create/start/stop/restore/standby) and resource admission logic, which can prevent workloads from starting if misconfigured; changes are largely additive and surfaced via explicit 409 errors.
> 
> **Overview**
> **Instance admission now uses `lib/resources` for aggregate capacity checks.** `instances.Manager` gains a pluggable `ResourceValidator` (wired in `cmd/api/main.go`) and `CreateInstance`, `StartInstance`, and `RestoreInstance` reject requests when CPU/memory/network/disk I/O/GPU capacity is insufficient, surfacing `ErrInsufficientResources`.
> 
> **API behavior changes:** `POST /instances` now returns `409` for insufficient resources, `StartInstance` returns `409` with `insufficient_resources`, and the OpenAPI client/spec are updated accordingly. The resources endpoint now includes disk I/O capacity/status and per-instance `disk_io_bps` allocations.
> 
> **vGPU lifecycle and state rules tightened:** standby is blocked for vGPU instances, vGPU mdevs are recreated on start and destroyed/cleared on stop, and allocations tracking now includes `DiskIOBps`. Legacy aggregate CPU/memory env limits and related instance-side aggregate limit code/tests are removed in favor of oversubscription ratios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca5e720a89022a93fb1ce2d5738c834965e3ad99. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cmd/api/config/config.go
```diff
@@ -166,9 +165,8 @@ func Load() *Config {
 		MaxVcpusPerInstance:  getEnvInt("MAX_VCPUS_PER_INSTANCE", 16),
 		MaxMemoryPerInstance: getEnv("MAX_MEMORY_PER_INSTANCE", "32GB"),
 
-		// Resource limits - aggregate (0 or empty = unlimited)
-		MaxTotalVcpus:         getEnvInt("MAX_TOTAL_VCPUS", 0),
-		MaxTotalMemory:        getEnv("MAX_TOTAL_MEMORY", ""),
```
sjmiller609, line 171: the old way to account for host limits was just setting these, but now we have the resources module which works for more resources types and does host discovery

hiroTamada (approved):
Looks good overall - nice refactor to centralize resource validation in `lib/resources` instead of the ad-hoc aggregate checks.

Agreeing with Bugbot's findings that should be addressed (can be follow-up PRs):

1. **High: Missing network/diskIO limits on start** - `startInstance` passes `0, 0, 0` for network download/upload and disk I/O instead of the stored values. Restarted instances won't have their bandwidth validated against capacity.

2. **Medium: RestoreInstance missing validation** - `restoreInstance` doesn't call `validateResourceAllocation`, allowing standby→running transitions to bypass capacity checks.

These are edge cases (start from stopped, restore from standby) but could lead to oversubscription if resource availability changed while the instance was down.

PR #89, opened 2026-02-10
refactor: cross-platform foundation for macOS support

rgarcia:
## Summary

- Split platform-specific code into `_linux.go` and `_darwin.go` files across `resources`, `network`, `devices`, `ingress`, `vmm`, and `vm_metrics` packages
- Add hypervisor abstraction with registration pattern (`RegisterSocketName`, `RegisterVsockDialerFactory`, `RegisterClientFactory`) to decouple instance management from specific hypervisor implementations
- Add `"vz"` to the OpenAPI hypervisor type enum, erofs disk format support, and insecure registry option for builds
- Cross-compile guest-agent and init binaries for Linux (needed when building on macOS)

No behavioral changes on Linux. macOS can now compile but has no VM functionality yet.

This is part 1 of 2 for macOS/Virtualization.framework support. Part 2 (`feat/vz-hypervisor`) adds the actual vz hypervisor implementation.

## Test plan

- [x] `make build-linux` succeeds
- [x] `make test-linux` passes (failures are Docker Hub rate limiting, not code changes)
- [ ] CI passes on Linux runner
- [ ] CI passes on macOS runner (compilation check via `test-darwin`)

---

> [!NOTE]
> **Medium Risk**
> Touches VM lifecycle/build pipeline (vsock connection path, builder image preparation, log persistence) and image/disk creation logic, which could affect build reliability and instance startup even on Linux despite being largely additive/guarded.
> 
> **Overview**
> Establishes a cross-platform foundation by splitting multiple subsystems into Linux vs macOS implementations (networking, devices/GPU passthrough, resources, VM metrics, VMM/ingress binaries), adding macOS stubs/no-ops where features aren’t supported yet, and updating the API hypervisor enum to include `vz`.
> 
> Refactors hypervisor integration to be registration/factory-based (`RegisterClientFactory`, platform-specific starters, per-instance `GetVsockDialer`) and adds platform-specific hypervisor access checks (KVM on Linux vs Virtualization.framework on Apple Silicon).
> 
> Improves the build system: builder-agent now streams logs over vsock while retaining an authoritative buffered log; build manager can bootstrap a builder image locally from an embedded Dockerfile via Docker + OCI cache import when `BUILDER_IMAGE` is unset, gates build execution until the builder image is ready, and switches to per-instance vsock dialing. Image handling is adjusted to always target Linux VM platform for pulls/manifests and disk image creation is made 4K-sector aligned for Virtualization.framework compatibility.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 964eacb4eb80b9f57b09f07d256c7c031cba18f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/resources/cpu_darwin.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"runtime"
+)
+
+// detectCPUCapacity returns the number of logical CPUs on macOS.
+// Uses runtime.NumCPU() which calls sysctl on macOS.
+func detectCPUCapacity() (int64, error) {
+	return int64(runtime.NumCPU()), nil
+}
```
sjmiller609, line 13: this looks simpler than the linux one, idk if we could do the same there or not

lib/resources/disk_darwin.go
```diff
@@ -0,0 +1,49 @@
...
+		var ds datasize.ByteSize
+		if err := ds.UnmarshalText([]byte(cfg.DiskLimit)); err != nil {
+			return nil, err
+		}
+		capacity = int64(ds.Bytes())
+	} else {
+		// Auto-detect from filesystem using statfs
+		var stat unix.Statfs_t
+		dataDir := cfg.DataDir
+		if err := unix.Statfs(dataDir, &stat); err != nil {
```
sjmiller609: the implementations for darwin and linux look almost identical here

sjmiller609 (approved):
looks good, maybe some opportunity for deduplication

rgarcia:
Addressed the remaining review feedback in 3658218:

**Cursorbot bugs (all fixed):**
- **RecoverPendingBuilds race** — Moved `RecoverPendingBuilds()` from `NewManager()` to `Start()`, running it sequentially after `ensureBuilderImage` completes. Recovered builds no longer race against builder readiness.
- **Missing erofs sector alignment** — Added the same `alignToSector` + truncate logic that ext4 already had to `convertToErofs`.
- **Redundant ForLinux aliases** — Removed `InspectManifestForLinux`/`PullAndUnpackForLinux` and updated the single caller in `initrd.go`.

**Steven's observations:**
- **cpu_darwin.go simplicity** — Good observation. Linux could potentially use `runtime.NumCPU()` too, but the `/proc/cpuinfo` parsing handles edge cases like hyperthreading topology (siblings × sockets) which `runtime.NumCPU()` may not match on certain server configurations. Worth revisiting as a follow-up but not blocking.
- **disk_darwin.go duplication** — Agreed, the implementations are nearly identical. The difference is Darwin uses `unix.Statfs_t` with a `/` fallback while Linux uses `syscall.Statfs_t` without. These could be unified using `golang.org/x/sys/unix` on both platforms in a follow-up refactor.

Additionally, this commit includes the builder image rework:
- Embedded the Dockerfile with `go:embed` instead of reading from filesystem
- Removed the hardcoded `"hypeman/builder:latest"` default — if `BUILDER_IMAGE` is unset, builds from the embedded Dockerfile
- After `docker build` + `docker save`, writes the image directly into the OCI layout cache and calls `ImportLocalImage`, bypassing `docker push` entirely

PR #90, opened 2026-02-10
feat: add macOS VM support via Apple Virtualization.framework

rgarcia:
## Summary

- Add vz hypervisor implementation using Apple's Virtualization.framework via a codesigned subprocess (`vz-shim`)
- vsock-based guest communication, shared directory mounts for disk access, macOS-native networking via vmnet
- Makefile targets: `build-darwin`, `test-darwin`, `dev-darwin`, `sign-darwin`
- CI: macOS runner for `test-darwin`
- `scripts/install.sh` and `scripts/uninstall.sh` with macOS support (launchd, Homebrew PATH, codesign)

**Depends on** #89 (cross-platform foundation) — merge that first, then rebase this onto `main`.

## Test plan

- [ ] CI passes on Linux runner (no regressions)
- [ ] CI passes on macOS runner (`test-darwin`)
- [ ] `make build-darwin && make sign-darwin` succeeds on macOS
- [ ] E2E install test passes on macOS (`scripts/e2e-install-test.sh`)

---

> [!NOTE]
> **High Risk**
> Introduces a new hypervisor backend and a privileged/codesigned subprocess plus platform-specific install/service management, which can impact VM lifecycle, guest connectivity, and production deployments on macOS.
> 
> **Overview**
> Adds a new macOS-only `vz` hypervisor implementation that runs VMs in a separate `vz-shim` subprocess, exposes a Cloud-Hypervisor-like control API over a Unix socket, and proxies vsock connections via a per-VM `vz.vsock` socket.
> 
> Updates instance creation/boot defaults for `vz` (vsock socket path + `console=hvc0`) and routes API operations (`exec`, `cp`, `stat`) through `InstanceManager.GetVsockDialer` to support multiple hypervisors consistently. Guest init now always overwrites the guest-agent binary to avoid stale/corrupt binaries after restarts.
> 
> Adds macOS developer and deployment plumbing: new `.env.darwin.example`, `.air.darwin.toml`, Makefile targets for building/signing (`build-darwin`, `sign-darwin`, `dev-darwin`, `sign-vz-shim`), a self-hosted macOS CI job plus an E2E install test, and expanded `scripts/install.sh`/`scripts/uninstall.sh` to support macOS (launchd service, codesigning, macOS paths, CLI artifact differences).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 07550edb24633b9b79b838fa6519c283a45460c8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


rgarcia:
@cursor push 1353ca63cc

cmd/vz-shim/main.go
```diff
@@ -0,0 +1,163 @@
...
+	if err := os.MkdirAll(filepath.Dir(logPath), 0755); err != nil {
+		return fmt.Errorf("create log directory: %w", err)
+	}
+
+	file, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+	if err != nil {
+		return fmt.Errorf("open log file: %w", err)
+	}
+
+	slog.SetDefault(slog.New(slog.NewJSONHandler(file, &slog.HandlerOptions{Level: slog.LevelDebug})))
```
sjmiller609, line 163: is this what's ending up in the vmm.log file?

cmd/vz-shim/server.go
```diff
@@ -0,0 +1,260 @@
...
+		conn, err := listener.Accept()
+		if err != nil {
+			slog.Debug("vsock listener closed", "error", err)
+			return
+		}
+		go s.handleVsockConnection(conn)
+	}
+}
+
+func (s *ShimServer) handleVsockConnection(conn net.Conn) {
```
sjmiller609, line 194: an integration test checking that exec and graceful shutdown works would cover this well

cmd/vz-shim/vm.go
```diff
@@ -0,0 +1,276 @@
...
+	if virtualCPUCount < minAllowed {
+		virtualCPUCount = minAllowed
+	}
+
+	return virtualCPUCount
+}
+
+func computeMemorySize(requested uint64) uint64 {
+	if requested == 0 {
+		requested = 2 * 1024 * 1024 * 1024 // 2GB default
```
sjmiller609: nit: seems possibly magic, idk if we should just fail when memory isn't requested or if shared config somewhere else

lib/hypervisor/vz/client.go
```diff
@@ -0,0 +1,172 @@
...
+// NewClient creates a new vz shim client.
+func NewClient(socketPath string) (*Client, error) {
+	transport := &http.Transport{
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return net.Dial("unix", socketPath)
+		},
+	}
+	httpClient := &http.Client{
+		Transport: transport,
+		Timeout:   30 * time.Second,
```
sjmiller609: nit: very long timeout for local call maybe

lib/hypervisor/vz/client.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	return io.ReadAll(resp.Body)
+}
+
+func (c *Client) DeleteVM(ctx context.Context) error {
+	return c.doPut(ctx, "/api/v1/vm.shutdown", nil)
+}
```
sjmiller609, line 110: I'm not sure if this is exactly the same but in my PR we have new graceful shutdown handling it would signal guest to graceful shutdown user program then it exits cleanly

sjmiller609: we just might want a similar test for mac to check shutdown behaviors seem right or maybe OOM check and bad exit code check like we have on the CH test that does similar

sjmiller609: it might be calling at a different layer than this and that could would be reused I'm not sure

sjmiller609 (approved):
Just nits or possible additional color, looks good

scripts/install.sh
```diff
@@ -285,15 +336,45 @@ else
...
+	<key>com.apple.security.virtualization</key>
+	<true/>
+	<key>com.apple.security.network.server</key>
+	<true/>
+	<key>com.apple.security.network.client</key>
+	<true/>
+</dict>
+</plist>
+ENTITLEMENTS
+        codesign --force --sign - --entitlements "$ENTITLEMENTS_TMP" "${TMP_DIR}/${BINARY_NAME}" 2>/dev/null || true
```
hiroTamada: nit: this silently swallows codesign failures. if signing fails, the binary won't have the virtualization entitlement and will crash at runtime with a confusing error. consider at least warning on failure instead of `|| true`.

lib/hypervisor/README.md
```diff
@@ -4,20 +4,29 @@ Provides a common interface for VM management across different hypervisors.
...
 - **Feature parity where possible** - Common operations work the same way
 - **Graceful degradation** - Features unsupported by a hypervisor can be detected and handled
 
+## Implementations
+
+| Hypervisor | Platform | Process Model | Control Interface |
+|------------|----------|---------------|-------------------|
+| Cloud Hypervisor | Linux | External process | HTTP API over Unix socket |
+| QEMU | Linux | External process | QMP over Unix socket |
+| vz | macOS | Subprocess (vz-shim) | gRPC over Unix socket |
```
hiroTamada: nit: vz control interface is HTTP over Unix socket (matching the Cloud Hypervisor API pattern), not gRPC.

scripts/e2e-install-test.sh
```diff
@@ -0,0 +1,223 @@
...
+[ -n "$API_KEY" ] || fail "Failed to generate API token"
+pass "Generated API token"
+
+# Set CLI env
+export HYPEMAN_API_KEY="$API_KEY"
+export HYPEMAN_BASE_URL="http://localhost:${PORT:-8080}"
+
+# Determine CLI path
+if [ "$OS" = "darwin" ]; then
+    HYPEMAN_CMD="/usr/local/bin/hypeman"
```
hiroTamada: nit: both branches set the same path (`/usr/local/bin/hypeman`), so the if/else is a no-op.

hiroTamada (approved):
Solid PR. Clean hypervisor abstraction, good use of build tags, thorough docs and E2E coverage. The vz-shim subprocess design (surviving hypeman restarts, CH-compatible control API) is well thought out. The API handler refactor to use `GetVsockDialer` is a nice decoupling.

A few minor nits inline, nothing blocking.

cmd/vz-shim/main.go
rgarcia: Yes — vz-shim logs to the path set in `config.LogPath`, which starter.go sets to `{instanceDir}/logs/vz-shim.log`. This is the vmm.log equivalent for vz.

cmd/vz-shim/server.go
rgarcia: Added `TestVZBasicLifecycle` and `TestVZExecAndShutdown` in `manager_darwin_test.go` — covers exec, graceful shutdown, restart, and nonexistent command error handling.

cmd/vz-shim/vm.go
rgarcia: Added clarifying comment. The caller always provides MemoryBytes (defaults to 1GB in create.go), so 0 shouldn't happen in practice — this is a safety fallback.

lib/hypervisor/vz/client.go
rgarcia: Reduced to 10s.

lib/hypervisor/vz/client.go
rgarcia: Fixed `handleShutdown` to only send ACPI power button (graceful signal) without immediate force-kill fallback. This aligns with CH/QEMU semantics: `DeleteVM` = graceful signal, `Shutdown` = kill VMM process. The force-kill path remains in `handleVMMShutdown` / `killHypervisor()`. Added integration tests covering shutdown behavior.

scripts/install.sh
rgarcia: Good catch — changed to warn on failure instead of silently swallowing.

lib/hypervisor/README.md
rgarcia: Fixed — it's HTTP over Unix socket, matching the CH pattern.

scripts/e2e-install-test.sh
rgarcia: Removed the dead if/else — just sets the path directly now.

rgarcia:
@cursor push eaaacb0dd7

PR #92, opened 2026-02-11
perf(builds): native overlayfs snapshotter + zstd compression

hiroTamada:
## Summary

- Switch BuildKit from fuse-overlayfs to native kernel overlayfs snapshotter, running as root inside ephemeral microVMs (the VM is the security boundary)
- Mount tmpfs at `/var/lib/buildkit` to avoid nested overlayfs mknod conflict (VM rootfs is already overlayfs; char 0:0 whiteout markers can't be created on nested overlayfs)
- Use zstd compression for output layers (`compression=zstd,force-compression=true`) for faster decompression
- Update default kernel to v1.4 with `CONFIG_OVERLAY_FS_REDIRECT_DIR=y` and `CONFIG_OVERLAY_FS_INDEX=y` (required for native overlayfs)
- Increase builder VM defaults to 4 vCPUs / 4GB RAM to support tmpfs-backed builds
- Grant scoped pull access for internal registry base images by parsing Dockerfile FROM lines

### Performance (nodejs22-base, ~125MB compressed)

| Phase | Before (gzip + fuse-overlayfs) | After (zstd + native overlayfs + tmpfs) |
|---|---|---|
| Layer extraction | 12.4s | **2.1s** (6x faster) |
| Export/push | 9.2s | **~0s** (cached layers) |
| Total build (incl. post-build import) | ~31s | **~12s** |

### Prerequisites

- Base images need to be republished with zstd compression (e.g. `docker buildx build --output compression=zstd,force-compression=true,oci-mediatypes=true`)
- Kernel v1.4 (`ch-6.12.8-kernel-1.4-202602101`) with overlayfs config enabled

## Test plan

- [ ] Verify build succeeds with internal registry base image (zstd-compressed)
- [ ] Verify build succeeds with Docker Hub base image (gzip-compressed)
- [ ] Verify native overlayfs snapshotter is used (check `mount` output inside RUN step shows `runc-overlayfs`)
- [ ] Verify builder runs as root (`id` shows `uid=0`)
- [ ] Verify tmpfs is mounted at `/var/lib/buildkit`
- [ ] Verify kernel v1.4 is active (`zcat /proc/config.gz | grep OVERLAY`)
- [ ] Test builds with network egress (npm install, pip install)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
[...]

PR #75, opened 2026-01-28
fix: implement proper Docker registry token auth for BuildKit

hiroTamada:
## Summary

BuildKit was failing to authenticate with our registry due to mismatches in the auth flow. This PR implements proper Docker Registry Token Authentication.

### Root causes fixed:
1. **Bearer vs Basic challenge** - BuildKit expects `WWW-Authenticate: Bearer realm="...token"`, not Basic
2. **JWT location in Basic auth** - BuildKit sends `identitytoken` as password (`:jwt`), not username (`jwt:`)
3. **Claim format mismatch** - Middleware checked for legacy `repos` claim but tokens use `repo_access`
4. **CA cert not system-wide** - BuildKit's HTTP client for `/v2/token` uses OS trust store, not buildkitd.toml
5. **Hardcoded HTTPS scheme** - Token endpoint URL was always `https://` but production uses HTTP internally

### Changes:
- **lib/middleware/oapi_auth.go**: Bearer token flow, support both claim formats, remove IP fallback, detect scheme from request
- **lib/registry/token.go**: Accept JWT in password field for Basic auth
- **lib/builds/builder_agent/main.go**: Install CA cert system-wide, add identitytoken to docker config
- **lib/builds/images/generic/Dockerfile**: Use moby/buildkit:latest, add crane, fix stage ordering

## Test plan
- [x] Local build test passes with new auth flow
- [x] Local build test passes with HTTP registry (production-like config)
- [x] Middleware and registry unit tests pass
- [x] Deploy to staging and verify builds work

## Ops changes required
1. Deploy updated hypeman API server
2. Pull new builder image: `docker.io/onkernel/builder-generic:latest`

---

> [!NOTE]
> **Medium Risk**
> Touches registry authentication and TLS handling for build pushes/caches; misconfiguration or logic bugs could block builds or weaken registry access controls.
> 
> **Overview**
> Implements **Docker Registry Token Authentication** for the built-in OCI registry by adding a `/v2/token` endpoint and returning proper `WWW-Authenticate: Bearer realm=...` challenges for unauthenticated registry requests, removing the prior internal-IP fallback.
> 
> Updates build/registry plumbing to support BuildKit’s credential formats and per-repo permissions (new `repo_access` claims with legacy `repos` fallback), and extends build configuration to pass registry TLS options (insecure mode + optional CA cert file) through to builder VMs. The builder agent now generates `config.json`/`buildkitd.toml`, installs custom CA certs system-wide, and registry/image references now strip URL schemes for correctness.
[...]

PR #70, opened 2026-01-26
feat(builds): implement two-tier build cache with per-repo token scopes

hiroTamada:
## Summary

Implements a two-tier caching system for builds that separates global (shared) cache from tenant-specific cache, with fine-grained JWT token permissions for secure access control.

### Architecture

```
┌─────────────────────────────────────────────────────────────┐
│                     Global Cache                            │
│                cache/global/{key}                           │
│         (populated by admin builds only)                    │
└─────────────────────────────────────────────────────────────┘
                           │
              ┌────────────┼────────────┐
              │ pull       │ pull       │ pull
              ▼            ▼            ▼
       ┌──────────┐  ┌──────────┐  ┌──────────┐
       │ Tenant A │  │ Tenant B │  │ Tenant C │
       │  Cache   │  │  Cache   │  │  Cache   │
       └──────────┘  └──────────┘  └──────────┘
```

### Changes

- **Token Structure**: New `RepoPermission` struct with per-repository scope (`push`/`pull`)
- **Build Request**: New `is_admin_build` and `global_cache_key` fields
- **Manager**: Generates tokens with fine-grained repo access based on build type
- **Builder Agent**: Configures BuildKit with appropriate `--import-cache` and `--export-cache` flags
- **Registry Middleware**: Validates per-repo scopes from JWT tokens
- **API Layer**: Parses new multipart form fields

### Build Flow

| Build Type | Import From | Export To |
|------------|-------------|-----------|
| Admin (`is_admin_build=true`) | `cache/global/{key}` | `cache/global/{key}` |
| Regular tenant | `cache/global/{key}` + `cache/{tenant}` | `cache/{tenant}` |

### Usage Examples

```bash
# Regular tenant build - imports from global "node" cache, exports to tenant cache
curl -X POST -F "source=@source.tar.gz" \
  -F "global_cache_key=node" \
  -F "cache_scope=my-tenant" \
  http://localhost:8083/builds

# Admin build - populates global "ubuntu" cache
curl -X POST -F "source=@source.tar.gz" \
  -F "global_cache_key=ubuntu" \
  -F "is_admin_build=true" \
  http://localhost:8083/builds
```

## Test Plan

- [x] Unit tests for token generation with per-repo permissions
- [x] Unit tests for `IsRepositoryAllowed`, `GetRepoScope`, `IsPushAllowedForRepo`, `IsPullAllowedForRepo`
- [x] E2E test: Regular tenant build imports from global cache, exports to tenant cache
[...]

PR #96, opened 2026-02-13
Add image_name parameter to builds

rgarcia:
## Summary
- Adds optional `image_name` parameter to the build system, allowing callers to specify a custom image name for build output
- When `image_name` is set, images are pushed to `{registry}/{image_name}` instead of the default `{registry}/builds/{id}`
- Changes propagated through `CreateBuildRequest`, `BuildConfig`, builder agent output ref, API multipart form parsing, registry token scoping, and OpenAPI spec

## Test plan
- [ ] Create a build without `image_name` — verify existing `builds/{id}` behavior unchanged
- [ ] Create a build with `image_name=myapp` — verify image pushed to `{registry}/myapp`
- [ ] Verify registry token is scoped to the custom image name repo
- [ ] Verify recovered builds use the correct image name for token refresh

---

> [!NOTE]
> **Medium Risk**
> Touches the build/registry image naming and readiness flow; mistakes could cause images to be stored under the wrong ref or builds to report ready before an image is usable.
> 
> **Overview**
> Build creation now accepts an optional `image_name` parameter (API multipart + OpenAPI/OAPI types), validates it up front, and propagates it through `CreateBuildRequest`/`BuildConfig`.
> 
> Build completion flow is updated to **always wait for conversion of `builds/{id}`**, then optionally re-tag the resulting digest to the requested `image_name` via `ImportLocalImage` and (best-effort) wait for that tag to become ready; build metadata `ImageRef` reflects the custom name only if re-tag succeeds.
> 
> Registry/image handling is normalized to store and look up images by **short repo name (no host prefix)**, updating conversion triggering, build/image wait logic, and associated tests; `providers` also rewrites `RegistryURL` when configured as `localhost`/`127.0.0.1` so builder VMs can reach the host registry. Adds a `make run` target for non-hot-reload execution.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4f7d2bc2373d06d9f9a67c2640ab4407fc6da202. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


hiroTamada (approved):
Looks good.

I personally dont like how we need to specify registry url. Should this be optional/default?
For `hypeman run` we try internal registry first and `docker.io` if registry name is not specified?

@sjmiller609

sjmiller609 (approved):

sjmiller609:
> I personally dont like how we need to specify registry url. Should this be optional/default?
> For hypeman run we try internal registry first and docker.io if registry name is not specified?


I agree. also it would be best if "hypeman push .." "hypeman pull ..." "hypeman run ..." are always consistent in whatever the user puts for the image name there. e.g. if I push by some name and that works then I should be able to run by that same name and it's running the image I just pushed

rgarcia:
## DX Fixes Applied (ac17646)

Addressed the consistency concerns raised by @sjmiller609. Here's what was wrong and what's fixed:

### Problem: Push / Run naming mismatch

When you pushed an image via `hypeman push alpine:3.19 myapp`, the registry stored it as `localhost:8081/myapp:latest` (with the registry host prefix). But `hypeman run myapp` normalized the name to `docker.io/library/myapp:latest` — a completely different key. So **push and run couldn't round-trip**.

### Fixes

**1. Strip registry host prefix on push (registry.go)**

The registry's manifest PUT handler was prepending `req.Host` (e.g., `localhost:8081`) to the repo name before storing. Now it uses just the path component, so Docker's `ParseNormalizedNamed` handles normalization consistently:

- `push myapp` → stored as `docker.io/library/myapp:latest`
- `run myapp` → looks up `docker.io/library/myapp:latest` → ✅ found

**2. Backward-compat fallback in GetImage (images/manager.go)**

Added a fallback: if the normalized name isn't found, also try prepending the internal registry URL. This handles images that were stored with the old `localhost:PORT/name` format.

**3. Build `image_ref` uses short name (builds/manager.go)**

Build results now return `image_ref: "mybuiltapp"` instead of `"localhost:8081/mybuiltapp"`, so the user can directly `hypeman run mybuiltapp` with the value from the build output.

**4. Registry auth accepts user JWTs (middleware/oapi_auth.go)**

`hypeman push` from the CLI sends a regular user JWT token, but the `/v2/` registry endpoints only accepted special registry-scoped tokens. Added a fallback to validate regular user JWTs for registry paths.

**5. Regenerated oapi code + added `make run` target**

The `image_name` field from the OpenAPI spec wasn't in the generated Go code, causing 400 validation errors. Regenerated. Also added a `make run` target for running without live reload.

### Verified DX flow

```
$ hypeman push alpine:3.19 myapp
Pushed localhost:8081/myapp

$ curl http://localhost:8081/images/myapp
{"name": "docker.io/library/myapp:latest", "status": "ready", ...}

$ hypeman run myapp
Creating instance myapp-6dh6...
diitkynoorkq644o28x06cib   ← works!
```

### Still open

- **Build e2e test**: `hypeman build` with `image_name` accepted the API call correctly, but the builder VM can't pull base images from Docker Hub (network isolation issue unrelated to this PR). The `image_ref` would return `"mybuiltapp"` on success.
- **CLI `build` command**: Doesn't expose `--image-name` flag yet — would need a CLI-side change to pass through the new API field.

rgarcia:
## Build VM Networking Fixed (b2f5d1b)

The build system now works end-to-end. Here's what was broken and what's fixed:

### Root cause: builder VMs couldn't reach the host registry

The builder VM was configured with `registry_url: "localhost:8081"`, but inside the VM `localhost` refers to the VM itself — not the host machine. Three separate issues compounded:

1. **Registry URL not reachable** — `localhost:8081` inside VM = connection refused
2. **BuildKit HTTPS mismatch** — the baked builder image wrote a `buildkitd.toml` with `https=true`, but the registry is HTTP
3. **iptables NAT collision** — multiple hypeman instances on the same host shared the same `hypeman-nat` iptables comment, causing them to overwrite each other's MASQUERADE/FORWARD rules. Our instance's rules were getting deleted by other instances' startup.

### Fixes

**1. Registry URL rewriting for VMs (providers.go)**

`ProvideBuildManager` now rewrites `localhost:PORT` → `GATEWAY_IP:PORT` (e.g., `10.101.0.1:8081`) using the subnet config. Builder VMs reach the host via the bridge gateway IP.

**2. Proper BuildKit config in builder agent (builder_agent/main.go)**

Added `setupBuildKitConfig()` that writes a `buildkitd.toml` with `http=true` + `insecure=true` for the internal registry. Also sets `DOCKER_CONFIG=/home/builder/.docker` explicitly for rootlesskit user namespace compatibility.

**3. Multi-tenant iptables isolation (bridge.go)**

Changed iptables comments from generic `hypeman-nat` to bridge-scoped `hypeman-nat-{bridge_name}`. Each hypeman instance now manages its own rules without interfering with others.

**4. Registry auth path pattern fix (oapi_auth.go)**

Fixed `registryPathPattern` regex to support repository names with 3+ path segments (`a/b/c`). Addresses bugbot's "deep image paths break build pushes" concern.

**5. VM subnet auth fallback (oapi_auth.go)**

`isInternalVMRequest` was hardcoded to `10.102.x.x` — now accepts the subnet CIDR as a parameter so it matches the actual configured subnet.

### Verified e2e flow

```
$ curl -X POST localhost:8081/builds -F source=@test.tar.gz -F image_name=mybuiltapp
{"id": "xjop5eqomak8gkxco2odqnoo", "status": "queued"}

# Build completes in ~3.6s:
$ curl localhost:8081/builds/xjop5eqomak8gkxco2odqnoo
{"status": "ready", "image_ref": "mybuiltapp", "image_digest": "sha256:0f464e..."}

$ hypeman run mybuiltapp
Creating instance mybuiltapp-uzyx...
e6wo9ytjpl5gf3hngurx8w4k  ← works!
```

### Bugbot responses

- **"Qualified names no longer round-trip"** — False positive. `ParseNormalizedNamed` handles normalization consistently for both push and run. The backward-compat fallback in `GetImage` covers legacy images stored with the old prefix.
- **"Deep image paths break build pushes"** — Valid concern, fixed by updating `registryPathPattern` to use `(.+?)/(manifests|blobs|tags|referrers)/` instead of the old 2-segment limit.

rgarcia:
**Rebased onto main + addressed bugbot comment**

### Bugbot: "Repo parsing breaks reserved path names"
This is now a non-issue after rebasing. The base branch already switched from a regex to `v2.Router()` from docker/distribution (introduced in the two-tier cache PR #70), which properly parses all OCI repository names including those with reserved segment names like `blobs` or `manifests`. The regex we had (`registryPathPattern`) no longer exists in the codebase.

### Rebase summary
Rebased onto `main` (which included ~30 commits since our branch point). Key changes integrated:
- **Two-tier build cache** (#70) — merged `ImageName` into `RepoPermission`-based token scoping
- **Docker registry token auth** (#75) — kept their `v2.Router()` repo parsing, added our user JWT fallback and subnet CIDR checks on top
- **Pre-cache base images via BuildKit mirror** (#91) — kept their `setupBuildkitdConfig` with TLS/CA cert support, added our multi-dir config writes
- **Native overlayfs snapshotter** (#92) — kept their `BUILDKITD_FLAGS` env var approach (replaces our simpler `DOCKER_CONFIG` approach)
- **erofs optimization + revert** (#94, #98) — handled `drivers.go` deletion
- **Metadata field** (#97) — no conflicts

All conflicts resolved carefully to preserve both sides' functionality. Builds and vets clean.

cmd/api/api/builds.go
```diff
@@ -137,6 +137,15 @@ func (s *ApiService) CreateBuild(ctx context.Context, request oapi.CreateBuildRe
...
 			globalCacheKey = string(data)
+		case "image_name":
+			data, err := io.ReadAll(part)
+			if err != nil {
+				return oapi.CreateBuild400JSONResponse{
+					Code:    "invalid_request",
+					Message: "failed to read image_name field",
+				}, nil
+			}
+			imageName = string(data)
```
hiroTamada, line 149: nit: might be worth validating image_name here (e.g. via `images.ParseNormalizedRef`) to give a fast 400 instead of silently falling back to builds/{id} after a successful build.

hiroTamada (approved):
clean design — builder stays simple (pushes to builds/{id}), server handles the re-tag. one minor nit on upfront validation.

PR #116, opened 2026-02-28
Hypothetical monorepo setup

sjmiller609:
this would be merged via merge request


sjmiller609:
seems too complicated with the stainless syncing, and the updating of sdks takes too long

PR #125, opened 2026-03-07
Snapshot

sjmiller609:
> [!NOTE]
> **High Risk**
> Adds new snapshot lifecycle operations that copy/replace VM guest directories and interact with hypervisor standby/restore flows, so bugs could lead to data loss or broken instance state transitions. Also expands the public API surface and manager interface, requiring careful validation across hypervisors.
> 
> **Overview**
> Adds a **centrally managed snapshot** feature for instances, including storage under `data/snapshots/` with metadata + guest payload, and new manager methods (`CreateSnapshot`, `RestoreSnapshot`, `ForkSnapshot`, `List/Get/DeleteSnapshot`) plus `ErrSnapshotNotFound`.
> 
> Exposes snapshots via new HTTP endpoints (`POST /instances/{id}/snapshots`, `POST /instances/{id}/snapshots/{snapshotId}/restore`, `GET/DELETE /snapshots`, `GET /snapshots/{snapshotId}`, `POST /snapshots/{snapshotId}/fork`) with updated generated `oapi` client/server types and response parsing.
> 
> Implements snapshot semantics for `Standby` vs `Stopped` kinds (including target-state defaults, hypervisor override only for stopped snapshots, sparse-copy enforcement, name uniqueness, and read-only volume safety), updates fork target-state transitions to re-check guest-agent readiness for running returns, and adds integration/unit coverage across Firecracker/QEMU/Cloud Hypervisor/VZ.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34aa76499d9626b1ea298253e23790af7ddac343. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sjmiller609:
depends on https://github.com/kernel/hypeman/pull/123

lib/instances/types.go
```diff
@@ -182,6 +183,41 @@ type ForkInstanceRequest struct {
...
+// Snapshot is a centrally stored immutable snapshot resource.
+type Snapshot = snapshot.Snapshot
+
+// ListSnapshotsFilter contains optional filters for listing snapshots.
+type ListSnapshotsFilter = snapshot.ListSnapshotsFilter
+
+// CreateSnapshotRequest is the domain request for creating a snapshot.
+type CreateSnapshotRequest struct {
+	Kind SnapshotKind // Required: Standby or Stopped
+	Name string       // Optional: unique per source instance
```
rgarcia, line 205: I think at some point we should support arbitrary key-val (string-string) metadata along with name on resources. Kinda like how tags work in AWS

lib/snapshot/README.md
```diff
@@ -0,0 +1,71 @@
...
+- Captures standby-style state, including memory/device snapshot state plus disks.
+- Intended for fast resume-style recovery.
+- Can be created from `Running` or `Standby`.
+- Does **not** allow hypervisor switching on restore/fork.
+
+### `Stopped`
+- Captures disk-focused state from a stopped VM.
+- Intended for cold-start style restore/fork.
+- Can be created only from `Stopped`.
+- Allows optional hypervisor switching on restore/fork.
```
rgarcia: This is wild

rgarcia (approved):

lib/instances/types.go
sjmiller609: Good call. I agree this should exist, but I kept it out of this PR to avoid expanding the snapshot API surface further while we stabilize lifecycle behavior. I opened follow-up issue #126 to add first-class string metadata/tags support across resources, including snapshots.

lib/snapshot/README.md
sjmiller609: Updated in 34aa764: I clarified that hypervisor switching is only allowed for `Stopped` snapshots because memory state is not carried across. That reads less surprisingly now.

lib/instances/snapshot.go
```diff
@@ -0,0 +1,550 @@
...
+		return nil, fmt.Errorf("%w: snapshot %s belongs to instance %s", ErrInvalidRequest, snapshotID, rec.Snapshot.SourceInstanceID)
+	}
+
+	sourceMeta, err := m.loadMetadata(id)
+	if err != nil {
+		return nil, err
+	}
+	sourceInst := m.toInstance(ctx, sourceMeta)
+	if sourceInst.State == StateRunning {
+		return nil, fmt.Errorf("%w: cannot restore snapshot while source is %s", ErrInvalidState, sourceInst.State)
```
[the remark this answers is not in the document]
sjmiller609: seems ok to me

PR #133, opened 2026-03-08
chore: test reliability and speed skill

sjmiller609:
> [!NOTE]
> **Medium Risk**
> Touches runtime paths for ingress binary extraction and default-network discovery, so any bugs could affect multi-process startups or instance networking on Linux. Changes are bounded (file-lock + atomic rename; short self-heal polling) and mainly aimed at reducing CI flakiness.
> 
> **Overview**
> Improves Linux integration test reliability by hardening embedded Caddy extraction: `ExtractCaddyBinary` now uses a per-binary `Flock` and temp-file + atomic rename writes (including the `.sha256`) to avoid concurrent `ETXTBUSY`/partial-write races.
> 
> Makes default-network allocation more robust under concurrent tests by adding `getDefaultNetworkWithSelfHeal`, which re-runs initialization and briefly polls for bridge readiness before failing.
> 
> Reduces test flakiness/latency by removing an external S3 curl dependency in `TestFirecrackerNetworkLifecycle` (replaced with a local probe server), trimming redundant reachability assertions in running-fork tests, and lowering guest exec `WaitForAgent` timeout in `execInInstance` to speed up polling-based assertions.
> 
> Adds a new `skills/test-agent` skill with prompt and investigation notes documenting the flake root causes and validation runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c94b92d8f2843659110c16dd385491bd14f7ef45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #134, opened 2026-03-08
Optional VM egress MITM proxy with mock-secret header rewriting

sjmiller609:
## Summary
This PR adds an optional, default-off egress MITM proxy mode for Hypeman VMs so workloads can run with mock secrets in-VM while real secrets stay on the host.

When enabled per instance, Hypeman now:
- Starts/uses a host-side HTTP/HTTPS MITM proxy on the VM bridge gateway.
- Injects proxy env vars into the guest (`HTTP_PROXY` / `HTTPS_PROXY`, lower-case variants).
- Installs proxy CA material in guest init so TLS MITM can be trusted by system components.
- Rewrites outbound HTTP header values by replacing configured mock literals with real values from host env vars.
- Enforces egress path on Linux using per-instance iptables FORWARD rules for TCP ports 80/443 so direct internet egress bypass is rejected.

## Why
This enables safer execution of untrusted or lower-trust workloads in VMs without placing real secrets inside guest env/config, while still allowing authenticated outbound API traffic.

## API / Config changes
- `CreateInstanceRequest` now supports top-level `egress_proxy`:
  - `enabled: bool`
  - `mock_to_real_env_var: map[string]string` (mock literal -> host env var name)
- OpenAPI schema updated accordingly.
- Instance domain model now persists optional `EgressProxyConfig`.

## Implementation details
- New module: `lib/egressproxy/`
  - Host CA generation/load and on-the-fly cert signing.
  - HTTP proxying and HTTPS CONNECT MITM handling.
  - Header replacement policy resolved per source VM IP.
  - Linux enforcement helpers for egress gating.
  - Behavior-focused README.
- Lifecycle wiring in instances manager:
  - Register proxy policy/enforcement on create/start/restore when enabled.
  - Remove policy/enforcement on stop/standby/delete and rollback paths.
- Guest init support:
  - New guest config section for egress proxy settings and CA PEM.
  - CA installation call added in both exec and systemd modes.

## Tests
Added integration test:
- `TestEgressProxyRewritesHTTPSHeaders`
  - Boots a VM with egress proxy mode enabled.
  - Sends HTTPS request with mock secret in `Authorization` header.
  - Verifies upstream receives rewritten real secret value.

## Validation run
Executed on `deft-kernel-dev` as root:
- `sudo -n /usr/local/go/bin/go test ./cmd/api/api -run TestDoesNotExist -count=1`
- `sudo -n /usr/local/go/bin/go test ./lib/instances -run TestEgressProxyRewritesHTTPSHeaders -count=1 -v`
- `sudo -n /usr/local/go/bin/go test ./... -run TestDoesNotExist -count=1`

All passed for this change set.

## Notes
- Enforcement currently targets default HTTP/HTTPS ports (80/443) by design.
- Header replacement applies to HTTP headers only (not bodies).

---

> [!NOTE]
> **High Risk**
> Introduces a host-side MITM proxy and Linux `iptables` enforcement that alters VM outbound networking and handles real secret material, so misconfiguration or bugs could impact security and connectivity.
> 
> **Overview**
> Adds an **optional, default-off egress MITM proxy mode** for instances, including new API fields `network.egress` (with enforcement mode) and `credentials` to keep real secrets host-side while the guest receives `mock-*` placeholders.
> 
> Wires the feature through instance lifecycle (`create`/`start`/`restore` and cleanup on `stop`/`standby`/`delete`): registers per-instance proxy + enforcement, regenerates config disks to inject proxy settings and mock env vars, and persists new metadata (`NetworkEgress`, `Credentials`).
> 
> Introduces `lib/egressproxy` (CA generation + per-host cert signing, CONNECT MITM, HTTPS-only header rewriting with domain allowlists, and Linux `iptables` egress blocking) plus guest-init support to install the proxy CA, expands test coverage (unit + VM integration), and updates prewarm/e2e scripts and tests to support registry mirroring and new OpenAPI fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a25a59b12b7f72374303be15b626a05f9da8e3ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/egressproxy/service.go
```diff
@@ -0,0 +1,399 @@
...
+	for k, vv := range resp.Header {
+		for _, v := range vv {
+			w.Header().Add(k, v)
+		}
+	}
+	w.WriteHeader(resp.StatusCode)
+	_, _ = io.Copy(w, resp.Body)
+}
+
+func (s *Service) handleConnect(w http.ResponseWriter, r *http.Request, sourceIP string) {
```
sjmiller609, line 296: This is a custom SSL bumping proxy implementation, but I'm wondering if a good dependency might be more reliable. It might be worth starting out with this custom implementation until there's some reason why it doesn't work, then consider dependencies.

lib/ingress/binaries_linux.go
```diff
@@ -57,13 +58,29 @@ func ExtractCaddyBinary(p *paths.Paths) (string, error) {
 		return "", fmt.Errorf("create caddy binary dir: %w", err)
 	}
 
-	// Write binary to filesystem
-	if err := os.WriteFile(extractPath, data, 0755); err != nil {
+	lockFile, err := os.OpenFile(extractPath+".lock", os.O_CREATE|os.O_RDWR, 0644)
```
sjmiller609, line 61: there are a few changes in this PR from the other PR test flakes addressed: https://github.com/kernel/hypeman/pull/133

but I didn't want to remove them because I wanted this CI runs to benefit from the de-flaking

lib/egressproxy/service.go
sjmiller609: I agree with the tradeoff. We kept this in-tree for now to keep policy coupling explicit (per-VM rule model, enforcement integration, and testability hooks) and to avoid introducing a new proxy dependency until we have concrete operational gaps. If we hit reliability/maintenance pressure, we can revisit dependency-backed alternatives with this behavior as the contract baseline.

lib/ingress/binaries_linux.go
sjmiller609: Kept. I intentionally retained the de-flaking adjustments from #133 in this branch so CI for this PR stays representative and stable while this feature is under review.

lib/egressproxy/service.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if existing := s.certCache[host]; existing != nil {
+		return existing, nil
+	}
+	s.certCache[host] = cert
+	return cert, nil
+}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in a9231cd. Added a bounded leaf-cert cache with FIFO eviction (`defaultLeafCertCacheLimit=512`) so cache growth is capped and old host entries are removed when capacity is reached.

cmd/api/api/test_prewarm_test.go
```diff
@@ -0,0 +1,65 @@
...
+		mapped = registry + "/" + repo + ":" + ref.Tag()
+	case ref.Digest() != "":
+		mapped = registry + "/" + repo + "@" + ref.Digest()
+	default:
+		mapped = registry + "/" + repo + ":latest"
+	}
+
+	apiRegistryLogOnce.Do(func() {
+		t.Logf("using test registry mirror source=%s mapped=%s", source, mapped)
+	})
```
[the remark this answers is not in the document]
sjmiller609: Fixed in a9231cd. Removed `t.Logf` from inside the package-level `sync.Once` closures; `sync.Once` now only computes the message and each test invocation logs through its own current `*testing.T`.

lib/instances/fork.go
```diff
@@ -467,6 +467,19 @@ func cloneStoredMetadataForFork(src StoredMetadata) StoredMetadata {
...
+			EnforcementMode: src.EgressProxy.EnforcementMode,
+		}
+		if src.EgressProxy.MockEnvVars != nil {
+			cfg.MockEnvVars = append([]string(nil), src.EgressProxy.MockEnvVars...)
+		}
+		if src.EgressProxy.MockEnvVarDomains != nil {
+			cfg.MockEnvVarDomains = cloneMockEnvVarDomains(src.EgressProxy.MockEnvVarDomains)
+		}
+		dst.EgressProxy = cfg
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 26c4a29. `cloneStoredMetadataForFork` now delegates to `cloneEgressProxyConfig(...)` so there is a single clone path for `EgressProxyConfig`.

lib/egressproxy/service.go
```diff
@@ -0,0 +1,506 @@
...
+		"tcp",
+		addr,
+		tlsCfg,
+	)
+	if err != nil {
+		return err
+	}
+	_ = conn.Close()
+	return nil
+}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 26c4a29 by removing the CONNECT preflight TLS probe. Upstream verification now happens on the single `RoundTrip` path, eliminating the extra handshake while still using standard TLS verification roots.

lib/egressproxy/enforce_linux.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+func enforcementComment(instanceID, suffix string) string {
+	short := instanceID
+	if len(short) > 8 {
+		short = short[:8]
+	}
+	return fmt.Sprintf("hypeman-egress-%s-%s", short, suffix)
+}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 26c4a29. `enforcementComment` now includes the full instance ID instead of an 8-char prefix, removing cross-instance collision risk when removing iptables rules by comment.

cmd/api/api/api_test.go
```diff
@@ -146,6 +146,7 @@ func ctxWithImage(svc *ApiService, name string) context.Context {
 // Returns the image name on success, or fails the test on error/timeout.
 func createAndWaitForImage(t *testing.T, svc *ApiService, imageName string, timeout time.Duration) string {
 	t.Helper()
+	imageName = apiTestImageRef(t, imageName)
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 3266c10. The API tests now pass raw image refs into `createAndWaitForImage` and assign the helper return value back to `imageName`, so the mapping is applied exactly once.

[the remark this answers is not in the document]
sjmiller609: Addressed in 3266c10 with the same callsite cleanup: `createAndWaitForImage` is now the single place that maps mirrored test image refs for these API tests.

lib/egressproxy/service.go
```diff
@@ -0,0 +1,477 @@
...
+				Host:     r.Host,
+				Path:     r.URL.Path,
+				RawPath:  r.URL.RawPath,
+				RawQuery: r.URL.RawQuery,
+			}
+		}
+	}
+
+	outReq.RequestURI = ""
+	s.applyHeaderReplacements(sourceIP, "", outReq.Header, false)
```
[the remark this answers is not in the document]
sjmiller609: Hardened in 060e1d1. The HTTP path now propagates a normalized destination host into `applyHeaderReplacements` (still `isHTTPS=false`, so no substitutions occur), which removes the empty-host fragility noted here.

lib/egressproxy/service.go
```diff
@@ -0,0 +1,477 @@
...
+		return existing, nil
+	}
+	if s.certCacheLimit > 0 && len(s.certCache) >= s.certCacheLimit && len(s.certCacheOrder) > 0 {
+		evictHost := s.certCacheOrder[0]
+		s.certCacheOrder = s.certCacheOrder[1:]
+		delete(s.certCache, evictHost)
+	}
+	s.certCache[host] = cert
+	s.certCacheOrder = append(s.certCacheOrder, host)
+	return cert, nil
```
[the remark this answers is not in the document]
sjmiller609: Agreed this is a benign efficiency race, not a correctness issue. Keeping current behavior for now since it is functionally safe and bounded by the cache limit.

cmd/api/api/test_prewarm_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if apiRegistryLogMessage != "" {
+		t.Log(apiRegistryLogMessage)
+	}
+	return mapped
+}
+
+func isTestPrewarmStrict() bool {
+	v := strings.TrimSpace(os.Getenv(testPrewarmStrictEnv))
+	return v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes")
+}
```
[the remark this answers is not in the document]
sjmiller609: Acknowledged. I’m leaving the duplication as-is in this PR because these helpers live in separate test packages and this change set is already large; we can extract shared test-only utility code in a dedicated follow-up.

lib/egressproxy/enforce_linux.go
```diff
@@ -0,0 +1,121 @@
...
+	for i := len(ruleNums) - 1; i >= 0; i-- {
+		delCmd := exec.Command("iptables", "-D", "FORWARD", ruleNums[i])
+		_ = delCmd.Run()
+	}
+	return nil
+}
+
+func enforcementComment(instanceID, suffix string) string {
+	return fmt.Sprintf("hypeman-egress-%s-%s", instanceID, suffix)
+}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 3266c10. `enforcementComment` now sanitizes the instance ID, and rule removal matches the exact rendered comment marker (`/* ... */`) rather than substring-only matching.

lib/egressproxy/cert.go
```diff
@@ -0,0 +1,162 @@
...
+	return &tlsCert, nil
+}
+
+func normalizeHost(rawHost string) string {
+	host := strings.TrimSpace(rawHost)
+	if h, _, err := net.SplitHostPort(host); err == nil {
+		return h
+	}
+	return host
+}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 3ef9157. Removed the unused `normalizeHost` helper from `lib/egressproxy/cert.go` so `normalizeDestinationHost` remains the single normalization path.

openapi.yaml
```diff
@@ -162,6 +162,45 @@ components:
           example:
             PORT: "3000"
             NODE_ENV: production
+        egress_proxy:
```
rgarcia: i like the direction here, but i’m a little worried the public api is getting shaped around the current implementation (`egress_proxy` + mocked env var rewriting) rather than the longer-term policy model we may want.

after looking at a few similar products, the ones that seem to age best expose this as a network/credential policy problem, not a proxy feature:
- [vercel sandbox firewall](https://vercel.com/docs/vercel-sandbox/concepts/firewall) has first-class network policy (`allow-all` / `deny-all` / custom allow+deny rules) and then layers [credential brokering / header transforms](https://vercel.com/changelog/safely-inject-credentials-in-http-headers-with-vercel-sandbox) on top
- [deno sandbox](https://docs.deno.com/sandbox/security) separates `allowNet` from `secrets`, where secrets are host-scoped and substituted on outbound requests without ever entering the vm
- [e2b](https://e2b.dev/docs/sandbox/internet-access) already has `allowOut` / `denyOut`, and even their open [credential brokering proposal](https://github.com/e2b-dev/E2B/issues/1160) models secret injection as an egress transform layered onto network policy
- [daytona](https://www.daytona.io/docs/en/network-limits) is simpler, but even there the api is framed as network restriction (`networkAllowList` / `networkBlockAll`), not proxy config

the reason i’m flagging it is that this shape feels pretty mechanism-specific:
- `egress_proxy` bakes in “host-side mitm proxy” as the conceptual model
- `mock_env_vars` assumes env vars are the long-term secret source abstraction
- `mock_env_var_domains` couples destination policy to that env-var abstraction
- `enforcement_mode` is really egress/firewall policy, but it lives inside the same object as secret substitution

if we ever want to grow this into broader outbound controls (allow/deny rules, port restrictions, runtime policy updates, non-env-backed credentials, different injection strategies, explicit private-network blocking, etc), i think this api may box us in a bit.

maybe the public shape wants to be closer to something like this, even if the backend implementation for v1 is still proxy-based:

```yaml
network:
  egress:
    default: allow | deny
    allow:
      - hosts: ["api.openai.com", "*.anthropic.com"]
      - cidrs: ["10.0.0.0/8"]
    deny:
      - cidrs: ["169.254.169.254/32"]
    enforcement:
      mode: all | http_https_only

credentials:
  OPENAI_API_KEY:
    source:
      env: OPENAI_API_KEY
    inject:
      - hosts: ["api.openai.com"]
        as:
          header: Authorization
          format: "Bearer ${value}"
```

or some variation of that. the main thing i’m after is separating:
1. what outbound destinations are reachable
2. what credentials can be brokered where
3. how we currently implement that under the hood

that feels a little more future-proof to me than making `egress_proxy` itself the core api concept.

rgarcia: one adjacent use case that makes me want a slightly more general abstraction here is HTTP request signing for web bot auth / RFC 9421.

we already support the Cloudflare MV3 extension approach today, but long term i’d much rather keep the signing key entirely outside the vm and have Kernel sign outbound requests on the host. conceptually that seems very close to what this PR is already doing: terminate HTTPS on the host, inspect the final outbound request, and mutate/add headers before forwarding.

that makes me think the underlying primitive here may be broader than “mock env var rewriting”. it feels closer to host-managed egress transforms / credential brokering, where:
- mock secret substitution is one transform
- request signing (for web bot auth or similar) is another transform
- destination allow/deny policy is a separate concern again

for a web-bot-auth style profile, the policy might look more like “for these hosts, attach `Signature`, `Signature-Input`, and `Signature-Agent` using key X / algorithm Y”, with the private key never entering the guest.

i’m not necessarily suggesting we fully build that API in this PR, but i do think it’s one more reason to be careful about baking the public surface too tightly around `egress_proxy` + `mock_env_vars`. if request signing is a plausible next step, it’d be nice for the API and/or internal model to leave room for generic host-side egress request transforms on top of the same mediation path.

lib/instances/restore.go
```diff
@@ -169,6 +174,46 @@ func (m *manager) restoreInstance(
...
+			instForConfig := &Instance{StoredMetadata: *stored}
+			if err := m.createConfigDisk(ctx, instForConfig, imageInfo, proxyCfg, proxyGuestConfig); err != nil {
+				log.ErrorContext(ctx, "failed to refresh config disk for restore", "instance_id", id, "error", err)
+				releaseNetwork()
+				return nil, fmt.Errorf("refresh restore config disk: %w", err)
+			}
+			if stored.EgressProxy != nil && stored.EgressProxy.Enabled {
+				proxyRegistered = true
+			}
+		}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in a25a59b. Restore config-disk refresh is now gated by requiresRestoreConfigDiskRefresh(stored) (network enabled plus network egress enabled). Non-mediated restores no longer call GetImage or createConfigDisk, removing that failure path. Added regression coverage in TestRequiresRestoreConfigDiskRefresh.

openapi.yaml
sjmiller609: Implemented in a25a59b. The public create API now pivots from egress_proxy to a policy model: network.egress plus credentials (source + inject rules), and the internal model is split the same way (NetworkEgressPolicy + CredentialPolicy). The host-side MITM remains an implementation adapter only. The credential inject schema is intentionally extensible for future host-managed transforms (including signing), but this PR only implements the existing header-template substitution runtime.

lib/instances/restore.go
```diff
@@ -169,6 +174,46 @@ func (m *manager) restoreInstance(
...
+			return nil, fmt.Errorf("get network allocation for proxy setup: %w", allocErr)
+		}
+		if allocErr == nil && alloc != nil {
+			proxyCfg := &network.NetworkConfig{
+				IP:        alloc.IP,
+				MAC:       alloc.MAC,
+				Gateway:   alloc.Gateway,
+				Netmask:   alloc.Netmask,
+				TAPDevice: alloc.TAPDevice,
+			}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in a25a59b. Restore now builds proxy config from allocation via networkConfigFromAllocation, and allocation derivation now includes DNS so createConfigDisk receives GuestDNS correctly. Added regression coverage in TestNetworkConfigFromAllocation_PreservesDNS.

lib/instances/create.go
```diff
@@ -474,6 +490,29 @@ func validateCreateRequest(req CreateInstanceRequest) error {
...
+		if err != nil {
+			return err
+		}
+		req.NetworkEgress.EnforcementMode = mode
+	}
+	normalizedCredentials, err := normalizeCredentialPolicies(req.Credentials)
+	if err != nil {
+		return err
+	}
+	req.Credentials = normalizedCredentials
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 865b50c. `validateCreateRequest` now takes `*CreateInstanceRequest` and is called with `&req` from `createInstance`, so normalized credentials are persisted on the request used for metadata/runtime setup. Added regression coverage in `TestValidateCreateRequest_NormalizesCredentialsInPlace`.

lib/egressproxy/service.go
```diff
@@ -0,0 +1,481 @@
...
+	destinationHost := normalizeDestinationHost(outReq.URL.Host)
+	if destinationHost == "" {
+		destinationHost = normalizeDestinationHost(outReq.Host)
+	}
+	s.applyHeaderReplacements(sourceIP, destinationHost, outReq.Header, false)
+
+	resp, err := s.transport.RoundTrip(outReq)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("proxy upstream error: %v", err), http.StatusBadGateway)
+		return
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 865b50c. HTTP proxy failures now return a generic `proxy upstream error` body while logging the detailed upstream failure on the host (`slog.Warn`) so guest VMs do not receive internal DNS/TLS/transport details. Added regression coverage in `TestHandleHTTPProxyRequest_DoesNotLeakUpstreamErrorDetails`.

openapi.yaml
```diff
@@ -181,6 +274,8 @@ components:
               type: string
               description: Upload bandwidth limit (VM→external, e.g., "1Gbps", "125MB/s"). Defaults to proportional share based on CPU allocation.
               example: "1Gbps"
+            egress:
```
rgarcia, line 288: might be worth adding a field-level description here clarifying that `network.egress` enables the mediated egress path, and omitting it preserves normal direct outbound networking without credential rewriting. I think a description here would mitigate the risk that a user thinks they need to fill out the `egress` object if they want egress networking at all.

openapi.yaml
```diff
@@ -117,6 +117,83 @@ components:
...
+          default: false
+          example: true
+        enforcement:
+          $ref: "#/components/schemas/CreateInstanceRequestNetworkEgressEnforcement"
+
+    CreateInstanceRequestCredentialSource:
+      type: object
+      required: [env]
+      properties:
+        env:
```
rgarcia, line 155: might be worth adding a description here explaining the mechanics of the `env` source, e.g. that this reads the real value from the request env map, while mediated egress replaces the guest-visible env var with a placeholder and only the proxy path resolves it back to the real value

openapi.yaml
```diff
@@ -162,6 +239,22 @@ components:
           example:
             PORT: "3000"
             NODE_ENV: production
+        credentials:
```
rgarcia, line 251: might be worth clarifying in this description that keys in `credentials` are the guest-visible env var names. those guest env vars get placeholder values, and the proxy resolves them back to the real secret from `source.env` according to the `inject` rules

lib/instances/create.go
```diff
@@ -403,8 +406,21 @@ func (m *manager) createInstance(
 
 	// 16. Create config disk (needs Instance for buildVMConfig)
 	inst := &Instance{StoredMetadata: *stored}
+	var proxyGuestConfig *egressproxy.GuestConfig
+	if networkName != "" {
```
rgarcia: consider tightening this contract a bit: `networkName != ""` is an indirect proxy for networking state, while the real feature gate lives inside `maybeRegisterEgressProxy` on `stored.NetworkEgress.Enabled`. using a more direct predicate here (or calling the helper unconditionally and letting it no-op) would make the create/start/restore flows easier to reason about

lib/instances/egress_proxy.go
```diff
@@ -0,0 +1,272 @@
...
+
+	out := make([]egressproxy.SecretRewriteRuleConfig, 0, len(names))
+	for _, credentialName := range names {
+		policy := credentials[credentialName]
+		real := strings.TrimSpace(env[policy.Source.Env])
+		if real == "" {
+			continue
+		}
+
+		rule := egressproxy.SecretRewriteRuleConfig{
```
rgarcia: the API shape exposes `inject.as.header` and `inject.as.format`, but by the time we build the proxy rule here we only retain mock/real values plus allowed domains. as written, the low-level behavior looks like generic header value substitution rather than "inject this credential as this specific header using this format" - is that mismatch intentional?

lib/system/init/egress_proxy.go
```diff
@@ -0,0 +1,32 @@
...
+	if err := os.MkdirAll(filepath.Dir(caPath), 0755); err != nil {
+		log.Error("hypeman-init:egress-proxy", "failed to create CA directory", err)
+		return
+	}
+	if err := os.WriteFile(caPath, []byte(cfg.EgressProxy.CACertPEM), 0644); err != nil {
+		log.Error("hypeman-init:egress-proxy", "failed to write proxy CA", err)
+		return
+	}
+
+	cmd := exec.Command("/bin/sh", "-c", "if command -v update-ca-certificates >/dev/null 2>&1; then update-ca-certificates >/dev/null 2>&1; fi")
```
rgarcia: consider making the "no update-ca-certificates available" case explicit instead of silently succeeding here. as written, an image without that tool will still log CA installation as successful, but proxied HTTPS from inside the guest may later fail with certificate trust errors and the user won't get a clear setup-time signal

openapi.yaml
sjmiller609: Addressed in 2dd5cfc. I expanded the `network.egress` schema description to make the default path explicit: omitting `network.egress` preserves normal direct outbound networking when `network.enabled=true`, and enabling it opts into the mediated egress path plus host-managed credential rewriting.

openapi.yaml
sjmiller609: Addressed in 2dd5cfc. I clarified the `credentials` object description so the map keys are explicitly documented as the guest-visible env var names, those guest vars receive placeholder values, and the real value comes from `source.env` and is only materialized on the mediated egress path according to the `inject` rules.

lib/instances/egress_proxy.go
sjmiller609: Addressed in 2dd5cfc. I changed the runtime to carry `inject.as.header` and `inject.as.format` all the way into the host proxy rules instead of collapsing to generic mock/real substitution. Matching HTTPS requests now get the configured header materialized from the configured format on the host, and I updated unit/integration coverage to assert that injection behavior directly.

lib/system/init/egress_proxy.go
sjmiller609: Addressed in 2dd5cfc. Guest init no longer silently treats the missing-tool case as success: it now explicitly logs when `update-ca-certificates` is unavailable and says trust-store refresh was skipped, while still logging the refreshed-trust-store path separately when the tool is present. I also added coverage for the PATH lookup helper in `lib/system/init/egress_proxy_test.go`.

lib/instances/create.go
sjmiller609: Addressed in 2dd5cfc. The create flow now calls `maybeRegisterEgressProxy(...)` unconditionally and lets that helper own the real feature gate / no-op behavior, so create/start/restore are all using the same direct predicate instead of `networkName != ""` as an indirect proxy.

openapi.yaml
sjmiller609: Addressed in 2dd5cfc. The `credentials[*].source.env` field description now explicitly says this reads the real value from the request `env` map, while the guest-visible env var can be replaced with a mock placeholder and only the mediated egress path resolves it back to the real value on the host.

rgarcia (approved):
still seems like there's no hard api/control-plane error when mitm ca cert setup fails in the guest... seems not ideal to have to inspect guest logs to find that this failure. fine merging as is to keep things moving though

sjmiller609:
> still seems like there's no hard api/control-plane error when mitm ca cert setup fails in the guest... seems not ideal to have to inspect guest logs to find that this failure. fine merging as is to keep things moving though

I'll take a look. I still need to review the changes it decided to do since the previous review, and I'll see if I can handle that too. thanks for the careful reviews.

PR #146, opened 2026-03-13
perf: speed up firecracker standby with diff snapshot reuse

rgarcia:
## Summary
- cut Firecracker standby latency by reusing a hidden retained snapshot base across restore/standby cycles so later standby snapshots can use fast diff snapshots instead of rewriting a full memory image
- keep the user-facing standby semantics the same: `snapshot-latest` is only visible while an instance is actually in `Standby`, while the retained `snapshot-base` stays internal and gets cleaned up on stop
- model snapshot-base reuse and graceful VMM shutdown as hypervisor capabilities instead of encoding those decisions in Firecracker-specific paths or `ErrNotSupported` control flow
- move static capability lookup to a type-level registry so orchestration code can make lifecycle decisions from hypervisor type without constructing starter/client values just to ask about static behavior
- harden snapshot failure rollback by discarding a promoted snapshot target after snapshot errors so we do not reuse a partially written retained base on the next standby cycle
- local benchmark harness and workflow live in companion PR #145

## Why These Refactors Are In This PR
- The original performance win came from Firecracker-specific behavior, but the lifecycle code was cleaner once the reusable parts were expressed as capabilities. `SupportsSnapshotBaseReuse` lets the manager reason about retained bases generically instead of branching on `hypervisor.Type`.
- The graceful shutdown optimization started as "skip the wait when `Shutdown()` returns `ErrNotSupported`", but that is really a static property of the hypervisor, not a runtime error case. `SupportsGracefulVMMShutdown` makes that intent explicit and keeps error handling separate from control flow.
- Registering capabilities by hypervisor type avoids the earlier smell of needing a starter or zero-value client instance just to answer static questions about what a hypervisor supports.

## Firecracker Snapshot Behavior
- Firecracker now uses `SnapshotType: Full` when no retained memory base exists.
  This covers the first standby after boot and retries after a failed snapshot discarded the retained base.
- Firecracker uses `SnapshotType: Diff` only when the retained memory base already exists.
  In the normal happy path, that means the second and later standby cycles reuse `snapshot-base` and only write changed pages.
- After a successful restore, `snapshot-latest` is moved back into the hidden `snapshot-base` so the next standby can reuse it.

## Behavior Changes
- `StopInstance` now removes both visible standby snapshots and any retained hidden base.
- On snapshot failure after promoting a retained base, the promoted directory is discarded instead of renamed back into place.
- Graceful VMM shutdown behavior is now driven by hypervisor capabilities, and unsupported graceful shutdown paths skip the wait and go straight to force-kill.

## Performance
Local benchmark harness from #145:
- Baseline: `score_ms 5146.5`, `standby_p50_ms 5032.0`
- Cleaned-up retained diff-base result: `score_ms 264.7`, `standby_p50_ms 157.0`

Integration timing run from `go test ./lib/instances -run TestFirecrackerStandbyAndRestore -v`:
- first standby, full snapshot expected: `3.33545878s`
- first restore to running: `101.817536ms`
- first restore to exec-ready: `1.618021087s`
- second standby, diff snapshot expected: `557.88755ms`
- second restore to running: `103.666293ms`
- second restore to exec-ready: `115.73631ms`

That integration test also verifies guest state persists across both cycles by writing one file before the first standby, another after the first restore, and asserting both files still exist after the second restore.

## Test Plan
- [x] `make test TEST=TestFirecrackerStandbyAndRestore`
- [x] `make test TEST=TestFirecrackerStopClearsStaleSnapshot`
- [x] `go test ./lib/hypervisor/...`
- [x] `go test ./lib/hypervisor/firecracker -run TestSnapshotParamPaths`
- [x] `go test ./lib/instances -run 'TestDiscardPromotedRetainedSnapshotTargetAfterSnapshotError'`
- [x] `go test ./lib/instances -run TestFirecrackerStandbyAndRestore -v`
- [x] benchmark locally with the harness from #145 via `prepare.go` + `run.go -budget 180s`

---

> [!NOTE]
> **Medium Risk**
> Touches core instance lifecycle (standby/restore/stop) and process-shutdown behavior, so regressions could leave stale snapshots or incorrectly kill/retain VMM processes. Changes are capability-gated and covered by expanded unit/integration tests, reducing but not eliminating risk.
> 
> **Overview**
> Speeds up Firecracker standby by **reusing a hidden retained snapshot base** across restore/standby cycles, so later standbys can write *diff* snapshots instead of full memory images; a new internal `snapshot-base` directory is introduced and `snapshot-latest` remains the only user-visible standby artifact.
> 
> Adds a **type-level hypervisor capability registry** (`RegisterCapabilities`/`CapabilitiesForType`) and extends `hypervisor.Capabilities` with `SupportsSnapshotBaseReuse` and `SupportsGracefulVMMShutdown`, which now drive orchestration decisions (retaining snapshot base on restore, cleaning it on stop, and skipping graceful shutdown waits for hypervisors that can’t do it).
> 
> Hardens snapshot lifecycle handling: promotes/rolls back retained bases around standby snapshot creation, moves `snapshot-latest` to `snapshot-base` after successful restore, and ensures both visible and hidden snapshot dirs are removed on stop; tests are updated/added to validate full-vs-diff snapshot selection and retained-base cleanup/rollback behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b0c4c171acb78acace382769586401089afd67b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/instances/standby.go
+++ b/lib/instances/standby.go
@@ -88,24 +88,24 @@ func (m *manager) standbyInstance(
 
 	// 7. Create snapshot
 	snapshotDir := m.paths.InstanceSnapshotLatest(id)
-	retainedBaseDir := m.paths.InstanceSnapshotFirecrackerBase(id)
+	retainedBaseDir := m.paths.InstanceSnapshotBase(id)
 	promotedRetainedBase := false
-	if stored.HypervisorType == hypervisor.TypeFirecracker {
+	if hv.Capabilities().SupportsSnapshotBaseReuse {
 		var err error
-		promotedRetainedBase, err = prepareFirecrackerSnapshotTarget(snapshotDir, retainedBaseDir)
+		promotedRetainedBase, err = prepareRetainedSnapshotTarget(snapshotDir, retainedBaseDir)
 		if err != nil {
 			_ = hv.Resume(ctx)
-			return nil, fmt.Errorf("prepare firecracker snapshot target: %w", err)
+			return nil, fmt.Errorf("prepare retained snapshot target: %w", err)
 		}
 	}
 	log.DebugContext(ctx, "creating snapshot", "instance_id", id, "snapshot_dir", snapshotDir)
-	if err := createSnapshot(ctx, hv, snapshotDir, stored.HypervisorType); err != nil {
+	if err := createSnapshot(ctx, hv, snapshotDir, hv.Capabilities().SupportsSnapshotBaseReuse); err != nil {
 		// Snapshot failed - try to resume VM
 		log.ErrorContext(ctx, "snapshot failed, attempting to resume VM", "instance_id", id, "error", err)
 		hv.Resume(ctx)
 		if promotedRetainedBase {
-			if rollbackErr := restoreFirecrackerRetainedBase(snapshotDir, retainedBaseDir); rollbackErr != nil {
-				log.WarnContext(ctx, "failed to restore firecracker retained snapshot base after snapshot error", "instance_id", id, "error", rollbackErr)
+			if rollbackErr := restoreRetainedSnapshotBase(snapshotDir, retainedBaseDir); rollbackErr != nil {
+				log.WarnContext(ctx, "failed to restore retained snapshot base after snapshot error", "instance_id", id, "error", rollbackErr)
 			}
 		}
 		return nil, fmt.Errorf("create snapshot: %w", err)
@@ -162,12 +162,12 @@ func (m *manager) standbyInstance(
 }
 
 // createSnapshot creates a snapshot using the hypervisor interface
-func createSnapshot(ctx context.Context, hv hypervisor.Hypervisor, snapshotDir string, hvType hypervisor.Type) error {
+func createSnapshot(ctx context.Context, hv hypervisor.Hypervisor, snapshotDir string, reuseSnapshotBase bool) error {
 	log := logger.FromContext(ctx)
 
-	// Firecracker diff snapshots can reuse the previous memory file as a base.
-	// Other hypervisors keep the existing behavior of recreating the directory.
-	if hvType != hypervisor.TypeFirecracker {
+	// Hypervisors that do not reuse an on-disk snapshot base keep recreating the
+	// directory before each snapshot.
+	if !reuseSnapshotBase {
 		os.RemoveAll(snapshotDir)
 	}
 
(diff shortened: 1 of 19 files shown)


lib/paths/paths.go
```diff
@@ -199,6 +199,11 @@ func (p *Paths) InstanceSnapshotLatest(id string) string {
 	return filepath.Join(p.InstanceSnapshots(id), "snapshot-latest")
 }
 
+// InstanceSnapshotFirecrackerBase returns the hidden retained Firecracker diff base.
+func (p *Paths) InstanceSnapshotFirecrackerBase(id string) string {
```
rgarcia: is there a way to make this not firecracker-specific? there's a chance that other hypervisors support diff snapshots so this path could be generic

lib/instances/stop.go
```diff
@@ -234,6 +234,10 @@ func (m *manager) stopInstance(
 	if err := os.RemoveAll(snapshotDir); err != nil {
 		log.WarnContext(ctx, "failed to remove stale snapshot directory on stop", "instance_id", id, "snapshot_dir", snapshotDir, "error", err)
 	}
+	retainedBaseDir := m.paths.InstanceSnapshotFirecrackerBase(id)
+	if err := os.RemoveAll(retainedBaseDir); err != nil {
+		log.WarnContext(ctx, "failed to remove retained firecracker snapshot base on stop", "instance_id", id, "snapshot_dir", retainedBaseDir, "error", err)
+	}
```
rgarcia, line 242: same comment here--would like to make this generic and not leak hypervisor information

lib/instances/standby.go
```diff
@@ -147,11 +162,14 @@ func (m *manager) standbyInstance(
...
+func createSnapshot(ctx context.Context, hv hypervisor.Hypervisor, snapshotDir string, hvType hypervisor.Type) error {
 	log := logger.FromContext(ctx)
 
-	// Remove old snapshot
-	os.RemoveAll(snapshotDir)
+	// Firecracker diff snapshots can reuse the previous memory file as a base.
+	// Other hypervisors keep the existing behavior of recreating the directory.
+	if hvType != hypervisor.TypeFirecracker {
+		os.RemoveAll(snapshotDir)
+	}
```
rgarcia, line 177: could this instead be modeled as a hypervisor capabilitity instead of conditioning on hypervisor.Type?

then the logic in standbyInstance() can be conditioned on this capability instead of HypervisorType

sjmiller609 (approved):

ryw:
love use of autoresearch :)

PR #148, opened 2026-03-17
Add scoped permissions for API keys

sjmiller609:
## Summary

- Adds a `lib/scopes` package that defines permission scopes following the `resource:action` pattern (e.g., `instance:read`, `image:write`, `build:delete`)
- JWT tokens can now carry a `permissions` claim with an array of scope strings; tokens without this claim retain full access (backward compatible)
- Auth middleware (`JwtAuth` and `OapiAuthenticationFunc`) extracts the `permissions` claim and stores parsed scopes in the request context
- A new scope enforcement middleware rejects requests missing the required scope with 403 Forbidden
- The `hypeman-token` CLI gains `-scopes` (comma-separated scope list) and `-list-scopes` flags
- WebSocket endpoints (`exec`, `cp`) enforce `instance:write` scope via `RequireScope`

## Scope mapping

All API routes are mapped to scopes in `lib/scopes/scopes.go:routeScopes`. The resource types are:

| Scope prefix | Resources |
|---|---|
| `instance:` | instances, exec, cp, start/stop/standby/restore/fork |
| `image:` | images |
| `volume:` | volumes, attach/detach |
| `snapshot:` | snapshots, instance snapshots |
| `build:` | builds |
| `device:` | devices |
| `ingress:` | ingresses |
| `resource:` | health, resources |

Each prefix supports `read`, `write`, and `delete` actions (except `resource:` which is read-only). The wildcard `*` grants all permissions.

## Usage

```bash
# Full-access token (backward compatible, no permissions claim)
hypeman-token -user-id myuser -duration 8760h

# Read-only token for instances and images
hypeman-token -user-id myuser -scopes "instance:read,image:read"

# Token that can manage instances but not delete them
hypeman-token -user-id myuser -scopes "instance:read,instance:write"

# List available scopes
hypeman-token -list-scopes
```

## Test plan

- [x] `go test ./lib/scopes/` — scope parsing, validation, context helpers, RequireScope middleware
- [x] `go test ./lib/middleware/` — legacy tokens get full access, scoped tokens have permissions in context, wildcard tokens work
- [x] `go vet` clean on all changed packages
- [ ] Integration test with live API: create a scoped token, verify allowed endpoints succeed and denied endpoints return 403

---

> [!NOTE]
> **Medium Risk**
> Introduces per-route authorization enforcement based on JWT `permissions` scopes; mistakes in scope mapping or middleware ordering could inadvertently block or expose API operations.
> 
> **Overview**
> Adds a new **scoped permissions** system for API-key JWTs via a `permissions` claim, including a `lib/scopes` package (scope definitions, route-to-scope mapping, and chi middleware) that returns `403 Forbidden` when required scopes are missing.
> 
> Updates API auth to extract and store scopes in request context (legacy tokens without `permissions` remain full-access; malformed claims become *deny-all*), wires scope enforcement into the main API router, and adds explicit `instance:write` gating for the WebSocket `exec`/`cp` endpoints.
> 
> Extends the `hypeman-token` generator CLI with `-scopes` and `-list-scopes`, and adds tests to ensure scope parsing/enforcement works and that every registered API route is mapped (or intentionally public).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd33c361613097bd9226e69ffb206102533d2abd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/scopes/scopes.go
```diff
@@ -0,0 +1,245 @@
+// Package scopes defines API permission scopes for hypeman API keys.
+//
+// Scopes follow the pattern "resource:action" where resource is one of the
+// API resource types and action is read, write, or delete. Tokens without
+// a "permissions" claim are treated as having full access for backward
+// compatibility with existing tokens.
```
sjmiller609, line 6: 👍

lib/scopes/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+			if pattern == "" {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			required, ok := ScopeForRoute(r.Method, pattern)
+			if !ok {
+				// Route not mapped — allow through
+				next.ServeHTTP(w, r)
+				return
```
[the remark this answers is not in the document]
sjmiller609: This analysis is incorrect for our usage pattern. chi resolves the route **before** running middleware within a `r.Group()` block — `RoutePattern()` is populated by the time our middleware executes. I verified this empirically: when middleware is registered with `r.Use()` inside a group that also registers routes (which is exactly what `oapi.HandlerWithOptions` does with `r.Group(func(r chi.Router) { r.Get(...) })`), chi matches the route first and then runs the middleware chain.

To prove this definitively, I added `TestMiddleware_EnforcesScopes` in `lib/scopes/middleware_test.go` — an integration test using a real chi router that confirms the middleware blocks requests with insufficient scopes and allows those with matching scopes. All 7 test cases pass.

lib/middleware/oapi_auth.go
```diff
@@ -480,8 +485,37 @@ func JwtAuth(jwtSecret string) func(http.Handler) http.Handler {
...
+	raw, ok := claims["permissions"]
+	if !ok {
+		return ctx // no permissions claim — legacy full-access token
+	}
+
+	// The claim is a JSON array of strings, which jwt.MapClaims decodes
+	// as []interface{}.
+	arr, ok := raw.([]interface{})
+	if !ok {
+		return ctx
```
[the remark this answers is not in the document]
sjmiller609: Valid finding — fixed in 1f3e39e. `extractPermissions` now stores an empty permission set (`[]scopes.Scope{}`) when the `permissions` claim is present but not a valid JSON array. This means a malformed claim denies all access instead of granting full access. Added a test case (`malformed permissions claim denies all`) to verify the fix.

lib/scopes/scopes.go
```diff
@@ -0,0 +1,255 @@
...
+func ParseScopes(s string) ([]Scope, error) {
+	if s == "" {
+		return nil, nil
+	}
+	parts := strings.Split(s, ",")
+	out := make([]Scope, 0, len(parts))
+	for _, p := range parts {
+		sc := Scope(strings.TrimSpace(p))
+		if !sc.Valid() {
+			return nil, fmt.Errorf("unknown scope: %q", p)
```
masnwilliams: nit: this shows the untrimmed input — if someone passes ` "bogus:thing"` the error includes the leading space which makes it look like the space is the problem. consider using sc instead of p here

lib/scopes/scopes.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// Resource/health scopes (read-only)
+	ResourceRead Scope = "resource:read"
+
+	// Wildcard scope — grants all permissions
+	All Scope = "*"
+)
+
+// AllScopes is the complete list of valid scopes (excluding wildcard).
+var AllScopes = []Scope{
```
masnwilliams: nit: exported mutable slice — anything can append/modify at runtime. low risk for an internal package, but worth a comment noting it's treated as immutable (or make it a func returning a copy)

lib/scopes/scopes.go
```diff
@@ -0,0 +1,255 @@
...
+	"POST /instances/{id}/standby":                        InstanceWrite,
+	"POST /instances/{id}/start":                          InstanceWrite,
+	"GET /instances/{id}/stat":                            InstanceRead,
+	"GET /instances/{id}/stats":                           InstanceRead,
+	"POST /instances/{id}/stop":                           InstanceWrite,
+	"DELETE /instances/{id}/volumes/{volumeId}":           VolumeWrite,
+	"POST /instances/{id}/volumes/{volumeId}":             VolumeWrite,
+
+	// WebSocket endpoints (outside OpenAPI but still authed)
+	"GET /instances/{id}/exec": InstanceWrite,
```
masnwilliams: these RouteScopes entries for exec/cp are never checked by Middleware() — they're outside the OpenAPI router group and use RequireScope directly. they only exist to satisfy the route coverage test. could move them to PublicRoutes and adjust the test, so the map only contains entries actually used by the middleware

lib/middleware/oapi_auth.go
```diff
@@ -480,8 +485,40 @@ func JwtAuth(jwtSecret string) func(http.Handler) http.Handler {
...
+	arr, ok := raw.([]interface{})
+	if !ok {
+		// permissions claim present but not a valid array — deny all
+		return scopes.ContextWithPermissions(ctx, []scopes.Scope{})
+	}
+
+	perms := make([]scopes.Scope, 0, len(arr))
+	for _, v := range arr {
+		if s, ok := v.(string); ok {
+			perms = append(perms, scopes.Scope(s))
```
masnwilliams: no Valid() check on extracted scopes — invalid strings silently fail to match which is safe (fail-closed), but a typo in token generation (e.g. "instnace:read") won't surface until a user hits a 403. might be worth logging or validating here

lib/middleware/oapi_auth_test.go
```diff
@@ -335,3 +336,98 @@ func TestJwtAuth_RequiresAuthorization(t *testing.T) {
...
+		"exp":         time.Now().Add(time.Hour).Unix(),
+		"permissions": perms,
+	}
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	tokenString, err := token.SignedString([]byte(testJWTSecret))
+	require.NoError(t, err)
+	return tokenString
+}
+
+func TestJwtAuth_ScopedPermissions(t *testing.T) {
```
masnwilliams, line 357: nice coverage on JwtAuth with scoped tokens — worth adding a test for the OapiAuthenticationFunc path with scoped tokens too, since that's the other entry point for auth

masnwilliams (approved):
solid design — clean scope system, good backward compat (nil = full access), fail-closed on malformed claims, and the route coverage tests are a great guardrail. left a few nits, nothing blocking.

lib/scopes/scopes.go
sjmiller609: Fixed — error message now uses the trimmed variable `sc` instead of the raw input `p`.

lib/scopes/scopes.go
sjmiller609: Converted `AllScopes` from an exported `var` to a function that returns a copy of the internal slice. Callers now use `AllScopes()`. The unexported `allScopes` slice is used directly by `Valid()` to avoid allocation.

lib/scopes/scopes.go
sjmiller609: Moved exec/cp entries out of `RouteScopes` into a new `DirectScopeRoutes` map for routes that enforce scopes via `RequireScope` directly (outside the OpenAPI middleware group). Updated the route coverage test to check `DirectScopeRoutes` as well, and added a test ensuring `DirectScopeRoutes` and `RouteScopes` don't overlap.

lib/middleware/oapi_auth.go
sjmiller609: Added a `Valid()` check on each scope during extraction. Invalid scopes now log a warning via `log.WarnContext` so typos like `"instnace:read"` surface in logs instead of silently failing to match.

lib/middleware/oapi_auth_test.go
sjmiller609: Added `TestOapiAuthenticationFunc_ScopedPermissions` with four subcases: legacy token (full access), scoped token (specific permissions in context), wildcard token (grants all), and malformed permissions claim (denies all). Uses a helper `newBearerAuthInput` to construct the `openapi3filter.AuthenticationInput` directly.

PR #149, opened 2026-03-18
Add optional snapshot compression defaults and standby integration

sjmiller609:
## Summary
- add optional snapshot compression config (zstd/lz4) to snapshot and instance APIs
- add server-level and per-instance snapshot compression defaults with precedence: request override > instance policy > server default
- add async snapshot memory compression jobs for standby and centrally stored standby snapshots
- add restore-time snapshot memory preparation (cancel/wait active compression + on-demand decompression)
- include compression metadata/state on snapshot resources

## API / Config
- `POST /instances/{id}/standby` accepts optional `compression`
- `POST /instances/{id}/snapshots` accepts optional `compression`
- `POST /instances` accepts optional `snapshot_policy`
- config adds `snapshot.compression_default` (`enabled`, `algorithm`, `level`)

## Validation and behavior
- compression is standby-only (rejected for stopped snapshots)
- zstd level validated (1-19)
- lz4 rejects level input
- standby validates compression policy before state transitions

## Tests
- `go test -run '^$' ./cmd/api/api ./lib/instances ./lib/providers ./integration`
- `go test ./lib/instances -run 'TestNormalizeCompressionConfig|TestResolveSnapshotCompressionPolicyPrecedence|TestValidateCreateRequestSnapshotPolicy|TestValidateCreateSnapshotRequestRejectsStoppedCompression'`

---

> [!NOTE]
> **High Risk**
> High risk because it changes core instance lifecycle paths (standby/restore/fork/snapshot/delete) to manage asynchronous compression jobs and on-demand decompression, which can affect VM correctness, performance, and snapshot data integrity.
> 
> **Overview**
> Adds optional snapshot compression (*zstd* or *lz4*) across the API and instance domain model: `POST /instances` can set `snapshot_policy`, `POST /instances/{id}/standby` and `POST /instances/{id}/snapshots` accept per-request compression overrides, and instance/snapshot responses now include compression policy + compression state/size metadata.
> 
> Implements asynchronous snapshot-memory compression for standby instances and centrally stored standby snapshots with configurable defaults (server `snapshot.compression_default`, per-instance policy, then request override). Foreground operations (`restore`, `fork`, `create snapshot`, `delete`) now cancel/wait in-flight compression and ensure raw memory is available (including on-demand decompression) before proceeding.
> 
> Updates plumbing and tests accordingly: `instances.NewManager` takes snapshot defaults, `StandbyInstance` takes a request object, adds OpenTelemetry metrics for compression jobs/fallbacks/preemptions, skips copying `.zst.tmp`/`.lz4.tmp` artifacts during forks, and introduces middleware to normalize empty standby POST bodies to `{}` for strict request decoding.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8638cce42f987c6efe174e990553dd23abdf82a6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


skills/hypeman-remote-linux-tests/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+go test ./lib/instances -run TestQEMUStandbyRestoreCompressionScenarios -count=1 -v
+```
+
+For unit-only validation:
+
+```bash
+go test ./lib/instances -run 'TestNormalizeCompressionConfig|TestResolveSnapshotCompressionPolicyPrecedence' -count=1
+```
+
+Run one hypervisor at a time when the tests boot real VMs. This avoids noisy failures from competing KVM guests and makes logs much easier to interpret.
```
sjmiller609:
```suggestion
```

skills/hypeman-remote-linux-tests/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Reporting Guidance
+
+When reporting results, separate:
+
+1. Code failures
+2. Test harness issues
+3. Remote host environment issues
+
+This matters in Hypeman because Linux integration tests often fail before product code executes if the remote machine is missing bundled binaries, system tools, or short enough paths.
```
sjmiller609:
```suggestion
```

openapi.yaml
```diff
@@ -305,6 +305,8 @@ components:
           enum: [cloud-hypervisor, firecracker, qemu, vz]
           description: Hypervisor to use for this instance. Defaults to server configuration.
           example: cloud-hypervisor
+        snapshot_policy:
+          $ref: "#/components/schemas/SnapshotPolicy"
```
sjmiller609, line 325: you can specify the snapshot compression to use for the instance when you create it.

openapi.yaml
```diff
@@ -433,6 +483,14 @@ components:
           example: pre-upgrade
         tags:
           $ref: "#/components/schemas/Tags"
+        compression:
+          $ref: "#/components/schemas/SnapshotCompressionConfig"
```
sjmiller609, line 506: you can also specify the compression when you request the snapshot

cmd/api/api/optional_standby_body.go
```diff
@@ -0,0 +1,36 @@
...
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+)
+
+var standbyRoutePattern = regexp.MustCompile(`^/instances/[^/]+/standby$`)
+
+// NormalizeOptionalStandbyBody rewrites empty standby POST bodies to "{}"
+// so the generated strict handler can decode them without special casing.
```
sjmiller609, line 10: this seems a bit odd, but I did decide I like this option best in comparison to having a breaking change. I.e. we have to have this middleware injecting empty payload in order for "no payload" to work with our existing input validation type checking

lib/instances/snapshot_compression.go
```diff
@@ -0,0 +1,560 @@
...
+func snapshotMemoryFileCandidates(snapshotDir string) []string {
+	return []string{
+		filepath.Join(snapshotDir, "memory-ranges"),
+		filepath.Join(snapshotDir, "memory"),
+		filepath.Join(snapshotDir, "snapshots", "snapshot-latest", "memory-ranges"),
+		filepath.Join(snapshotDir, "snapshots", "snapshot-latest", "memory"),
+	}
+}
+
+func compressSnapshotMemoryFile(ctx context.Context, rawPath string, cfg snapshotstore.SnapshotCompressionConfig) (int64, int64, error) {
```
sjmiller609, line 502: I'm actually going to add a fallback behavior intentionally: use native compression where possible, fallback to go-only compression if native tools aren't installed, emitting warning logs. That's worth it because native is faster

lib/instances/snapshot_compression.go
```diff
@@ -149,6 +173,125 @@ func (m *manager) snapshotJobKeyForSnapshot(snapshotID string) string {
...
+		"error", err,
+		"fallback_backend", "go",
+		"install_tip", nativeCodecInstallTip(algorithm),
+	)
+	if runtime.manager != nil {
+		runtime.manager.recordSnapshotCodecFallback(ctx, algorithm, operation, reason)
+	}
+}
+
+func runWithNativeFallback(ctx context.Context, runtime nativeCodecRuntime, algorithm snapshotstore.SnapshotCompressionAlgorithm, operation snapshotCodecOperation, nativeRunner func(binaryPath string) error, goRunner func() error) error {
```
sjmiller609, line 273: intentionally added fallback

cmd/api/api/instances.go
```diff
@@ -985,3 +1017,53 @@ func instanceToOAPI(inst instances.Instance) oapi.Instance {
 
 	return oapiInst
 }
+
+func toDomainSnapshotCompressionConfig(cfg oapi.SnapshotCompressionConfig) (*snapshot.SnapshotCompressionConfig, error) {
```
masnwilliams, line 1021: this accepts any algorithm string and any level value — the error return is dead code (always nil). a client can send `algorithm: "brotli"` or `level: 999` and it passes through unchecked. needs server-side validation here for algorithm ∈ {zstd, lz4} and per-algorithm level ranges, especially since the openapi schema `level.maximum: 19` is only correct for zstd (lz4 max is 9).

lib/instances/snapshot_compression.go
```diff
@@ -0,0 +1,839 @@
...
+		rawPath, ok := findRawSnapshotMemoryFile(target.SnapshotDir)
+		if !ok {
+			if compressedPath, algorithm, found := findCompressedSnapshotMemoryFile(target.SnapshotDir); found && target.SnapshotID != "" {
+				cfg := compressionMetadataForExistingArtifact(target.Policy, algorithm)
+				var compressedSizeBytes *int64
+				if st, statErr := os.Stat(compressedPath); statErr == nil {
+					size := st.Size()
+					compressedSizeBytes = &size
+				}
+				_ = m.updateSnapshotCompressionMetadata(target.SnapshotID, snapshotstore.SnapshotCompressionStateCompressed, "", &cfg, compressedSizeBytes, nil)
```
masnwilliams: metadata update errors are silently discarded at all 4 call sites in this function. if `saveSnapshotRecord` fails (disk full, I/O error), the on-disk compression state desyncs from the filesystem — compressed file exists but metadata says "none", or vice versa. at minimum these should `log.Error` so the failure is observable.

cmd/api/config/config.go
```diff
@@ -446,8 +468,33 @@ func (c *Config) Validate() error {
 	if c.Build.Timeout <= 0 {
 		return fmt.Errorf("build.timeout must be positive, got %d", c.Build.Timeout)
 	}
+	algorithm := strings.ToLower(c.Snapshot.CompressionDefault.Algorithm)
```
masnwilliams, line 497: nit: validation lowercases the algorithm here but the config struct stores the original casing. if downstream code reads `c.Snapshot.CompressionDefault.Algorithm` without lowercasing, `"Zstd"` passes validation but won't match domain constants. consider normalizing to lowercase in the struct after validation.

lib/providers/providers_test.go
```diff
@@ -0,0 +1,71 @@
...
+				Level:     intPtr(7),
+			},
+		},
+	}
+
+	defaults := snapshotDefaultsFromConfig(cfg)
+	assert.Nil(t, defaults.Compression)
+}
+
+func TestSnapshotDefaultsFromConfigOmitsLevelForLZ4(t *testing.T) {
```
masnwilliams: nit: test name says "omits level" but the assertion verifies level IS preserved (`Level == 7`). name and behavior don't match.

masnwilliams (commented):
solid feature — async compression design is clean, cancellation at every lifecycle point is thorough, and the standby body middleware is a nice backward-compat solution. main concern is the missing server-side validation for compression config (algorithm + level). see inline comments.

masnwilliams (approved):
approving but see comments above

lib/instances/snapshot_compression.go
sjmiller609: Addressed in `3b9264e`. The four metadata updates now log an error instead of being silently discarded, and the logs include the snapshot id, snapshot dir, and target compression state so we can see when the record and filesystem drift.

cmd/api/config/config.go
sjmiller609: Addressed on the branch as well. Config validation now normalizes `snapshot.compression_default.algorithm` back into the struct after validation, and the config test covers mixed-case input so downstream code always sees the canonical lowercase value.

lib/instances/snapshot.go
```diff
@@ -335,6 +386,14 @@ func (m *manager) forkSnapshot(ctx context.Context, snapshotID string, req ForkS
...
 	defer cu.Clean()
 
+	target, err := m.cancelAndWaitCompressionJob(ctx, m.snapshotJobKeyForSnapshot(snapshotID))
+	if err != nil {
+		return nil, fmt.Errorf("wait for snapshot compression to stop: %w", err)
+	}
+	if target != nil {
+		m.recordSnapshotCompressionPreemption(ctx, snapshotCompressionPreemptionForkSnapshot, *target)
+	}
+
```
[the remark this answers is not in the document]
sjmiller609: Addressed in `3b9264e`. `forkSnapshot` now materializes raw memory with `ensureSnapshotMemoryReady(...)` before copying the snapshot payload, and there is a regression test for forking from a fully compressed source snapshot.

cmd/api/api/instances.go
sjmiller609: Addressed earlier on this branch. The API layer now validates the requested snapshot compression algorithm and per-algorithm level range server-side before converting to the domain type, so invalid values like `brotli` or out-of-range levels are rejected instead of flowing through.

lib/providers/providers_test.go
sjmiller609: Addressed as part of the provider test cleanup. The test name/behavior now match again so it is clear whether the level is expected to be preserved or omitted for the given config path.

lib/instances/fork.go
```diff
@@ -421,7 +421,7 @@ func (m *manager) applyForkTargetState(ctx context.Context, forkID string, targe
 			if _, err := m.startInstance(ctx, forkID, StartInstanceRequest{}); err != nil {
 				return nil, fmt.Errorf("start forked instance for standby transition: %w", err)
 			}
-			return returnWithReadiness(m.standbyInstance(ctx, forkID))
+			return returnWithReadiness(m.standbyInstance(ctx, forkID, StandbyInstanceRequest{}, false))
```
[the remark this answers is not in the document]
sjmiller609: Addressed in `8638cce`. `forkInstanceFromStoppedOrStandby` now prepares standby snapshot memory before copying the source guest directory by canceling any active instance compression job and materializing raw memory if needed. I also added `TestForkInstanceFromStandbyCancelsCompressionJobAndCopiesRawMemory` to cover the in-flight standby compression case.

lib/instances/snapshot_compression.go
```diff
@@ -0,0 +1,847 @@
...
+func (m *manager) ensureSnapshotMemoryReady(ctx context.Context, snapshotDir, jobKey string, hvType hypervisor.Type) error {
+	start := time.Now()
+
+	if jobKey != "" {
+		target, err := m.cancelAndWaitCompressionJob(ctx, jobKey)
+		if err != nil {
+			return err
+		}
+		if target != nil {
+			m.recordSnapshotCompressionPreemption(ctx, snapshotCompressionPreemptionRestoreInstance, *target)
```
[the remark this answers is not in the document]
sjmiller609: Addressed in `a560a76`. `ensureSnapshotMemoryReady(...)` now takes the preemption operation as an explicit argument, so standby fork preemptions are recorded as `fork_instance` instead of being hardcoded to `restore_instance`. I also added `TestEnsureSnapshotMemoryReadyRecordsProvidedPreemptionOperation` to lock the label behavior down.

PR #158, opened 2026-03-19
feat: support updating egress proxy secret envs for key rotation

sjmiller609:
## Summary
- Adds `PATCH /instances/{id}` API endpoint for updating mutable instance properties
- Adds `UpdateInstanceRules()` to egress proxy service for live header inject rule updates
- Adds `UpdateInstance()` to instance manager with `UpdateInstanceRequest` type
- Enables credential/secret key rotation without instance restart — the host-side egress proxy recompiles header inject rules in-place with the new secret values
- Extends egress proxy integration test to verify key rotation flow end-to-end

## How it works
The egress proxy runs on the host and intercepts outbound HTTPS traffic from the guest. Credential secrets are never exposed to the guest — the guest only sees mock placeholder values. When secrets are updated via the API:

1. New env values are merged into instance metadata
2. Credential env bindings are re-validated
3. Header inject rules are recompiled from the updated env + existing credential policies
4. The egress proxy's in-memory rules are swapped atomically via `UpdateInstanceRules()`

No guest restart or signaling is needed because the proxy operates entirely on the host side.

## Test plan
- Extended `TestEgressProxyRewritesHTTPSHeaders` integration test to include:
  - Update credential env value via `UpdateInstance()`
  - Verify guest-visible env still shows mock placeholder (secrets never leak)
  - Verify subsequent HTTPS requests use the rotated credential value
  - Verify old credential value is no longer injected
- `go build ./...` passes
- `go vet ./...` passes
- All egress proxy unit tests pass
- All builds package unit tests pass
- CI must pass

---

> [!NOTE]
> **Medium Risk**
> Adds a new instance mutation API and runtime egress-proxy rule updates; incorrect validation/rollback could leave proxies injecting stale or inconsistent credentials for running VMs.
> 
> **Overview**
> Adds `PATCH /instances/{id}` to update an instance’s env vars (limited to credential-policy `source.env` keys) to support **live secret/key rotation** without restarting the VM.
> 
> Implements `manager.UpdateInstance` with validation, metadata persistence, and **egress proxy rule recompilation** plus rollback if saving metadata fails; the egress proxy gains `UpdateInstanceRules` and improved logging/tracing/metrics around register/update/unregister and request handling.
> 
> Updates OpenAPI/stainless specs, scopes, and tests (new unit tests and an integration test that rotates a key and verifies the guest still sees mock env while outbound HTTPS uses the rotated value).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5bdd7a625fb817f21a49dbb2935c261b47cbcb20. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lib/instances/update.go
```diff
@@ -0,0 +1,122 @@
...
+// which causes the egress proxy header inject rules to be recomputed
+// with the new secret values — enabling key rotation without restart.
+func (m *manager) updateInstance(ctx context.Context, id string, req UpdateInstanceRequest) (*Instance, error) {
+	log := logger.FromContext(ctx)
+
+	// 1. Load and validate current state
+	meta, err := m.loadMetadata(id)
+	if err != nil {
+		return nil, ErrNotFound
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed on f409273. updateInstance now logs and propagates real loadMetadata errors instead of rewriting them to ErrNotFound, so corrupted or unreadable metadata returns the correct internal error path instead of a misleading 404.

lib/instances/update.go
```diff
@@ -0,0 +1,122 @@
...
+				return nil, fmt.Errorf("save metadata: %w (failed to roll back egress proxy rules: %v)", err, rollbackErr)
+			}
+			log.WarnContext(ctx, "rolled back egress proxy header inject rules after metadata save failure", "instance_id", id, "error", err)
+			return nil, fmt.Errorf("save metadata: %w", err)
+		}
+	} else {
+		if err := m.saveMetadata(meta); err != nil {
+			return nil, fmt.Errorf("save metadata: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed on f409273. The env patch is now staged in a separate map (`nextEnv`) and only assigned to `meta.Env` after the proxy rule update succeeds. If metadata persistence fails, the code restores `meta.Env` to `prevEnv` after rolling the runtime rules back as well.

lib/egressproxy/service.go
```diff
@@ -474,3 +608,46 @@ func sourceIPFromRemoteAddr(remoteAddr string) string {
...
+		log := logger.FromContext(ctx)
+		if log != nil && log != slog.Default() {
+			return log
+		}
+	}
+	if s.log != nil {
+		return s.log
+	}
+	return slog.Default()
+}
```
sjmiller609, line 629: we should just use the logger from context, not falling back

lib/egressproxy/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+
+	registeredInstances, err := meter.Int64ObservableGauge(
+		"hypeman_egress_proxy_registered_instances_total",
+		metric.WithDescription("Total number of instances currently registered with the egress proxy"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	controlPlaneDuration, err := meter.Float64Histogram(
```
masnwilliams, line 50: the default OTel SDK histogram buckets are millisecond-oriented (`[0, 5, 10, 25, 50, ...]`) but these histograms record in seconds via `WithUnit("s")`. in practice, nearly all observations will land in the first bucket — control plane ops are ~1-50ms (0.001-0.05s), upstream requests ~10ms-5s.

consider adding explicit bucket boundaries so you get useful latency distribution data, e.g.:
- control plane: `0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 5`
- upstream: `0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30, 60`

ref: https://github.com/open-telemetry/opentelemetry-go/issues/5821

lib/egressproxy/service.go
```diff
@@ -349,10 +474,16 @@ func (s *Service) handleConnect(w http.ResponseWriter, r *http.Request, sourceIP
...
 		removeHopByHopHeaders(req.Header)
-		s.applyHeaderInjections(sourceIP, targetHost, req.Header, true)
+		injected := s.applyHeaderInjections(sourceIP, targetHost, req.Header, true) > 0
 
+		start := time.Now()
 		resp, err := s.transport.RoundTrip(req)
 		if err != nil {
+			log := s.loggerForContext(r.Context())
+			log.WarnContext(r.Context(), "egress proxy upstream request failed", "protocol", "https", "injected", injected, "error", err)
+			s.metrics.recordRequest(r.Context(), "https", "upstream_error", injected)
```
masnwilliams, line 490: metrics are well-covered for failures inside `transport.RoundTrip` (DNS/TCP/TLS to target all land here). but the CONNECT setup phase above this loop — specifically line 441 where `200 Connection Established` is sent before leaf cert creation and client TLS handshake — has no metrics or logging. if cert gen (line 443) or the client handshake (line 451) fail, the client sees a successful CONNECT followed by a dead connection with no signal.

might be worth adding a metric or log for cert-gen/client-handshake failures in the setup phase — they would be rare but hard to diagnose in production without any observability.

lib/instances/update_test.go
```diff
@@ -0,0 +1,54 @@
+package instances
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateUpdateInstanceRequest(t *testing.T) {
```
masnwilliams, line 13: these validation tests are solid — might be worth adding unit tests for the `updateInstance` flow itself. the rollback path (metadata save failure → revert proxy rules), the `svc == nil` branch, and the double-failure case (rollback also fails) all have meaningful branching logic. the integration test covers the happy path e2e, but a unit test with mocked dependencies would lock down the error paths.

openapi.yaml
```diff
@@ -1718,7 +1733,60 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Error"
+        404:
+          description: Instance not found
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        409:
+          description: Invalid state (instance must be running)
```
masnwilliams: nit: this says "instance must be running" but the code also allows `StateInitializing` (update.go line 31). might want to update the description to "instance must be running or initializing" to match.

masnwilliams (commented):
solid PR — credential rotation with rollback semantics is well-designed. the integration test proving secrets never leak to the VM is the highlight. main feedback is around histogram bucket config for useful latency data and a minor observability gap in the CONNECT setup path.

openapi.yaml
sjmiller609: Updated on 9dc7e07. The OpenAPI 409 response description now says "instance must be running or initializing" so it matches the `updateInstance` state check.

lib/egressproxy/metrics.go
sjmiller609: Updated on 9dc7e07. `hypeman_egress_proxy_control_plane_duration_seconds` and `hypeman_egress_proxy_upstream_duration_seconds` now set explicit bucket boundaries in seconds, using the ranges you suggested so the latency histograms are useful at our observed scale.

lib/egressproxy/service.go
sjmiller609: Updated on 9dc7e07. The CONNECT setup path now emits both a warning log and `hypeman_egress_proxy_requests_total` outcomes for the two failure stages above `RoundTrip`: `connect_cert_error` for leaf cert generation and `connect_handshake_error` for the client TLS handshake.

lib/instances/update_test.go
sjmiller609: Updated on 9dc7e07. I pulled the branching logic into `applyUpdatedInstanceEnv(...)` and added unit tests for the three cases you called out: `svc == nil`, metadata save failure with proxy-rule rollback, and the double-failure path where rollback also fails.

lib/instances/update.go
```diff
@@ -0,0 +1,147 @@
...
+	for k, v := range req.Env {
+		nextEnv[k] = v
+	}
+
+	if err := validateCredentialEnvBindings(meta.Credentials, nextEnv); err != nil {
+		return nil, err
+	}
+
+	if err := applyUpdatedInstanceEnv(ctx, log, id, meta, prevEnv, nextEnv, m.saveMetadata, m.getEgressProxyIfExists()); err != nil {
```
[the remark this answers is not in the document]
sjmiller609: Fixed on ae7b94a. `updateInstance` now converts the concrete `*egressproxy.Service` through `updateInstanceRulesServiceOrNil(...)` before passing it into the helper, so a nil proxy pointer stays nil instead of becoming a non-nil interface wrapper. I also added `TestUpdateInstanceRulesServiceOrNil` to lock in that regression case.

masnwilliams (approved):
lgtm

PR #160, opened 2026-03-19
feat: add active ballooning reclaim controller

sjmiller609:
## Summary
- add a host-side active ballooning controller in `lib/guestmemory` with pressure sampling, proportional reclaim, protected floors, and manual reclaim holds
- expose `POST /resources/memory/reclaim`, wire the controller through the API startup path, and document/configure the new `hypervisor.memory.active_ballooning` settings
- extend the existing guest-memory integration tests for Cloud Hypervisor, QEMU, Firecracker, and VZ to validate runtime balloon targets and manual reclaim flows

## Validation
- `go test ./lib/guestmemory -count=1`
- `go test ./cmd/api/api -run 'TestReclaimMemory_' -count=1`
- `make test-guestmemory-vz`
- `make test-guestmemory-linux` on `deft-kernel-dev` from `/home/sjmiller609/codex-active-ballooning-plan/hypeman`

---

> [!NOTE]
> **Medium Risk**
> Introduces a new background control loop that can change VM memory balloon targets across hypervisors and adds a new API endpoint; mistakes could cause unexpected VM memory pressure or reclaim behavior.
> 
> **Overview**
> Adds an **active guest-memory ballooning controller** (`lib/guestmemory`) that samples host pressure (Linux `/proc` + PSI, macOS `vm_stat`/`memory_pressure`), computes a reclaim target with hysteresis/protected floors, and applies proportional runtime balloon target changes with cooldown/step limits, plus metrics/tracing/logging.
> 
> Exposes a new **manual reclaim API** `POST /resources/memory/reclaim` (with `hold_for`, `dry_run`, and feature-disabled vs internal error handling), wires the controller into API DI/startup, and extends config with `hypervisor.memory.active_ballooning` defaults + validation.
> 
> Enables runtime balloon control across hypervisors by extending the `hypervisor.Hypervisor` interface (`Set/GetTargetGuestMemoryBytes` + `SupportsBalloonControl`), implementing it for Cloud Hypervisor/QEMU/Firecracker/VZ (including `vz-shim` balloon endpoints), and updates integration tests/Makefile to cover runtime target changes and reduce CI flakiness (timeouts, per-test runs, better artifact logging, PID resolution).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 66518592f16b9ae3b47bc2b57a77d298c293b939. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sjmiller609:
Validated the feature end to end on `deft-kernel-dev` with fresh uncached runs.

What I checked:
- Booted real VMs under Cloud Hypervisor, QEMU, and Firecracker on `deft`
- Verified the guest-side helper came up far enough for runtime memory control to work
- Verified Hypeman could read each VM's current runtime memory target over the real hypervisor control interface
- Verified host-triggered reclaim changed the balloon target, respected the protected floor, and returned the VM to its normal target after clearing reclaim
- Reran the host-side controller/API checks for proactive reclaim and pressure-state behavior on `deft`

The main issue I had to fix was that the manual Linux path could reuse a stale non-Linux embedded `guest-agent` artifact after syncing from this laptop, which caused the guest to shut down early with `exec format error`. I also tightened Linux runtime PID handling around the manual guest-memory validation path so the tests follow the live VMM process more reliably.

lib/guestmemory/active_ballooning.go
```diff
@@ -0,0 +1,227 @@
...
+	if c.PollInterval <= 0 {
+		c.PollInterval = d.PollInterval
+	}
+	if c.PressureHighWatermarkAvailablePercent <= 0 || c.PressureHighWatermarkAvailablePercent >= 100 {
+		c.PressureHighWatermarkAvailablePercent = d.PressureHighWatermarkAvailablePercent
+	}
+	if c.PressureLowWatermarkAvailablePercent <= 0 || c.PressureLowWatermarkAvailablePercent >= 100 {
+		c.PressureLowWatermarkAvailablePercent = d.PressureLowWatermarkAvailablePercent
+	}
+	if c.PressureLowWatermarkAvailablePercent <= c.PressureHighWatermarkAvailablePercent {
```
masnwilliams, line 64: nit: this cascading reset is a bit fragile — e.g. `High=20, Low=15` silently resets both to defaults even though only the relationship is wrong. simpler to just reset both explicitly or clamp `Low = max(Low, High+1)` to respect the user's intent where possible.

lib/guestmemory/controller.go
```diff
@@ -0,0 +1,361 @@
...
+		logFromContext(ctx, c.log).InfoContext(ctx,
+			"guest memory manual reclaim hold expired",
+			"operation", "manual_reclaim",
+		)
+		state.manualHold = nil
+	}
+
+	if req.force && !req.dryRun {
+		switch {
+		case req.requestedReclaim <= 0 && req.holdFor <= 0:
```
masnwilliams: nit: this branch and the one below (`requestedReclaim > 0 && holdFor <= 0`) do the exact same thing — both clear the hold. could collapse to a single `if req.holdFor <= 0 { clear }` branch.

lib/guestmemory/controller.go
```diff
@@ -0,0 +1,361 @@
...
+					"instance_id", candidate.vm.ID,
+					"hypervisor", candidate.vm.HypervisorType,
+					"previous_target_guest_memory_bytes", candidate.currentTargetGuestBytes,
+					"target_guest_memory_bytes", appliedTarget,
+					"error", err,
+				)
+				resp.Actions = append(resp.Actions, action)
+				continue
+			}
+			state.lastApplied[candidate.vm.ID] = now
```
masnwilliams, line 294: `lastApplied` entries are never pruned — grows with VM churn over time. low risk in practice (bounded by VMs per host) but worth a cleanup pass each reconcile cycle to remove entries for VMs no longer in the candidate list.

lib/guestmemory/controller.go
```diff
@@ -0,0 +1,361 @@
...
+	now := time.Now()
+	sampleCtx, sampleSpan := c.startChildSpan(ctx, "guestmemory.sample_host_pressure")
+	sample, err := c.sampler.Sample(sampleCtx)
+	if err != nil {
+		c.metrics.RecordSamplerError(ctx, "host_pressure")
+		sampleSpan.RecordError(err)
+		sampleSpan.SetStatus(codes.Error, err.Error())
+		sampleSpan.End()
+		c.recordReconcileError(ctx, trigger, start, span, err)
+		return ManualReclaimResponse{}, err
```
masnwilliams: consider wrapping this: `fmt.Errorf("sample host pressure: %w", err)` — callers (especially the API handler) won't have context on what failed otherwise.

lib/hypervisor/socket_pid_linux.go
```diff
@@ -0,0 +1,127 @@
...
+	"strconv"
+	"strings"
+)
+
+// ResolveProcessPID finds the process currently holding the listening Unix
+// socket for the given hypervisor control path.
+func ResolveProcessPID(socketPath string) (int, error) {
+	socketRef, err := socketRefForPath(socketPath)
+	if err == nil {
+		if pid, err := pidBySocketRef(socketRef); err == nil {
```
masnwilliams: the `:=` here shadows the outer `err` from `socketRefForPath`. if `pidBySocketRef` fails, that error is lost — and downstream, the `if err != nil` check on line ~28 references the `pidByCmdline` error instead. the final `return 0, fmt.Errorf(...)` also appears to be dead code since `err != nil` is always true at that point. consider capturing errors explicitly to preserve diagnostic info.

lib/hypervisor/socket_pid_linux.go
```diff
@@ -0,0 +1,127 @@
...
+		pid, err := strconv.Atoi(entry.Name())
+		if err != nil {
+			continue
+		}
+
+		cmdline, err := os.ReadFile(filepath.Join("/proc", entry.Name(), "cmdline"))
+		if err != nil || len(cmdline) == 0 {
+			continue
+		}
+		if strings.Contains(string(cmdline), socketPath) {
```
masnwilliams: nit: substring match means `/run/vm-1.sock` would also match a process with `/run/vm-10.sock` in its cmdline. low risk since this is a fallback path, but splitting on `\0` and checking exact argument match would be more precise.

lib/instances/guestmemory_darwin_test.go
```diff
@@ -175,3 +176,30 @@ func mustReadDarwinRSSBytes(t *testing.T, pid int) int64 {
 	require.NoError(t, err)
 	return kb * 1024
 }
+
+func assertActiveBallooningLifecycleVZ(t *testing.T, ctx context.Context, inst *Instance) {
```
masnwilliams: this is identical to `assertActiveBallooningLifecycle` in the linux test file (~60 lines). consider extracting to the shared `guestmemory_active_ballooning_test_helpers_test.go` to avoid drift.

lib/scopes/scopes.go
```diff
@@ -209,8 +209,9 @@ var RouteScopes = map[string]Scope{
 	"GET /devices/{id}":      DeviceRead,
 
 	// Health & Resources
-	"GET /health":    ResourceRead,
-	"GET /resources": ResourceRead,
+	"GET /health":                    ResourceRead,
+	"GET /resources":                 ResourceRead,
+	"POST /resources/memory/reclaim": ResourceRead,
```
masnwilliams: this maps a mutating POST to `ResourceRead` — every other POST in this file uses a write scope. intentional (since no `ResourceWrite` exists yet), or should a write scope be added?

masnwilliams (approved):
solid feature — well-structured controller with good observability and test infrastructure. left a few comments, mostly nits. the scope question (comment 8) is the main one worth a look.

lib/guestmemory/controller.go
```diff
@@ -0,0 +1,361 @@
...
+				appliedTarget = candidate.currentTargetGuestBytes
+			}
+		}
+		if appliedTarget != candidate.currentTargetGuestBytes {
+			if delta > 0 {
+				appliedTarget = candidate.currentTargetGuestBytes + minInt64(delta, c.config.PerVMMaxStepBytes)
+			} else {
+				appliedTarget = candidate.currentTargetGuestBytes - minInt64(-delta, c.config.PerVMMaxStepBytes)
+			}
+		}
```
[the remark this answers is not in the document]
sjmiller609: Recomputed `delta` from the post-suppression target before the step-size clamp so min-adjustment and cooldown suppression can't leave the clamp working from stale math anymore. Added a focused controller test that exercises the suppressed-adjustment path.

lib/guestmemory/active_ballooning.go
sjmiller609: Good call. The normalization path already clamps `low = high + 1` instead of silently resetting both values, so I left the behavior as the intent-preserving version you suggested and resolved the thread.

lib/guestmemory/controller.go
sjmiller609: Added a prune pass at the end of each reconcile so `lastApplied` entries are dropped for VMs that are no longer in the candidate set.

lib/guestmemory/controller.go
```diff
@@ -0,0 +1,363 @@
...
+
+func (c *controller) reconcile(ctx context.Context, req reconcileRequest) (ManualReclaimResponse, error) {
+	trigger := reconcileTrigger(req)
+	start := time.Now()
+	ctx, span := c.startReconcileSpan(ctx, req)
+	defer span.End()
+
+	state := &c.reconcileMu
+	<-state.mu
+	defer func() { state.mu <- struct{}{} }()
```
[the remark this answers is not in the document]
sjmiller609: Switched the reconcile lock acquisition to a `select` on `ctx.Done()` so a canceled manual reclaim request now returns promptly instead of waiting indefinitely behind an in-flight reconcile. Added a regression test for the canceled-while-waiting case.

lib/hypervisor/qemu/qemu.go
```diff
@@ -175,3 +186,30 @@ func (q *QEMU) ResizeMemory(ctx context.Context, bytes int64) error {
...
+		}
+	}
+
+	bytes, err := q.client.QueryBalloon()
+	if err != nil {
+		Remove(q.socketPath)
+		return 0, fmt.Errorf("query balloon target: %w", err)
+	}
+	return bytes, nil
+}
```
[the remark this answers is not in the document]
sjmiller609: QEMU now persists the requested balloon target and falls back to the saved VM config on a cold read, so `GetTargetGuestMemoryBytes` no longer uses `query-balloon`'s lagging actual-memory value as the target. I also added cache coverage for restart/socket-churn behavior and a QEMU cold-start test.

PR #179, opened 2026-03-31
Fix tc class collision with retry and class ID persistence

sjmiller609:
## Summary

Replaces `tc class replace` with `tc class add` + retry loop to properly detect and handle class ID hash collisions.

**Problem:** Two different TAP names can hash to the same 16-bit class ID via `deriveClassID`. Using `tc class replace` silently overwrites the existing class, breaking rate limiting for the original VM.

**Solution:**
- `addVMClass` now uses `tc class add` and retries with linear probing on "File exists" errors (up to 5 attempts)
- The actual class ID assigned is persisted to a `classid` file in the instance directory
- `removeVMClass` and `deleteTAPDevice` use the stored class ID (falling back to `deriveClassID` for backwards compatibility with old allocations)
- `CleanupOrphanedClasses` considers both derived and stored class IDs when determining which classes are valid
- New Prometheus counter `hypeman_network_tc_class_collisions_total` with `attempt` label (`initial` / `retry`) tracks collision frequency

## Changes
- `lib/network/bridge_linux.go` — retry loop in `addVMClass`, classID threading through `createTAPDevice`/`deleteTAPDevice`/`removeVMClass`/`CleanupOrphanedClasses`
- `lib/network/bridge_darwin.go` — updated stubs to match new signatures
- `lib/network/types.go` — `ClassID` field on `Allocation`
- `lib/network/allocate.go` — persist/load classID, pass stored ID on release
- `lib/network/derive.go` — load stored classID in `deriveAllocation`
- `lib/network/metrics.go` — `hypeman_network_tc_class_collisions_total` counter

## Backwards compatibility
Old allocations without a `classid` file continue to work — `removeVMClass` and `CleanupOrphanedClasses` fall back to `deriveClassID`.

---

> [!NOTE]
> **Medium Risk**
> Touches Linux traffic-control setup/teardown and cleanup logic; mistakes could leak or delete the wrong `tc` classes and impact VM rate limiting. Changes are localized but affect host networking behavior across restarts.
> 
> **Overview**
> Fixes a Linux upload-shaping edge case where two TAPs could hash to the same HTB class ID and silently clobber each other by switching to `tc class add` with a small collision-retry probe loop in `addVMClass`.
> 
> Threads the *actual* assigned class ID through TAP creation and persists it to an instance `classid` file, adds `Allocation.ClassID`, uses it during release/teardown, and updates orphaned class cleanup to treat stored (probed) IDs as valid. Adds a new metric counter `hypeman_network_tc_class_collisions_total` (labeled `initial`/`retry`) to track collision frequency, and updates macOS stubs for the new function signatures.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9719ac1b12802b85a6240d5196722d0b907cac35. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/network/bridge_linux.go
+++ b/lib/network/bridge_linux.go
@@ -1003,7 +1003,12 @@ func (m *manager) CleanupOrphanedClasses(ctx context.Context) int {
 		}
 	}
 	// Also include stored class IDs from allocations (handles probed IDs).
-	allocs, _ := m.ListAllocations(ctx)
+	// If ListAllocations fails, bail out to avoid deleting valid probed classes.
+	allocs, allocErr := m.ListAllocations(ctx)
+	if allocErr != nil {
+		log.WarnContext(ctx, "skipping orphaned class cleanup: failed to list allocations", "error", allocErr)
+		return 0
+	}
 	for _, alloc := range allocs {
 		if alloc.ClassID != "" {
 			validClassIDs[alloc.ClassID] = true
--- a/lib/network/bridge_linux.go
+++ b/lib/network/bridge_linux.go
@@ -730,10 +730,11 @@ func (m *manager) addVMClass(ctx context.Context, bridgeName, tapName string, ra
 				} else {
 					m.recordTCClassCollision(ctx, "retry")
 				}
-				// Increment class ID, wrapping within valid 16-bit range (0x0001-0xFFFF).
+				// Increment class ID, wrapping within valid 16-bit range.
+				// Skip 0 (invalid) and 1 (root class 1:1).
 				classIDVal++
-				if classIDVal > 0xFFFF {
-					classIDVal = 0x0001
+				if classIDVal == 0 || classIDVal == 1 {
+					classIDVal = 2
 				}
 				lastErr = fmt.Errorf("tc class add: %w (output: %s)", err, string(output))
 				continue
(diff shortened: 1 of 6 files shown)


lib/network/bridge_linux.go
```diff
@@ -705,14 +705,14 @@ func (m *manager) addVMClass(bridgeName, tapName string, rateBps, ceilBps int64)
 	}
 	ceilStr := formatTcRate(ceilBps)
 
-	// 1. Add HTB class for this VM
-	cmd := exec.Command("tc", "class", "add", "dev", bridgeName, "parent", htbRootClassID,
+	// 1. Add or update HTB class for this VM
+	cmd := exec.Command("tc", "class", "replace", "dev", bridgeName, "parent", htbRootClassID,
```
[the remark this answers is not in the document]
sjmiller609: will address

lib/network/bridge_linux.go
```diff
@@ -953,18 +989,26 @@ func (m *manager) CleanupOrphanedClasses(ctx context.Context) int {
 		return 0
 	}
 
-	// Build set of class IDs that belong to existing TAP devices
+	// Build set of class IDs that belong to existing TAP devices.
+	// Include both derived class IDs and stored class IDs from allocations
```
hiroTamada, line 993: nit: swallowing the error here means a transient `ListAllocations` failure would return an empty map, making every class look orphaned and get deleted. might be worth logging a warning and returning early on error.

lib/network/bridge_linux.go
```diff
@@ -689,65 +693,92 @@ func (m *manager) setupBridgeHTB(ctx context.Context, bridgeName string, capacit
...
+		// Try tc class add (NOT replace) so we detect collisions.
+		cmd := exec.Command("tc", "class", "add", "dev", bridgeName, "parent", htbRootClassID,
+			"classid", fullClassID, "htb", "rate", rateStr, "ceil", ceilStr, "prio", "1")
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			AmbientCaps: []uintptr{unix.CAP_NET_ADMIN},
+		}
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			// Check for "File exists" collision (exit status 2).
+			var exitErr *exec.ExitError
```
hiroTamada, line 726: nit: the `goto classAdded` is fine and clear, but this loop could also be a small extracted function that returns `(string, error)` — matter of taste.

hiroTamada (approved):
Nice fix — clean linear probing with good backwards compat and observability. Two minor nits inline.

PR #183, opened 2026-04-04
Auto-standby

sjmiller609:
## Summary
- add Linux-only auto-standby built around host conntrack state in a new `lib/autostandby` package
- persist and expose per-instance `auto_standby` policy through instance metadata and API surfaces
- start the auto-standby controller from the API process and add a default-skipped VM-level E2E test for host->guest inbound TCP activity

## Testing
- `go test -count=1 ./lib/autostandby`
- `go test -count=1 -run "Test(ValidateUpdateInstanceRequest|CloneStoredMetadataForFork_DeepCopiesReferenceFields)$" ./lib/instances`
- `go test -count=1 -run "Test(CreateInstance_MapsAutoStandbyPolicy|UpdateInstance_MapsAutoStandbyPatch)$" ./cmd/api/api`
- `go test -run "^$" ./cmd/api`
- `sudo -n env PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:$PATH HYPEMAN_RUN_AUTO_STANDBY_E2E=1 go test -count=1 -run ^TestAutoStandbyCloudHypervisorActiveInboundTCP$ ./lib/instances` on `deft-kernel-dev`

### Integration test coverage
The default-skipped Linux integration test exercised a real Cloud Hypervisor VM with networking enabled and a real conntrack-backed auto-standby controller.

It verified that:
- a host-to-guest TCP connection to nginx appears as qualifying inbound activity in conntrack
- the instance stays `Running` while that inbound TCP connection remains open
- once the final inbound TCP connection closes, the controller allows the configured idle timeout to elapse and then transitions the instance to `Standby`
- the test uses the real Linux conntrack path instead of ingress state or TAP byte counters

---

> [!NOTE]
> **Medium Risk**
> Introduces a new background controller that can transition VMs to `Standby` based on host conntrack state and persists new per-instance metadata, which could affect instance lifecycle behavior if misconfigured. Risk is mitigated by being opt-in via an `auto_standby` policy and having validation plus tests, but it touches core instance create/update paths and process startup.
> 
> **Overview**
> Adds an opt-in **auto-standby** feature that monitors host-side IPv4 TCP conntrack activity and automatically places eligible Linux VMs into `Standby` after a configured idle timeout.
> 
> Exposes the per-instance `auto_standby` policy through the instance create/update APIs (with validation and OAPI mapping), persists controller-owned runtime timestamps in `metadata.json`, and adds a new per-instance `GetAutoStandbyStatus` diagnostic endpoint.
> 
> Wires the new `autostandby.Controller` into the API process lifecycle (Wire provider + startup goroutine), and includes unit tests plus a default-skipped Linux E2E test that verifies host→guest TCP activity prevents standby until connections close.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 896ee1a7d9db4ff21fcd3903e0f4a7f063f52345. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/api/api/auto_standby.go
```diff
@@ -0,0 +1,62 @@
...
+		out.IdleTimeout = *policy.IdleTimeout
+	}
+	if policy.IgnoreSourceCidrs != nil {
+		out.IgnoreSourceCIDRs = append([]string(nil), (*policy.IgnoreSourceCidrs)...)
+	}
+	if policy.IgnoreDestinationPorts != nil {
+		out.IgnoreDestinationPorts = make([]uint16, 0, len(*policy.IgnoreDestinationPorts))
+		for _, port := range *policy.IgnoreDestinationPorts {
+			if port < 0 || port > 65535 {
+				return nil, fmt.Errorf("auto_standby.ignore_destination_ports must be between 1 and 65535")
```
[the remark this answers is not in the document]
sjmiller609: Fixed in d301651. The API-layer validation now rejects port 0 by checking for the inclusive range 1..65535, which matches the error message, OpenAPI schema, and downstream policy normalization. I also added a focused API test that exercises a zero-valued `auto_standby.ignore_destination_ports` update and verifies we return `invalid_auto_standby` instead of forwarding the request.

lib/instances/update.go
```diff
@@ -32,14 +31,34 @@ func (m *manager) updateInstance(ctx context.Context, id string, req UpdateInsta
...
+		}
+
+		log.InfoContext(ctx, "instance updated", "instance_id", id)
+
+		updated, err := m.getInstance(ctx, id)
+		if err != nil {
+			return nil, fmt.Errorf("get updated instance: %w", err)
+		}
+		return updated, nil
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in d301651. `updateInstance` no longer mutates the loaded metadata in place before the env-update path completes. It now clones metadata into an isolated copy when `auto_standby` is present, applies the policy on that copy, and passes the copy through the env/rules save path. That means the save persists env + auto-standby together, and any failure path only affects the temporary copy instead of leaking partial in-memory mutation. I added focused coverage for that combined save path as well.

sjmiller609 (commented):
## Review: Auto-standby PR #183

Well-structured PR overall. The `autostandby` package is cleanly separated with good abstractions (InstanceStore, ConnectionSource interfaces), the controller logic is correct, the classifier is well-tested, and the policy normalization is thorough. The integration test is comprehensive and exercises the real conntrack path.

Here are the findings worth discussing:

### update.go — Pointer aliasing in the auto-standby + env combined path

Lines 46-50:
```go
nextMeta := meta
if req.AutoStandby != nil {
    nextMeta = cloneMetadata(meta)
    nextMeta.AutoStandby = cloneAutoStandbyPolicy(req.AutoStandby)
}
```

When `req.AutoStandby == nil`, `nextMeta` is the **same pointer** as `meta`. The env path then mutates `nextMeta.Env` via `applyUpdatedInstanceEnv` → `meta.Env = nextEnv`. This works, but only because `nextMeta == meta` in that branch — it's fragile and non-obvious. If someone later adds logic between here and the save that reads from `meta` expecting the original state, they'll get the mutated version silently.

Consider always cloning when you intend to mutate, or at minimum adding a comment explaining the aliasing is intentional.

When both AutoStandby and Env are set simultaneously, the clone happens correctly. But the asymmetry between the two code paths makes the code harder to reason about.

### controller.go — ListInstances fetches all instances every poll cycle

`Poll()` calls `c.store.ListInstances(ctx)` which maps to `s.manager.ListInstances(ctx, nil)` — this lists **all** instances regardless of state or auto-standby config. With a 5-second poll interval, this is fine at small scale but could be worth optimizing later (e.g., filtering to running instances with auto-standby enabled at the store level, or at least only querying running instances).

Not a blocker, but worth a TODO or tracking ticket if the instance count per host is expected to grow.

### conntrack_linux.go — IPv4 only

`ListConnections` queries `unix.AF_INET` only. The README mentions "Linux only" and "TCP only" but doesn't mention IPv4-only. Worth noting in the README for completeness.

### classifier.go — Policy re-compilation on every poll

`ActiveInboundCount` → `compilePolicy` → `NormalizePolicy` on every call. This re-parses the idle timeout duration and all CIDRs from string form each time. Not a problem at current scale (once per eligible instance per 5s poll), but an easy optimization target if it ever matters — the compiled form could be cached by instance ID and invalidated on policy change.

### types.go — TCPState Active() includes CloseWait and LastAck

`Active()` returns true for `TCPStateCloseWait` and `TCPStateLastAck`. These are connection teardown states — including them means the idle timer won't start until the TCP connection is fully closed on both sides, which is the safe choice. Just confirming this is intentional (it seems correct for avoiding premature standby).

### update.go — Auto-standby updates allowed in any instance state

The state check on line 43 only applies to env updates:
```go
if len(req.Env) > 0 && inst.State != StateRunning && inst.State != StateInitializing {
```

This means auto-standby policy can be set on stopped/standby instances. This looks intentional (pre-configure before start), just confirming.

### metadata_clone.go — cloneStoredMetadataForFork naming

`cloneMetadata` delegates to `cloneStoredMetadataForFork`. The "ForFork" name is slightly misleading when used in the update context. The behavior is correct (it deep-copies all reference fields without resetting anything), but the name implies fork-specific semantics. Consider renaming to `deepCopyStoredMetadata` or similar, or at least a comment noting it's also used by the update path.

### compression_integration_test_helpers_test.go — Unrelated test guard

This file adds a skip guard for standby/restore compression tests (`HYPEMAN_RUN_STANDBY_RESTORE_COMPRESSION_TESTS`). Looks like a useful standalone improvement that could be a separate commit for clarity.

### Summary

The core auto-standby logic is solid. The main area I'd look at more carefully is the update.go pointer aliasing pattern — it works but adds cognitive overhead. Everything else is clean, well-tested, and follows existing codebase patterns.

cmd/api/main.go
```diff
@@ -495,6 +495,9 @@ func run() error {
 		logger.Info("starting guest memory controller")
 		return app.GuestMemoryController.Start(gctx)
 	})
+	if startAutoStandbyController(grp, gctx, logger, app.InstanceManager) {
```
hiroTamada: nit: this is hand-wired in `run()` while stuff like `GuestMemoryController` comes out of wire — not wrong since `NewController` already takes interfaces, but a `Provide*` (or similar) on the wire graph would match the rest of the composition root if we want consistency.

sjmiller609: Fixed in 5c1df95. I moved the auto-standby controller construction into the Wire graph via `providers.ProvideAutoStandbyController` (Linux provider + unsupported stub), added it to the `application` struct, regenerated `wire_gen.go`, and removed the hand-wired `startAutoStandbyController` path from `run()`. That brings it in line with the other background controllers at the composition root. Verified with `go test -count=1 ./lib/providers` and `go test -count=1 -run ^ ./cmd/api`.

lib/instances/metadata_clone.go
```diff
@@ -0,0 +1,13 @@
...
+
+// deepCopyMetadata returns a metadata copy that can be mutated without
+// affecting the originally loaded instance metadata.
+func deepCopyMetadata(src *metadata) *metadata {
+	if src == nil {
+		return nil
+	}
+
+	return &metadata{
+		StoredMetadata: cloneStoredMetadataForFork(src.StoredMetadata),
```
hiroTamada: nit: `deepCopyMetadata` calls `cloneStoredMetadataForFork` — works, but the name reads fork-specific; something neutral like `cloneStoredMetadata` (and reuse everywhere) might avoid implying this path is fork-only.

lib/autostandby/controller.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Controller decides when eligible instances should transition to standby.
+type Controller struct {
+	store        InstanceStore
+	source       ConnectionSource
+	log          *slog.Logger
+	now          func() time.Time
+	pollInterval time.Duration
+	idleSince    map[string]time.Time
```
hiroTamada: suggestion: `idleSince` only lives in process memory today, so an API restart resets the idle clock (README already calls this out). if we want standby timing to survive restarts or match “how long has this VM actually been idle” more closely, we could persist something like first-zero-activity timestamp (or last-activity) alongside instance metadata and reconcile from disk each poll.

hiroTamada: mainly for the billing concern and confusion. 

It is a tiny edge case, but I imagine it is worth noting.

lib/instances/metadata_clone.go
sjmiller609: Addressed earlier in 4043a13. The helper was renamed from the fork-specific `cloneStoredMetadataForFork` to the neutral `cloneStoredMetadata`, and the update/fork/snapshot paths now all reuse that shared name.

lib/autostandby/controller.go
sjmiller609: This is handled on the current branch now. Auto-standby runtime timestamps are persisted in instance metadata, startup rebuilds state from disk plus a conntrack snapshot, pre-existing active connections are reconciled until they drain, and there is now a 5-minute periodic snapshot sync as a consistency sweep. So the original “idleSince only lives in process memory” concern is outdated relative to `789a08c` and `16e6c20`.

lib/autostandby/controller.go
```diff
@@ -0,0 +1,953 @@
...
+			return snapshot
+		}
+		snapshot.Status = StatusReadyForStandby
+		snapshot.Reason = ReasonReadyForStandby
+		return snapshot
+	}
+
+	snapshot.Status = StatusIdleCountdown
+	snapshot.Reason = ReasonIdleTimeoutNotElapsed
+	return snapshot
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 767b8a2. `Describe()` now snapshots every mutable controller field it needs while still holding `c.mu.RLock()` and only reads those copied values after unlock, so it no longer reads `activeInbound` or other mutable state concurrently with writers. I also re-ran `go test -race -count=1 ./lib/autostandby` after the change.

lib/autostandby/conntrack_events_linux.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+}
+
+func connectionEventFromNetlinkMessage(msg syscall.NetlinkMessage) (ConnectionEvent, bool, error) {
+	switch msg.Header.Type {
+	case unix.NLMSG_NOOP, unix.NLMSG_DONE:
+		return ConnectionEvent{}, false, nil
+	case unix.NLMSG_ERROR:
+		if len(msg.Data) >= 4 {
+			errno := -int32(binary.LittleEndian.Uint32(msg.Data[:4]))
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 767b8a2. `NLMSG_ERROR` now decodes the errno using `nl.NativeEndian()` instead of hardcoded little-endian, and I added a focused parser test to lock in the native-endian path.

lib/instances/lifecycle_events.go
```diff
@@ -0,0 +1,73 @@
+package instances
+
+import "sync"
+
+// LifecycleEventAction identifies which instance lifecycle action occurred.
+type LifecycleEventAction string
```
sjmiller609, line 25: I thought we already have a lifecycle subscription system, checking

lib/autostandby/controller.go
```diff
@@ -0,0 +1,1005 @@
...
+	state.instance = cloneInstance(inst)
+	state.standbyRequested = false
+
+	if !eligible(inst) {
+		c.clearStateLocked(state)
+		if inst.Runtime != nil || state.idleSince != nil || state.lastInboundAt != nil {
+			return c.persistRuntime(ctx, inst.ID, nil)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in ed02ddb. The ineligible-instance path now captures whether runtime existed before `clearStateLocked`, then persists `nil` based on that pre-clear signal. That ensures lifecycle events which make an instance ineligible immediately clear stale auto-standby runtime from disk instead of waiting for the periodic snapshot sync. I also added a regression test covering an update event that makes an instance ineligible and verifies the persisted runtime is cleared.

lib/providers/auto_standby_linux.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return &autostandby.Instance{
+		ID:             inst.Id,
+		Name:           inst.Name,
+		State:          string(inst.State),
+		NetworkEnabled: inst.NetworkEnabled,
+		IP:             inst.IP,
+		HasVGPU:        inst.GPUProfile != "" || inst.GPUMdevUUID != "",
+		AutoStandby:    inst.AutoStandby,
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in 6f3c229. `SubscribeInstanceEvents()` now hydrates `Runtime` from `GetAutoStandbyRuntime(...)` before forwarding non-delete lifecycle events into the auto-standby controller, so an already-idle instance keeps its persisted countdown across update/start/restore/fork lifecycle notifications instead of resetting to "now". I also added a focused Linux provider test that verifies the forwarded lifecycle event includes the persisted runtime timestamps.

lib/instances/fork.go
```diff
@@ -482,6 +482,9 @@ func cloneStoredMetadataForFork(src StoredMetadata) StoredMetadata {
 	if src.NetworkEgress != nil {
 		dst.NetworkEgress = cloneNetworkEgressPolicy(src.NetworkEgress)
 	}
+	if src.AutoStandby != nil {
+		dst.AutoStandby = cloneAutoStandbyPolicy(src.AutoStandby)
+	}
```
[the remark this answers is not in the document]
sjmiller609: Fixed in `76caba7`. `cloneStoredMetadata` now deep-clones `SnapshotPolicy` via the existing `cloneSnapshotPolicy(...)` helper, so the shared metadata copy path no longer aliases the nested compression config pointer. I also extended the existing metadata-clone regression test to mutate the cloned snapshot compression level and verify the source metadata remains unchanged.

lib/autostandby/controller.go
```diff
@@ -0,0 +1,1009 @@
...
+
+	if event.ObservedAt.IsZero() {
+		event.ObservedAt = c.now().UTC()
+	}
+	c.recordConntrackEvent(string(event.Type), "received")
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for id, state := range c.states {
```
hiroTamada, line 552: nit: every conntrack event iterates all tracked VMs (`for id, state := range c.states`). fine for current scale, but if a host ever runs hundreds of auto-standby VMs this becomes a hot path. a dest-IP → instance-ID index would make it O(1).

hiroTamada (approved):
well-designed feature. the single-goroutine event loop with conntrack integration is solid, state machine is correct, restart recovery via persisted runtime is thoughtful, and the periodic snapshot sync provides a good safety net. code is clean, well-tested, and the README is excellent.

one minor nit inline about O(N) iteration per conntrack event — not blocking, just something to revisit if host density grows significantly.

hiroTamada:
a few non-blocking observations from the review:

1. **lifecycle event gap for auto-standby-only updates** — the early-return path in `updateInstance` (when `len(req.Env) == 0`) saves metadata and returns without emitting a lifecycle event. the controller won't learn about the policy change until the next periodic snapshot sync (5 min). if this hasn't been wired up elsewhere in the lifecycle events system, might be worth a quick follow-up to emit an `InstanceEventUpdate` on that path.

2. **no test for `handleStandbyTimer` calling `StandbyInstance`** — the fakes are already set up to track it (`fakeInstanceStore.standbyIDs`), just no test exercises the timer-fires → standby-called → state-cleared path, or the retry-on-failure path. would be a good addition given it's the most critical action the controller takes.

3. **PR scope** — this bundles several distinct features (auto-standby, distributed tracing refactor, admission control, image retention, `WaitForState`). each piece is individually clean, but the combined diff (~13k lines, 151 files) makes it a large review surface. not a blocker, just noting for future PRs.

sjmiller609:
> a few non-blocking observations from the review:
> 
> 1. **lifecycle event gap for auto-standby-only updates** — the early-return path in `updateInstance` (when `len(req.Env) == 0`) saves metadata and returns without emitting a lifecycle event. the controller won't learn about the policy change until the next periodic snapshot sync (5 min). if this hasn't been wired up elsewhere in the lifecycle events system, might be worth a quick follow-up to emit an `InstanceEventUpdate` on that path.
> 2. **no test for `handleStandbyTimer` calling `StandbyInstance`** — the fakes are already set up to track it (`fakeInstanceStore.standbyIDs`), just no test exercises the timer-fires → standby-called → state-cleared path, or the retry-on-failure path. would be a good addition given it's the most critical action the controller takes.
> 3. **PR scope** — this bundles several distinct features (auto-standby, distributed tracing refactor, admission control, image retention, `WaitForState`). each piece is individually clean, but the combined diff (~13k lines, 151 files) makes it a large review surface. not a blocker, just noting for future PRs.

Lifecycle event gap
I checked this path again, and this one is already covered by the existing lifecycle wrapper. updateInstance(...) itself returns after saving metadata for the auto-standby-only case, but manager.UpdateInstance() still emits LifecycleEventUpdate after m.updateInstance(...) succeeds. So the controller should learn about the policy change immediately through the manager-level notification path rather than waiting for the 5-minute periodic sync. I don’t think we need an extra emit inside updateInstance itself.

No test for handleStandbyTimer
This is a fair point. We do have good coverage around state transitions, event handling, and the end-to-end Linux flow, but we do not currently have a focused unit test for the exact timer-fired -> StandbyInstance call -> state cleanup path, or the failure/retry behavior. I agree that would be a useful addition because it exercises the controller’s most important side effect directly.

PR scope
I think this comment may be based on a stale view of the branch. This PR is the auto-standby work plus the directly related wiring/docs/tests that were needed to land it safely. It shouldn’t include unrelated feature work like admission control, image retention, or WaitForState. I agree with the general principle about keeping review surfaces smaller, but I don’t think that list matches the current contents of this PR.

sjmiller609:
Follow-up on Hiro's top-level notes:

- I added focused controller tests for the standby timer path in `f9800fe`, covering both the successful `handleStandbyTimer -> StandbyInstance -> state cleared` flow and the failure/rearm path.
- I also added a manager-level regression test proving an auto-standby-only `UpdateInstance` still publishes `LifecycleEventUpdate` through the public manager wrapper, so that path does not wait for the 5-minute snapshot sync.
- On PR scope: agreed in principle about keeping review surfaces smaller, but I don't think the current branch includes unrelated features like admission control, image retention, or `WaitForState`; this PR is the auto-standby work plus closely related wiring/docs/tests.

PR #186, opened 2026-04-05
Add regex-based instance limit overrides

sjmiller609:
## Summary
- add ordered regex-based instance limit overrides in config
- resolve effective per-instance limits from matching name patterns
- apply the resolved limits on create and fork flows

## Testing
- go test ./cmd/api/config ./lib/providers
- go test ./lib/instances -run ^ (with temporary placeholder embedded binaries in this checkout)


sjmiller609 (commented):
## Code Review: Regex-based instance limit overrides

Overall this is well-structured — clean separation between config parsing, per-instance limits, and aggregate provisioned limits. The first-match-wins semantics are clearly documented and tested. A few items to address:

### 1. Duplicate `parseBandwidth` function

`cmd/api/config/limits_patterns.go` defines a local `parseBandwidth` that is a near-exact copy of `resources.ParseBandwidth` in `lib/resources/util.go`. Having two copies means they can diverge — validation could accept a bandwidth string that parsing later rejects (or vice versa). The config package doesn't import `lib/resources`, so there shouldn't be a cycle. Recommend importing and reusing `resources.ParseBandwidth` instead of duplicating it.

### 2. Semantic change: `MaxOverlaySize` now has `0 = unlimited`

Before this PR, the overlay size check was unconditional:
```go
if overlaySize > m.limits.MaxOverlaySize {
```
Now it's guarded:
```go
if effective.MaxOverlaySize > 0 && overlaySize > effective.MaxOverlaySize {
```
The comment in `manager.go` was updated to say "0 = unlimited", so this seems intentional. But it's a behavioral change — if anyone had `max_overlay_size: 0` (admittedly unlikely given the default is "100GB"), it would previously reject all instances and now it means unlimited. Worth a quick confirmation this is the desired behavior, or add a note in the PR description.

### 3. TOCTOU on aggregate limit checks

`validateNamedResourceLimits` calls `m.ListInstanceAllocations(ctx)` and then checks projected totals against the pattern limit. Two concurrent create requests could both pass validation before either is committed, exceeding the aggregate cap. This is probably acceptable for soft resource limits (similar to the existing `ResourceValidator`), but worth documenting as a known limitation if aggregate limits are expected to be strictly enforced.

### 4. Config validation mutates input during `Validate()`

`validateNamePatternLimits` trims whitespace and writes back into the config struct (e.g., `*cfg.MaxMemoryPerInstance = value`). This means `Validate()` has side effects — the config is different after validation than before. This works fine at startup, but could be surprising if `Validate()` is ever called in other contexts (tests, dry-run, etc.). Consider doing the normalization in `Load()` or a separate `Normalize()` step instead.

### 5. Missing test: global overlay limit still enforced when no pattern matches

The tests cover pattern matching and override scenarios well. One gap: there's no test verifying that the global `MaxOverlaySize` is still enforced for instances whose names don't match any pattern. The `ForName` fallback logic handles this correctly, but a regression test would be valuable since the overlay check was moved from inline code to the new `validateResourceLimitsForName` path.

### What looks good

- First-match-wins semantics are clearly documented and tested
- `ForName()` cleanly resolves effective limits with fallback to globals
- Aggregate provisioned limits correctly group by pattern index, preventing cross-pattern accounting
- Tests cover the key behaviors: first-match-wins, field-level fallback, zero-means-unlimited, aggregate projection
- Validation is consistently applied across create, fork, and snapshot-fork paths

sjmiller609:
not worth it for now

PR #189, opened 2026-04-06
Consolidate instance lifecycle subscriptions

sjmiller609:
## Summary
- replace the old per-instance state subscription path with one shared lifecycle event bus in `lib/instances`
- migrate `WaitForState` to filter the lifecycle bus and keep its polling fallback
- add lifecycle bus backpressure metrics for subscriber count, max queue depth, and dropped events

## Why This Change
We were about to introduce a second internal subscription mechanism for the upcoming auto-standby work. The existing `WaitForState` path used a per-instance state subscription that only carried `State`/`StateError`, while the new controller needs a global stream of lifecycle changes like create, update, fork, and delete.

Instead of keeping two different in-memory buses, this change consolidates on one richer lifecycle event stream and makes `WaitForState` a consumer of that shared path. That keeps the public API the same, reduces duplicated fanout logic, and gives us one place to observe subscriber backpressure.

## Testing
- `go test ./lib/instances -run 'TestWaitForState|TestLifecycleSubscribers|TestLifecycleEventMetrics'`
- `go test ./lib/builds -run 'Test'`
- `go test ./lib/instances -run '^$'`
- `go test ./lib/builds -run '^$'`
- `go test ./cmd/api/api -run '^$'`

---

> [!NOTE]
> **Medium Risk**
> Changes core instance lifecycle signaling by replacing per-instance state subscriptions with a shared lifecycle event bus, which could affect wait/polling behavior and internal controllers if events are missed or dropped. Adds new buffering/config/metrics paths that need validation in production workloads.
> 
> **Overview**
> **Consolidates internal instance change notifications** by replacing the per-instance `StateChange` subscription mechanism with a shared `LifecycleEvent` bus, and updates the `Manager` interface to expose `SubscribeLifecycleEvents`.
> 
> `WaitForState` now consumes this shared stream (filtering by `instance_id` and handling `delete` explicitly) while retaining the 5s polling fallback, and manager mutation methods (`CreateInstance`, `UpdateInstance`, `Start/Stop`, `Standby/Restore`, `Fork*`, `DeleteInstance`) publish lifecycle events on success.
> 
> Adds observability and tuning for backpressure: new metrics for subscriber count, max queue depth, and dropped events, plus a new `instances.lifecycle_event_buffer_size` config setting (default `256`) wired through providers into `instances.ManagerConfig` and validated in config/tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 134628d677cbb38284330c93bb7f9dc5c741c368. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/instances/wait.go
```diff
@@ -80,14 +80,20 @@ func WaitForState(ctx context.Context, mgr Manager, inst *Instance, targetState
...
-		case sc, ok := <-ch:
+		case event, ok := <-ch:
 			if !ok {
-				// Channel closed — instance was deleted.
 				return nil, ErrNotFound
 			}
-			latest = &Instance{}
-			latest.State = sc.State
-			latest.StateError = sc.StateError
+			if event.InstanceID != id {
```
hiroTamada, line 87: nit: with the shared bus, every WaitForState subscriber receives and discards events for all other instances. at current scale this is fine, but if concurrent WaitForState calls grow into the hundreds, the fan-out cost could add up. might be worth considering instance-level filtering at the subscription layer down the road (or a separate per-instance fast path alongside the global bus for auto_standby).

lib/instances/manager.go
```diff
@@ -57,9 +57,8 @@ type Manager interface {
 	SetResourceValidator(v ResourceValidator)
 	// GetVsockDialer returns a VsockDialer for the specified instance.
 	GetVsockDialer(ctx context.Context, instanceID string) (hypervisor.VsockDialer, error)
-	// Subscribe returns a channel that receives state change events for the
-	// given instance, plus an unsubscribe function the caller must defer.
-	Subscribe(instanceID string) (<-chan StateChange, func())
+	// SubscribeLifecycleEvents returns the shared internal lifecycle event stream.
+	SubscribeLifecycleEvents(consumer LifecycleEventConsumer) (<-chan LifecycleEvent, func())
```
hiroTamada, line 61: nit: this could have been an additive change — keep the old `Subscribe(instanceID)` on the interface, add `SubscribeLifecycleEvents` alongside it, and back both with the same bus internally. that way WaitForState doesn't need to change at all in this PR, and the interface break + mock updates happen as a separate cleanup. not blocking since Manager is internal, but it would've made this diff smaller and lower risk.

hiroTamada (approved):
solid refactor — clean fan-out bus with good observability. two minor comments.

lib/instances/wait.go
sjmiller609: I don't think that would be a problem even if there were 10k instances

PR #200, opened 2026-04-24
Add cloud-hypervisor v51.1 with backwards-compatible version flag (CVE-2026-27211)

ulziibay-kernel:
## Summary

Adds Cloud Hypervisor v51.1 alongside existing v49.0 with a config flag to control which version new instances use. This enables a safe, no-downtime upgrade path for CVE-2026-27211.

### What changed

- **Both v49.0 and v51.1 binaries are embedded** -- no need to drain standby instances before deploy
- **New config flag**: `hypervisor.cloud_hypervisor_version` (default: `v49.0`)
  - Env override: `HYPERVISOR__CLOUD_HYPERVISOR_VERSION=v51.1`
- Existing standby instances restore using their stored version (already the case)
- New instances use whichever version the flag is set to
- VMM client regenerated from v51.1 OpenAPI spec with `ImageType: Raw` on all disks (CVE fix)

### Upgrade path

1. Deploy this PR -- all instances continue on v49.0, no disruption
2. When ready, set `HYPERVISOR__CLOUD_HYPERVISOR_VERSION=v51.1`
3. New instances get v51.1; existing standbys still restore on v49.0
4. Once all v49.0 standbys have expired, v49.0 can be removed in a follow-up

### Security fix (CVE-2026-27211)

Fixes GHSA-jmr4-g2hv-mjj6: arbitrary host file exfiltration via crafted QCOW2 disk headers. Affects CH versions 34.0 through 50.0. Fixed in 50.1+.

v51.1 additionally includes:
- QCOW2 v3 improvements (live resize, variable refcount, dirty bit)
- DISCARD/WRITE_ZEROES support for virtio-blk
- THP for anonymous shared memory (performance)
- ACPI NUMA affinity for VFIO-PCI (GPU passthrough)

### Files changed

- `Makefile` -- download both v49.0 and v51.1 binaries
- `cmd/api/config/config.go` -- add `cloud_hypervisor_version` field (default `v49.0`)
- `lib/vmm/binaries_linux.go` -- embed both v49.0 and v51.1
- `lib/vmm/binaries_darwin.go` -- version constants for compile compat
- `lib/vmm/version.go` -- ParseVersion matches both versions
- `lib/vmm/vmm.go` -- regenerated from v51.1 OpenAPI spec
- `lib/vmm/client_test.go` -- tests for both versions
- `lib/hypervisor/cloudhypervisor/process.go` -- SetDefaultVersion/GetDefaultVersion, configurable GetVersion
- `lib/hypervisor/cloudhypervisor/config.go` -- ImageType: Raw on all disks (CVE fix, safe for v49.0)
- `lib/providers/providers.go` -- wire config flag at startup
- `lib/instances/version_upgrade_test.go` -- **new E2E test**: create v49.0 -> standby -> switch to v51.1 -> restore -> verify version preserved -> create new instance on v51.1

### Test plan

- `go build ./lib/vmm/... ./lib/hypervisor/cloudhypervisor/...`
- `go vet` clean
- `go test ./lib/vmm/...` -- both versions extract and start
- `go test ./lib/hypervisor/cloudhypervisor/...` passes
- `TestCloudHypervisorVersionUpgradeRestore` -- standby/restore across version change
- Full `go test ./lib/instances/...` (requires KVM, CI)

---

> [!NOTE]
> **Medium Risk**
> Changes which Cloud Hypervisor binary can be selected for new VM instances and updates the generated VMM API client; mistakes could break VM start/restore or lead to mismatched snapshot restores across versions.
> 
> **Overview**
> Adds Cloud Hypervisor `v51.1` as an additional embedded VMM binary (keeping `v49.0`) and updates build tooling to download/ensure both versions, enabling backwards-compatible upgrades.
> 
> Introduces a new config field `hypervisor.cloud_hypervisor_default_version` and runtime override (`cloudhypervisor.SetDefaultVersion`/`GetDefaultVersion`) so **new instances** can use a chosen CH version, while **existing instances** continue to restore using the version stored in metadata; `CreateInstanceRequest` now supports an explicit `HypervisorVersion` with validation.
> 
> Regenerates the Cloud Hypervisor OpenAPI client/spec from `v51.1` (including `/vm.resize-disk` and new schema fields) and sets CH disk `ImageType` to `Raw` in VM config generation; adds version-aware capabilities (e.g., disk resize on `v51.1`) and an integration test covering standby/restore across a default-version flip.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32e68205b40be49254ee6f4a1a7dfb0c71c7d2c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,10 @@ build-caddy: $(XCADDY)
 	@echo "Caddy binary built successfully"
 
 # Download Cloud Hypervisor API spec
+# All embedded CH versions currently share the same API spec (v0.3.0).
+# Multiple API versions will only be needed once there are breaking changes
+# introduced. Use version-specific Capabilities when new features are
+# introduced that aren't supported on previous versions.
 download-ch-spec:
 	@echo "Downloading Cloud Hypervisor API spec..."
 	@mkdir -p specs/cloud-hypervisor/api-v0.3.0
--- a/Makefile
+++ b/Makefile
@@ -37,18 +37,12 @@ install-tools: $(OAPI_CODEGEN) $(AIR) $(WIRE) $(XCADDY)
 # Download Cloud Hypervisor binaries
 download-ch-binaries:
 	@echo "Downloading Cloud Hypervisor binaries..."
-	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v48.0/{x86_64,aarch64}
-	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v49.0/{x86_64,aarch64}
-	@echo "Downloading v48.0..."
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v48.0/x86_64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v48.0/cloud-hypervisor-static
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v48.0/aarch64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v48.0/cloud-hypervisor-static-aarch64
-	@echo "Downloading v49.0..."
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v49.0/x86_64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v49.0/cloud-hypervisor-static
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v49.0/aarch64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v49.0/cloud-hypervisor-static-aarch64
+	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v50.1/{x86_64,aarch64}
+	@echo "Downloading v50.1..."
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v50.1/x86_64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v50.1/cloud-hypervisor-static
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v50.1/aarch64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v50.1/cloud-hypervisor-static-aarch64
 	@chmod +x lib/vmm/binaries/cloud-hypervisor/v*/*/cloud-hypervisor
 	@echo "Binaries downloaded successfully"
 
@@ -116,7 +110,7 @@ download-ch-spec:
 	@echo "Downloading Cloud Hypervisor API spec..."
 	@mkdir -p specs/cloud-hypervisor/api-v0.3.0
 	@curl -L -o specs/cloud-hypervisor/api-v0.3.0/cloud-hypervisor.yaml \
-		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v48.0/vmm/src/api/openapi/cloud-hypervisor.yaml
+		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v50.1/vmm/src/api/openapi/cloud-hypervisor.yaml
 	@echo "API spec downloaded"
 
 # Generate Go code from OpenAPI spec
@@ -168,7 +162,7 @@ ensure-ch-binaries:
 	else \
 		echo "Unsupported architecture: $$ARCH"; exit 1; \
 	fi; \
-	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v48.0/$$CH_ARCH/cloud-hypervisor ]; then \
+	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v50.1/$$CH_ARCH/cloud-hypervisor ]; then \
 		echo "Cloud Hypervisor binaries not found, downloading..."; \
 		$(MAKE) download-ch-binaries; \
 	fi
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,16 @@ $(XCADDY): | $(BIN_DIR)
 
 install-tools: $(OAPI_CODEGEN) $(AIR) $(WIRE) $(XCADDY)
 
-# Download Cloud Hypervisor binaries
+# Download Cloud Hypervisor binaries (both v49.0 and v51.1 for backwards-compatible upgrades)
 download-ch-binaries:
 	@echo "Downloading Cloud Hypervisor binaries..."
+	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v49.0/{x86_64,aarch64}
 	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v51.1/{x86_64,aarch64}
+	@echo "Downloading v49.0..."
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v49.0/x86_64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v49.0/cloud-hypervisor-static
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v49.0/aarch64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v49.0/cloud-hypervisor-static-aarch64
 	@echo "Downloading v51.1..."
 	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v51.1/x86_64/cloud-hypervisor \
 		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v51.1/cloud-hypervisor-static
@@ -162,7 +168,8 @@ ensure-ch-binaries:
 	else \
 		echo "Unsupported architecture: $$ARCH"; exit 1; \
 	fi; \
-	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v51.1/$$CH_ARCH/cloud-hypervisor ]; then \
+	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v49.0/$$CH_ARCH/cloud-hypervisor ] || \
+	   [ ! -f lib/vmm/binaries/cloud-hypervisor/v51.1/$$CH_ARCH/cloud-hypervisor ]; then \
 		echo "Cloud Hypervisor binaries not found, downloading..."; \
 		$(MAKE) download-ch-binaries; \
 	fi
--- a/Makefile
+++ b/Makefile
@@ -37,12 +37,12 @@ install-tools: $(OAPI_CODEGEN) $(AIR) $(WIRE) $(XCADDY)
 # Download Cloud Hypervisor binaries
 download-ch-binaries:
 	@echo "Downloading Cloud Hypervisor binaries..."
-	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v50.1/{x86_64,aarch64}
-	@echo "Downloading v50.1..."
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v50.1/x86_64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v50.1/cloud-hypervisor-static
-	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v50.1/aarch64/cloud-hypervisor \
-		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v50.1/cloud-hypervisor-static-aarch64
+	@mkdir -p lib/vmm/binaries/cloud-hypervisor/v51.1/{x86_64,aarch64}
+	@echo "Downloading v51.1..."
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v51.1/x86_64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v51.1/cloud-hypervisor-static
+	@curl -L -o lib/vmm/binaries/cloud-hypervisor/v51.1/aarch64/cloud-hypervisor \
+		https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v51.1/cloud-hypervisor-static-aarch64
 	@chmod +x lib/vmm/binaries/cloud-hypervisor/v*/*/cloud-hypervisor
 	@echo "Binaries downloaded successfully"
 
@@ -110,7 +110,7 @@ download-ch-spec:
 	@echo "Downloading Cloud Hypervisor API spec..."
 	@mkdir -p specs/cloud-hypervisor/api-v0.3.0
 	@curl -L -o specs/cloud-hypervisor/api-v0.3.0/cloud-hypervisor.yaml \
-		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v50.1/vmm/src/api/openapi/cloud-hypervisor.yaml
+		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v51.1/vmm/src/api/openapi/cloud-hypervisor.yaml
 	@echo "API spec downloaded"
 
 # Generate Go code from OpenAPI spec
@@ -162,7 +162,7 @@ ensure-ch-binaries:
 	else \
 		echo "Unsupported architecture: $$ARCH"; exit 1; \
 	fi; \
-	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v50.1/$$CH_ARCH/cloud-hypervisor ]; then \
+	if [ ! -f lib/vmm/binaries/cloud-hypervisor/v51.1/$$CH_ARCH/cloud-hypervisor ]; then \
 		echo "Cloud Hypervisor binaries not found, downloading..."; \
 		$(MAKE) download-ch-binaries; \
 	fi
--- a/cmd/api/config/config.go
+++ b/cmd/api/config/config.go
@@ -196,7 +196,7 @@ type HypervisorConfig struct {
 	Default                       string                 `koanf:"default"`
 	CloudHypervisorDefaultVersion string                 `koanf:"cloud_hypervisor_default_version"`
 	FirecrackerBinaryPath         string                 `koanf:"firecracker_binary_path"`
-	Memory                 HypervisorMemoryConfig `koanf:"memory"`
+	Memory                        HypervisorMemoryConfig `koanf:"memory"`
 }
 
 // HypervisorMemoryConfig holds guest memory management settings.
--- a/cmd/api/config/config.go
+++ b/cmd/api/config/config.go
@@ -193,9 +193,9 @@ type CapacityConfig struct {
 
 // HypervisorConfig holds hypervisor settings.
 type HypervisorConfig struct {
-	Default                string                 `koanf:"default"`
-	CloudHypervisorVersion string                 `koanf:"cloud_hypervisor_version"`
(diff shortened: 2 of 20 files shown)


ulziibay-kernel:
## Changes between v49.0 and v51.1 -- reviewer notes

> **Update (May 12)**: This PR is now **backwards-compatible**. Both v49.0 and v51.1 are embedded. A config flag (`hypervisor.cloud_hypervisor_version`, default `v49.0`) controls which version new instances use. Existing standby instances restore using their stored version. No drain strategy needed.

### Breaking changes addressed

| Change | Version | Impact on us | Action taken |
|---|---|---|---|
| **Sector-zero write prevention on autodetected raw images** | v50.1+ | **BREAKING** -- overlay disk (vdb) fails with I/O errors | **Fixed**: set `ImageType: Raw` explicitly on all disk configs |
| **`backing_files` defaults to `off`** | v50.1+ | Low risk -- we use raw images, not QCOW2 | No action needed |
| **Snapshot restore requires exact CH version match** | v50.0+ | **HIGH** -- standbys on v49.0 can't restore with v51.1 binary | **Fixed**: both binaries embedded, restore uses stored version |

### Upgrade path

1. Deploy this PR -- all instances continue on v49.0, no disruption
2. Set `HYPERVISOR__CLOUD_HYPERVISOR_VERSION=v51.1` when ready
3. New instances get v51.1; existing standbys still restore on v49.0
4. Once all v49.0 standbys expire, v49.0 can be removed in a follow-up

### Medium Impact

| Change | Version | Impact on us | Action needed |
|---|---|---|---|
| **Byte-range advisory locks on block devices** (was whole-file locks) | v50.0 | Low risk -- better compat with network storage | Monitor for lock-related errors |
| **Seccomp filter fixes** | v50.0 | Positive -- fixes seccomp violations in vsock thread | Monitor for seccomp kills in CH vmm.log |
| **CPUID fixes in guest** | v50.0 | Positive -- unlikely to break anything | None expected |
| **Nested virtualization now configurable** (`nested=on|off`, default `on`) | v50.0 | No change in behavior | None |

### Low Impact / Positive (v50.0-v51.1)

- Live migration performance improvement
- Live disk resizing API (`/vm.resize-disk`)
- QCOW2 v3 improvements (live resize, variable refcount, dirty bit)
- DISCARD/WRITE_ZEROES support for virtio-blk
- THP for anonymous shared memory (performance)
- ACPI NUMA affinity for VFIO-PCI (GPU passthrough)

cmd/api/config/config.go
```diff
@@ -403,8 +404,9 @@ func defaultConfig() *Config {
 		},
 
 		Hypervisor: HypervisorConfig{
-			Default:               "cloud-hypervisor",
-			FirecrackerBinaryPath: "",
+			Default:                "cloud-hypervisor",
+			CloudHypervisorVersion: "v49.0",
```
sjmiller609: should this be the new version? EDIT: no, it makes more sense for this to be empty or none (since this is an optional configuration), then we are treating no config to mean use the latest version. That way we are only specifying the latest version in one place (not in this file, somewhere else)

lib/hypervisor/cloudhypervisor/process.go
```diff
@@ -19,6 +20,29 @@ import (
 	"gvisor.dev/gvisor/pkg/cleanup"
 )
 
+var (
+	defaultVersionMu sync.RWMutex
+	defaultVersion   = vmm.V49_0
```
sjmiller609: I think we don't need this block and the default version will be somewhere else

lib/hypervisor/cloudhypervisor/process.go
```diff
@@ -52,10 +76,11 @@ func (s *Starter) GetBinaryPath(p *paths.Paths, version string) (string, error)
...
 }
 
-// GetVersion returns the latest supported Cloud Hypervisor version.
-// Cloud Hypervisor binaries are embedded, so we return the latest known version.
+// GetVersion returns the configured default Cloud Hypervisor version.
+// This controls which version is used for new instances; existing instances
+// use the version stored in their metadata.
 func (s *Starter) GetVersion(p *paths.Paths) (string, error) {
-	return string(vmm.V49_0), nil
+	return string(GetDefaultVersion()), nil
```
sjmiller609, line 94: The default version should be only set in one place, and I think it should be in binaries_linux, e.g.:

```
type CHVersion string

const (
	V48_0 CHVersion = "v48.0"
	V49_0 CHVersion = "v49.0"
)

const DefaultVersion = V49_0

var SupportedVersions = []CHVersion{V48_0, V49_0}
```
(using the actual correct versions being set in this PR)

then what this function should do is return the default for new instances, otherwise return what's set in the instance metadata

lib/instances/version_upgrade_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	t.Log("Creating instance with CH v49.0...")
+	inst, err := mgr.CreateInstance(ctx, CreateInstanceRequest{
+		Name:           "version-upgrade-test",
+		Image:          integrationTestImageRef(t, "docker.io/library/alpine:latest"),
+		Size:           1024 * 1024 * 1024,
+		OverlaySize:    10 * 1024 * 1024 * 1024,
+		Vcpus:          1,
+		NetworkEnabled: false,
+		Hypervisor:     hypervisor.TypeCloudHypervisor,
```
sjmiller609: I am wondering if this line can specify the version

lib/vmm/README.md
```diff
@@ -75,10 +75,7 @@ lib/vmm/
 ├── version.go          # Version parsing utilities
 ├── binaries/           # Embedded Cloud Hypervisor binaries
 │   └── cloud-hypervisor/
-│       ├── v48.0/
-│       │   ├── x86_64/cloud-hypervisor    (4.5MB)
-│       │   └── aarch64/cloud-hypervisor   (3.3MB)
-│       └── v49.0/
```
sjmiller609, line 81: these lines should remain since we are showing how we can support multiple versions

lib/vmm/README.md
```diff
@@ -87,8 +84,7 @@ lib/vmm/
 
 ## Supported Versions
```
sjmiller609, line 87: we should add a concise section in here, for the readme changes I like to prompt like "Add a concise section that explains how it works (but not the structure of the code)"

lib/vmm/version.go
```diff
@@ -20,8 +20,8 @@ func ParseVersion(binaryPath string) (CHVersion, error) {
 	versionStr := strings.TrimSpace(string(output))
 
 	// Try to match known versions
-	if strings.Contains(versionStr, "v48.0") {
-		return V48_0, nil
```
sjmiller609, line 24: 👍

specs/cloud-hypervisor/api-v0.3.0/cloud-hypervisor.yaml
```diff
@@ -163,6 +163,22 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/VmResizeDisk"
+        required: true
+      responses:
+        204:
+          description: The disk was successfully resized.
+        500:
+          description: The disk could not be resized.
+
```
sjmiller609, line 181: How to deal with different APIs for each CH version. Here are the options that I thought about:
1. Multiple API specs and mappings between which CH version to which spec version
2. Pin API version per CH version
3. Use the same api version for every CH version (which is what this PR is doing).

**Breaking changes:** This API spec bump does not introduce breaking changes. I think we can stick with 3, then revisit 2 or 1 if there are breaking changes in an API. So, breaking changes are being ignored until actual happens so we don't introduce YAGNI right now, at that point we can worry about different cloud hypervisor API versions being supported simultaneously.

**New features only supported in some versions:** There are new features not supported by the old version (disk resize).

We don't have any API in hypeman that will use this feature, so we don't need to worry about it right now. In the future, we should handle this using `Capabilities`, like how we do in the case of each VMM. So if we did introduce disk resize as a feature, then we would have `Capabilities` also wired as version-specific.

```
type Capabilities struct {
    Snapshot       bool
    MemoryHotplug  bool
    DiskResize     bool   // ← new, true only for v51.1+
    DiskRateLimit  bool
    // ...
}

func capabilities(v vmm.CHVersion) Capabilities {
    caps := baseCH()
    switch v {
    case vmm.V49_0:
        // (default — disk resize stays false)
    case vmm.V51_1:
        caps.DiskResize = true
    }
    return caps
}
```

Action requested: if it's not complicated / or a lot of LOC: it might be nice to add that capability in just to establish the pattern of per-version capabilities. Since we are introducing the disk resize API, we want AI to realize it's not supported on 49.

**Actual changes in how we call previous version** So we are actually sending one new field to the old version: "image type". So in the case of Image type: raw for example, I think this field is basically ignored or it actually works, and that should be OK either way. But that's the thing to confirm, basically so long as we can still launch the VM with that field set, then we are good.

Makefile
```diff
@@ -116,7 +116,7 @@ download-ch-spec:
 	@echo "Downloading Cloud Hypervisor API spec..."
 	@mkdir -p specs/cloud-hypervisor/api-v0.3.0
 	@curl -L -o specs/cloud-hypervisor/api-v0.3.0/cloud-hypervisor.yaml \
-		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v48.0/vmm/src/api/openapi/cloud-hypervisor.yaml
+		https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/refs/tags/v51.1/vmm/src/api/openapi/cloud-hypervisor.yaml
```
sjmiller609, line 123: Add a comment above this that multiple API versions will only be needed once there are breaking changes introduced, and we should use version-specific `Capabilities` when new features are introduced that aren't supported on previous versions.

cmd/api/config/config.go
```diff
@@ -193,9 +193,10 @@ type CapacityConfig struct {
 
 // HypervisorConfig holds hypervisor settings.
 type HypervisorConfig struct {
-	Default               string                 `koanf:"default"`
-	FirecrackerBinaryPath string                 `koanf:"firecracker_binary_path"`
-	Memory                HypervisorMemoryConfig `koanf:"memory"`
+	Default                string                 `koanf:"default"`
+	CloudHypervisorVersion string                 `koanf:"cloud_hypervisor_version"`
```
sjmiller609: I think this config should be named "cloud_hypervisor_default_version"

lib/hypervisor/cloudhypervisor/config.go
```diff
@@ -63,7 +63,8 @@ func ToVMConfig(cfg hypervisor.VMConfig) vmm.VmConfig {
 	disks := make([]vmm.DiskConfig, 0, len(cfg.Disks))
 	for _, d := range cfg.Disks {
 		disk := vmm.DiskConfig{
-			Path: ptr(d.Path),
+			Path:      ptr(d.Path),
+			ImageType: ptr(vmm.Raw),
```
sjmiller609, line 67: This is the one actual change in how we touch v49. I'm assuming it's ignored, but maybe it's actually used. So the thing to check is that v49 still launches a VM then we are good.

lib/hypervisor/cloudhypervisor/process.go
```diff
@@ -19,6 +20,29 @@ import (
...
+		defaultVersion = chv
+	}
+}
+
+// GetDefaultVersion returns the current default Cloud Hypervisor version.
+func GetDefaultVersion() vmm.CHVersion {
+	defaultVersionMu.RLock()
+	defer defaultVersionMu.RUnlock()
+	return defaultVersion
+}
```
sjmiller609: I think we should have the version specified when we create the instance, not here

sjmiller609 (commented):
excellent, just a few small feedbacks

cmd/api/config/config.go
ulziibay-kernel: Done — renamed to `cloud_hypervisor_default_version`, default is now empty string. Empty means "use `vmm.DefaultVersion` from code" (defined once in `binaries_linux.go`). Override via `HYPERVISOR__CLOUD_HYPERVISOR_DEFAULT_VERSION=v51.1`.

lib/hypervisor/cloudhypervisor/process.go
ulziibay-kernel: Done — removed the hardcoded default here. `defaultVersionOverride` is now a nullable pointer; when nil (the default), `GetDefaultVersion()` falls back to `vmm.DefaultVersion` defined in `binaries_linux.go`.

lib/hypervisor/cloudhypervisor/process.go
ulziibay-kernel: Done — `const DefaultVersion = V49_0` now lives in `binaries_linux.go`. `GetVersion()` returns `GetDefaultVersion()` which falls back to `vmm.DefaultVersion` when no config override is set. Also added `CapabilitiesForVersion(v)` that returns version-specific capabilities (`SupportsDiskResize = true` for v51.1+).

lib/instances/version_upgrade_test.go
ulziibay-kernel: Done — added `HypervisorVersion` field to `CreateInstanceRequest`. The test now passes `HypervisorVersion: string(vmm.V49_0)` explicitly. When empty, it falls back to `starter.GetVersion()` (the configured default).

specs/cloud-hypervisor/api-v0.3.0/cloud-hypervisor.yaml
ulziibay-kernel: Agreed on all three points. Going with option 3 (same API spec) for now.

For the Capabilities pattern — added `CapabilitiesForVersion(v)` in `cloudhypervisor.go` with `SupportsDiskResize = true` for v51.1+. The base `capabilities()` called in `init()` uses `vmm.DefaultVersion` so it matches the default.

Re: `ImageType: Raw` on v49.0 — confirmed, v49.0's serde silently ignores unknown JSON fields (`image_type` didn't exist in the v49.0 DiskConfig). The field is sent but ignored, VM launches fine.

Makefile
ulziibay-kernel: Done — added comment above `download-ch-spec` in Makefile.

cmd/api/config/config.go
ulziibay-kernel: Done — renamed to `cloud_hypervisor_default_version`.

lib/hypervisor/cloudhypervisor/config.go
ulziibay-kernel: Confirmed — v49.0 silently ignores the `image_type` field in the JSON payload. It didn't exist in the v49.0 DiskConfig schema, so serde's default behavior (ignore unknown fields) applies. VMs launch fine.

lib/hypervisor/cloudhypervisor/process.go
ulziibay-kernel: Done — added `HypervisorVersion string` to `CreateInstanceRequest`. When set, it's used directly in `create.go`; when empty, falls back to `starter.GetVersion()` (the configured default). The E2E test now uses it to pin v49.0 explicitly.

lib/vmm/README.md
ulziibay-kernel: Done — kept the multi-version lines and updated the Architecture section to show both v49.0 and v51.1 directory trees.

lib/vmm/README.md
ulziibay-kernel: Done — added a "Multi-Version Support" section in the README explaining how it works: DefaultVersion, config override, metadata-pinned restore, and when it's safe to remove old versions.

lib/hypervisor/cloudhypervisor/config.go
ulziibay-kernel: CI tests pass with DefaultVersion = V49_0. All integration tests that boot VMs (TestStandbyAndRestore, TestCloudHypervisorSnapshotFeature, etc.) use v49.0 by default with ImageType: Raw set on all disks. VMs launch and run successfully

cmd/api/config/config.go
ulziibay-kernel: addressed

cmd/api/config/config.go
ulziibay-kernel: ack. CloudHypervisorDefaultVersion is empty now

lib/hypervisor/cloudhypervisor/config.go
ulziibay-kernel: Manually verified on devserver: `hypeman run alpine:latest` boots a VM on CH v49.0 (the default) with the v51.1 API spec that includes `image_type: Raw` on all disks. VM starts and runs successfully (`exit_code: 0`). The `image_type` field is silently ignored by v49.0's serde (it doesn't exist in the v49.0 DiskConfig schema, so unknown fields are dropped during deserialization). Also confirmed via CI integration tests which all pass with `DefaultVersion = V49_0`.

ulziibay-kernel: Added logging to verify it as suhc:

```
{"time":"2026-05-13T17:35:07.449869398-04:00","level":"INFO","msg":"starting VMM and booting VM","hypervisor":"cloud-hypervisor","instance_id":"jino20sjplg08sz6hu7nardc","hypervisor":"cloud-hypervisor","version":"v49.0","subsystem":"API"}
```

lib/hypervisor/cloudhypervisor/cloudhypervisor.go
```diff
@@ -60,6 +67,11 @@ func capabilities() hypervisor.Capabilities {
 		SupportsGracefulVMMShutdown: true,
 		SupportsSnapshotBaseReuse:   false,
 	}
+	switch v {
+	case vmm.V51_1:
+		caps.SupportsDiskResize = true
+	}
+	return caps
```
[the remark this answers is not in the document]
sjmiller609: ok for now, establishing pattern for later

lib/hypervisor/cloudhypervisor/process.go
```diff
@@ -52,10 +87,11 @@ func (s *Starter) GetBinaryPath(p *paths.Paths, version string) (string, error)
 	return vmm.GetBinaryPath(p, chVersion)
 }
 
-// GetVersion returns the latest supported Cloud Hypervisor version.
-// Cloud Hypervisor binaries are embedded, so we return the latest known version.
+// GetVersion returns the configured default Cloud Hypervisor version.
+// This controls which version is used for new instances; existing instances
+// use the version stored in their metadata.
```
sjmiller609, line 92: 👍

lib/vmm/binaries_darwin.go
```diff
@@ -12,10 +12,12 @@ import (
 type CHVersion string
 
 const (
-	V48_0 CHVersion = "v48.0"
 	V49_0 CHVersion = "v49.0"
+	V51_1 CHVersion = "v51.1"
 )
 
+const DefaultVersion = V49_0
```
sjmiller609, line 19: should we be defaulting to V51_1?

lib/vmm/binaries_linux.go
```diff
@@ -13,20 +13,22 @@ import (
...
 type CHVersion string
 
 const (
-	V48_0 CHVersion = "v48.0"
 	V49_0 CHVersion = "v49.0"
+	V51_1 CHVersion = "v51.1"
 )
 
-var SupportedVersions = []CHVersion{V48_0, V49_0}
+const DefaultVersion = V49_0
```
sjmiller609, line 29: should we be defaulting to V51_1? It would be OK in follow up if intentional

sjmiller609 (approved):
Optional feedbacks. Looks great! Thanks

lib/vmm/binaries_darwin.go
ulziibay-kernel: We can. For me, the least risky was to just keep it as is. Maybe test it in prod and then switch up the default

lib/vmm/binaries_linux.go
ulziibay-kernel: Yeah lets do it in a follow up. For me, the least risky was to just keep it as is. Maybe test it in prod and then switch up the default

PR #201, opened 2026-04-27
add active host diagnostics endpoint

hiroTamada:
## Summary

Adds `GET /diagnostics` that synchronously probes:

- **DNS** — external resolution of `kernel.sh` via the host's stub resolver
- **Egress TCP** — outbound TCP/443 to `1.1.1.1` (no DNS dependency)
- **Disk** — free bytes on the data dir, fails below a 5 GiB threshold

Each check returns `{name, ok, duration_ms, error?, detail?}`. The top-level `ok` is true iff all checks pass.

## Why

Push-based OTEL metrics silently fail when a host is partially unreachable, so the control plane has no signal for network or DNS breakage. This endpoint is designed to be polled from a control-plane scheduled workflow so the inability to even reach the host becomes its own observable signal — and per-check breakdowns localize the failure (DNS vs network vs disk).

The companion control-plane workflow lives in a separate kernel-kernel PR.

## Test plan

- [x] Unit tests for `probeDisk` (success and missing path)
- [x] Unit test that `runDiagnosticChecks` returns all three named checks
- [ ] Manual smoke against a staging hypeman host
- [ ] Verify the kernel-kernel `HypemanMonitorScheduledWorkflow` consumes the response shape correctly

## Notes

- `/diagnostics` requires bearer auth like the other authenticated endpoints
- Probes run in parallel with a 3s per-probe timeout
- 200 is returned regardless of probe outcomes; consumers inspect the body

### DIFF
--- a/cmd/api/api/diagnostics.go
+++ b/cmd/api/api/diagnostics.go
@@ -1,0 +1,175 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/kernel/hypeman/lib/logger"
+	"github.com/kernel/hypeman/lib/oapi"
+)
+
+// dnsProbeHost is the external hostname resolved during the DNS check.
+// kernel.sh is owned by Kernel and answered by stable infrastructure;
+// resolving it tells us the host's stub resolver and uplink DNS can answer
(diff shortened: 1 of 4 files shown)


sjmiller609:
closing as stale: the companion consumer in kernel/kernel#1865 was closed unmerged, and this draft now conflicts with current API code.

PR #203, opened 2026-04-29
Speed up instance name resolution

sjmiller609:
## Summary
- resolve instance names and ID prefixes from metadata before hydrating a single matched instance
- avoid `GetInstance(name)` falling back to full `ListInstances`, which queried every hypervisor socket
- keep existing API/CLI prefix behavior, but require DNS ID-prefix lookups to use at least 8 characters
- centralize metadata name lookup used by fork validation

## Validation
- `make build`

## Notes
This preserves exact ID and exact name lookup behavior. The stricter 8-character prefix rule only applies to internal DNS resolution so short DNS labels do not accidentally match broad ID prefixes.

---

> [!NOTE]
> **Medium Risk**
> Changes core instance lookup semantics and introduces a stricter prefix rule for DNS resolution, which could affect callers relying on short prefixes or on the previous list-based resolution behavior.
> 
> **Overview**
> Speeds up `GetInstance` lookups by resolving exact names and ID prefixes via metadata first, then hydrating only the single matched instance (avoiding a full `ListInstances`/hypervisor query fanout).
> 
> Adds a DNS-specific resolution path that requires an **8-character minimum ID prefix** for internal DNS queries (via `IngressResolver`) to reduce accidental broad prefix matches, and centralizes post-lookup side effects (persisting exit info/boot markers) plus metadata-only name existence checks used by fork validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58516810e3993b55937b451b606a44f35b4ef62f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/instances/manager.go
+++ b/lib/instances/manager.go
@@ -299,6 +299,15 @@ func (m *manager) maybePersistBootMarkers(ctx context.Context, id string) {
 	m.persistBootMarkers(ctx, id)
 }
 
+func (m *manager) finalizeResolvedInstance(ctx context.Context, inst *Instance) {
+	if inst.State == StateStopped && inst.ExitCode != nil {
+		m.maybePersistExitInfo(ctx, inst.Id)
+	}
+	if (inst.State == StateRunning || inst.State == StateInitializing) && inst.BootMarkersHydrated {
+		m.maybePersistBootMarkers(ctx, inst.Id)
+	}
+}
+
 func (m *manager) recordImageUsage(ctx context.Context, imageInfo *images.Image) {
 	if m.imageUsageRecorder == nil || imageInfo == nil {
 		return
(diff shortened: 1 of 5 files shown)


lib/instances/manager.go
```diff
@@ -499,66 +508,36 @@ func (m *manager) ListInstances(ctx context.Context, filter *ListInstancesFilter
...
-		}
-		if (inst.State == StateRunning || inst.State == StateInitializing) && inst.BootMarkersHydrated {
-			m.maybePersistBootMarkers(ctx, inst.Id)
-		}
+		m.finalizeResolvedInstance(ctx, inst)
 		return inst, nil
 	}
 
-	// 2. List all instances for name and prefix matching
-	instances, err := m.ListInstances(ctx, nil)
```
sjmiller609, line 520: this is the slow path I'm optimizing:
- get instance by name
- lists all instances
- listing all instances queries all hypervisors to get state

but we don't need all the VM's hypervisor states to get an instance by name. So instead we do a metadata-only lookup, and consolidate the other metadata lookup that was in fork to use the same.

hiroTamada (approved):
looks good — nice cleanup. main thing worth confirming is the behavior change in finalizeResolvedInstance: boot markers now persist on name/prefix matches too, where previously only exact-ID matches did. assuming that's intentional, ship it.

PR #214, opened 2026-05-08
test: verify firecracker fork-from-snapshot CoW isolation

sjmiller609:
## Summary

Adds an integration test that documents and guards the CoW isolation properties of firecracker fork-from-snapshot. When a fork descends from a source's standby snapshot, the fork's mem-file must be a separate inode (reflink-cloned by `forkvm.CopyGuestDirectory`) so that mutating the fork — including taking a diff snapshot after divergence — never affects the source.

## What the test does

`TestFirecrackerForkFromSnapshot_DoesNotMutateSource` (Linux-only, gated on `/dev/kvm`):

1. Boot a firecracker source instance.
2. Write a source-only sentinel into the guest.
3. Standby the source — produces a full firecracker snapshot.
4. Fingerprint the source's `snapshot-latest/memory` (inode + SHA-256).
5. Fork the source. Assert the fork's mem-file inode ≠ source's; source's mem-file fingerprint unchanged.
6. Restore the fork. Verify it sees the pre-fork sentinel.
7. Write a fork-only sentinel; standby the fork — produces a diff snapshot on the fork's local mem-file.
8. Re-fingerprint the source's mem-file. Must be byte-identical. Soft disk-usage check (gated on FICLONE probe): the fork lifecycle should consume substantially less than `guest_mem` on reflink-capable filesystems.
9. Delete the fork. Re-fingerprint the source's mem-file — still byte-identical.
10. Restore the source. Verify the pre-fork sentinel survived and the fork-only sentinel did not leak across.

The disk-usage assertion is gated on a runtime FICLONE probe; on ext4 and other non-reflink filesystems the copy falls back to sparse full-copy, so the "much-less-than-guest-mem" bound only holds on reflink-capable filesystems.

## Test plan

- [ ] CI runs `TestFirecrackerForkFromSnapshot_DoesNotMutateSource` on linux/amd64
- [ ] Disk-usage assertion stays under the bound on reflink-capable CI filesystems (xfs/btrfs); skipped with log on ext4

### DIFF
--- a/lib/instances/firecracker_test.go
+++ b/lib/instances/firecracker_test.go
@@ -4,12 +4,16 @@ package instances
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
@@ -26,6 +30,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 func setupTestManagerForFirecrackerWithNetworkConfig(t *testing.T, networkCfg config.NetworkConfig) (*manager, string) {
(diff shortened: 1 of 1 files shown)


lib/instances/fork.go
```diff
@@ -255,19 +255,39 @@ func (m *manager) forkInstanceFromStoppedOrStandby(ctx context.Context, id strin
...
 			return nil, fmt.Errorf("fork requires sparse-capable filesystem (SEEK_DATA/SEEK_HOLE unsupported): %w", err)
 		}
 		return nil, fmt.Errorf("clone guest directory: %w", err)
 	}
 
+	if shareMemFile {
+		if err := m.installForkSharedMemFile(dstDir, id); err != nil {
+			return nil, fmt.Errorf("install shared mem-file: %w", err)
+		}
+	}
```
sjmiller609: should we also do this for the root disk overlays? any other possible large files in the directory?

sjmiller609: no — mem-file is uniquely safe because firecracker mmaps it MAP_PRIVATE; disks are block-written and would corrupt the source.

lib/instances/fork.go
```diff
@@ -255,19 +255,39 @@ func (m *manager) forkInstanceFromStoppedOrStandby(ctx context.Context, id strin
...
+	// shareMemFile gates mem-file fan-out from the source's standby snapshot.
+	// Firecracker only: it mmaps the snapshot mem-file MAP_PRIVATE on restore,
+	// so all forks safely COW from the same backing file. Cloud-hypervisor and
+	// other hypervisors take a copy-mode path and don't benefit. Restricted to
+	// Template sources because they are explicitly promoted as fork-only and
+	// can never be restored — sharing the mem-file with a non-Template source
+	// would let a later RestoreInstance mutate the file out from under live
+	// forks.
+	shareMemFile := source.State == StateTemplate && stored.HypervisorType == hypervisor.TypeFirecracker
+
```
hiroTamada: I dont think you can use `StateTemplate` since it is going be deprecated

hiroTamada (approved):

PR #213, opened 2026-05-08
templates: add fork-only template registry and promotion lifecycle

sjmiller609:
## Summary

- New `lib/templates` package with a JSON file-backed `Registry` (Save/Get/GetByName/List/Delete + IncrementForkCount/DecrementForkCount).
- Adds `Template` model with hypervisor type/version, image+digest, vCPU/memory, fork refcount, and a reserved `HotPagesPath` field for PR 8.
- Wires `lib/instances` template plumbing: promote/list/get/delete plus `paths.TemplatesDir()` etc.
- Stores `IsTemplate` / `TemplateID` / `ForkOfTemplate` on instance metadata; the manager owns a `templates.Registry` it constructs at startup.

## Why

Foundation for snapshot fan-out: templates are the durable "tagged Standby instance promoted to fork-only parent" record that downstream PRs (shared mem-file, GC, UFFD page server) hang off of. Splitting it out keeps each follow-up reviewable.

## Test plan

- [ ] `go test ./lib/templates/...`
- [ ] `go test ./lib/instances/...` for unit coverage of new metadata fields
- [ ] Manual: promote a Standby instance, list templates, attempt to delete with a live fork (must refuse), unfork, delete

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #215, opened 2026-05-08
templates: reconcile fork refcounts and sweep orphans on boot

sjmiller609:
**Stacked on:** #214 (shared mem-file) — review #213 → #214 first.

## Summary

- Adds `Registry.Reconcile(ctx, observed map[string]int)` that walks every template and rewrites `ForkCount` to match the observed map.
- Adds manager-side `reconcileTemplateState` that scans every instance's metadata, sums `ForkOfTemplate` references, calls `Registry.Reconcile`, then sweeps registry entries whose source instance is gone and `ForkCount` is 0.
- Wires reconciliation into manager init and exposes `ReconcileTemplateState` for callers that want to re-run it.
- Tests: `TestFileRegistry_Reconcile` exercises 3 templates with starting counts 5/0/7 reconciled to 2/3/0.

## Why

`ForkCount` drifts when forks are deleted out from under hypeman (crash, SIGKILL, on-disk surgery). Without reconciliation a template can be permanently "in use" and refuse delete. Reconciling on boot keeps the registry honest and lets `Delete` block on actual live forks instead of stale counters.

## Test plan

- [ ] `go test ./lib/templates/...`
- [ ] `go test ./lib/instances/...`
- [ ] Manual: kill a fork's process, drop its metadata, restart hypeman, observe `Reconcile` repaired the count and the template can be deleted

### DIFF
--- a/lib/forkvm/copy.go
+++ b/lib/forkvm/copy.go
@@ -11,10 +11,28 @@ import (
 
 var ErrSparseCopyUnsupported = errors.New("sparse copy unsupported")
 
+// CopyOptions tunes CopyGuestDirectory behavior. The zero value reproduces
+// the original full-copy semantics; callers can opt into skipping specific
+// paths when the consumer arranges its own substitute (e.g. a symlink to a
+// template-shared mem-file).
+type CopyOptions struct {
+	// SkipRelPaths lists relative paths under srcDir that should not be
+	// materialized in dstDir. Comparison is exact and uses forward-slash
+	// separators on all platforms.
+	SkipRelPaths []string
+}
+
 // CopyGuestDirectory recursively copies a guest directory to a new destination.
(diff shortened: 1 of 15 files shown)


sjmiller609:
Superseded by the state-machine redesign in #229. The original Registry+reconcile pair was specific to the separate file-based template registry, which is gone now — `ForkCount` lives on the instance row, decrement happens under the parent's instance lock atomically with each fork delete (`decrementTemplateForkCount` in lib/instances/delete.go), and Delete refuses Template with `ForkCount > 0`.

The drift concern (forks rm -rf'd off disk, manual metadata edits, etc.) is still valid, but lighter weight now: a boot-time scan that rewrites each Template's `ForkCount` from observed `ForkOfTemplate` references — and demotes to Standby when observed is zero — would be a small follow-up if drift shows up. Closing this one rather than retargeting since the implementation surface changed completely.

PR #220, opened 2026-05-08
api: share_memory fork mode + fork_count/mem_locked on Instance

sjmiller609:
## Summary

Exposes the uffd/template-fork machinery through the existing `POST /instances/{id}/fork` endpoint, without introducing a new `/templates` resource.

- `ForkInstanceRequest.share_memory: bool` — when true, the source instance is auto-promoted (idempotently) into a fan-out parent and the fork attaches to its mem-file instead of copying. Subsequent share_memory forks against the same source reuse the same registry entry.
- `Instance.fork_count: int` and `Instance.mem_locked: bool` — read-only fields so clients can introspect lock state on the existing `GetInstance` response.
- While `mem_locked = true`, start/restore/delete on the source return 409 (`ErrInvalidState`) until all share_memory forks are deleted. `templateGuard` now returns `ErrInvalidState` rather than `ErrNotSupported` so the lock surfaces as a transient 409, not a 501 capability gap.
- Validation: `share_memory` is incompatible with `template_id` and with `from_running=true` (the latter would re-restore the source after locking).

The `lib/templates` package and `TemplateID` field stay as internal machinery — no public templates resource. `share_memory=false` forks are unchanged and may still be created against a locked source.

## Test plan

- [x] `go vet ./...` clean
- [x] New unit tests in `lib/instances/share_memory_test.go`:
  - `TestValidateForkRequest_ShareMemoryConflicts` — rejects share_memory + template_id and share_memory + from_running
  - `TestEnsureShareMemoryTemplate_AutoPromoteAndReuse` — first call promotes, second reuses
  - `TestEnsureShareMemoryTemplate_RejectsNonStandby` — non-Standby source returns ErrInvalidState
  - `TestTemplateGuard_ReturnsInvalidStateNotUnsupported` — guard now maps to 409
  - `TestHydrateForkLockState` — fork_count/mem_locked reflect refcount changes
- [ ] CI on green firecracker E2E + cross-hypervisor regression


sjmiller609:
Closing as obsolete — superseded by the state-machine template design in #229.

The features this PR added are now provided by the state machine intrinsically:

- **`share_memory` opt-in**: no longer needed. In the state-machine design, the first fork from a firecracker Standby source implicitly promotes it to `StateTemplate`, and the fork attaches to the source's mem-file via the existing fan-out path. This is the default — there is no opt-out.
- **`mem_locked`**: redundant with `state == "Template"`. `StateTemplate` cannot wake/restore/delete while `ForkCount > 0`; clients already see this via the existing `state` field.
- **`fork_count` / template introspection**: `StoredMetadata.ForkCount` already exists; exposing it as a public API field can be a follow-up if there's demand.
- **`templateGuard` → `ErrInvalidState`**: handled by the state-machine transitions in #229.

If exposing `fork_count` / `is_template` / `fork_of_template` on the Instance API turns out to be useful, that should be a separate, much smaller PR.

PR #225, opened 2026-05-11
Speed up listing VMs

sjmiller609:
## Summary

Cache hypervisor state and bound /vm.info calls in list path

Builds on the list-path tracing PR (#208). Those spans showed single `instances.derive_state` spans of 10–37s on the dev fleet, with **no children** — all the time was inside the un-instrumented `hv.GetVMInfo` HTTP call. The cloud-hypervisor API socket serializes per VM, so a snapshot in flight from `auto_standby` parks every concurrent `/vm.info` call behind it, blocking the whole serial list loop.

This PR addresses both halves of the problem:

1. **Cache the last observed hypervisor VM state** per instance with a 5s TTL. List calls within the TTL skip the HTTP round-trip entirely. Lifecycle event notifications (`start` / `stop` / `standby` / `restore` / `delete` / …) write through the cache so it stays current with state changes hypeman itself drives.

2. **Bound the underlying `GetVMInfo` call with a 500 ms timeout.** On timeout the instance is reported as `Unknown` (matching today's behavior when the hypervisor is unreachable) and the next list call retries.

The TTL preserves the original purpose of the call — detecting guest-driven shutdowns the lifecycle bus can't see — while collapsing bursty list calls onto at most one `/vm.info` per instance per 5s. The timeout caps the worst case at 500 ms instead of 37 s.

## Test plan
- [x] `go vet ./lib/instances/...` clean
- [x] New unit tests cover TTL expiry, lifecycle event update, delete invalidation, and the live-socket vs no-socket mapping in `updateCachedHypervisorStateFromInstance`
- [x] Existing query / lifecycle / admission / parse tests pass
- [ ] After deploy: confirm `instances.derive_state` p99 on dev hypemen drops; confirm `hypervisor_state.cache_hit=true` shows up on most list calls

---

> [!NOTE]
> **Medium Risk**
> Changes instance state derivation to reuse cached hypervisor state and to timeout `GetVMInfo`, which could alter list results (e.g., more `Unknown` states) or temporarily surface stale state within the TTL window.
> 
> **Overview**
> Speeds up hot instance listing/state-derivation paths by adding an in-memory per-instance hypervisor state cache (5s TTL) and using it to avoid repeated `/vm.info` socket calls.
> 
> When a cache miss occurs, `deriveState` now wraps `hv.GetVMInfo` with a 500ms timeout and returns `StateUnknown` on failure/timeout rather than blocking for long periods. Lifecycle notifications write-through to the cache and deletions/no-socket states invalidate it, and new unit tests cover TTL behavior and cache update/invalidation via lifecycle events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86d079fa6d97478e3e24a2f1beec020b460bced1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/instances/query.go
+++ b/lib/instances/query.go
@@ -25,7 +25,23 @@ const (
 	programStartSentinelPrefix = "HYPEMAN-PROGRAM-START "
 	agentReadySentinelPrefix   = "HYPEMAN-AGENT-READY "
 	bootMarkerRescanInterval   = 1 * time.Second
+	// hypervisorStateCacheTTL bounds how long a cached hypervisor state may be
+	// reused before a fresh GetVMInfo call is required. Short enough to detect
+	// guest-driven shutdowns promptly, long enough that bursty list calls
+	// collapse onto one underlying socket query.
+	hypervisorStateCacheTTL = 5 * time.Second
+	// getVMInfoTimeout bounds a single hypervisor /vm.info query. The cloud
+	// hypervisor API socket serializes requests, so a snapshot in flight can
+	// otherwise park derive_state for tens of seconds.
+	getVMInfoTimeout = 500 * time.Millisecond
 )
+
+// hypervisorStateCacheEntry stores the last observed hypervisor VM state for
+// an instance, used to skip /vm.info queries on hot list paths.
+type hypervisorStateCacheEntry struct {
+	state       hypervisor.VMState
+	refreshedAt time.Time
+}
 
 // stateResult holds the result of state derivation
 type stateResult struct {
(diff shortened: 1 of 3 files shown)


lib/instances/query.go
```diff
@@ -56,39 +72,55 @@ func (m *manager) deriveStateWithOptions(ctx context.Context, stored *StoredMeta
...
-			"socket", stored.SocketPath,
-			"error", err,
-		)
-		return stateResult{State: StateUnknown, Error: &errMsg}
-	}
+	// 2. Socket exists - resolve hypervisor state, preferring the in-memory
+	// cache (populated by lifecycle events and prior queries) and falling
+	// back to a time-bounded /vm.info call.
+	var hvState hypervisor.VMState
+	if cached, ok := m.loadCachedHypervisorState(stored.Id); ok {
```
hiroTamada, line 86: should we add a `no-cache`/fresh-state option for read endpoints before making this the default? The cache seems reasonable for list latency, but `GET /instances/:id` is often used as an authoritative check or for polling after operations. Giving read callers a way to bypass the cached state would reduce the risk of acting on state that can be stale for up to 5s.

hiroTamada (commented):
Left one comment.

lib/instances/query.go
sjmiller609: I don't think so because we have cache clearing on state change, the timeout is 5s anyways, and there is a wait for state endpoint

hiroTamada: https://github.com/kernel/kernel/blob/main/packages/api/lib/computebackend/hypeman/instance.go#L464-L494 

?

sjmiller609: Yes that's right, and there is a public API. but my point is that list instances shouldn't return a stale result because the cache is cleared on state transition

hiroTamada: ah I see I see. Sorry about that

hiroTamada (approved):

lib/instances/query.go
sjmiller609: no prob, appreciate it

PR #229, opened 2026-05-13
Introduce "template" as a VM state

sjmiller609:
## Summary

Models "Template" as a state in the instance state machine. A Standby instance is auto-promoted to Template the first time it's forked from a snapshot. `ForkCount` is bumped on each subsequent fork and decremented when a fork is deleted. Templates cannot wake while `ForkCount > 0`; un-promote (Template -> Standby) and delete (Template -> Stopped) are both refused until forks drain. `RestoreInstance` on a Template with zero forks transparently un-promotes and then restores.

Three new fields live on `StoredMetadata`: `IsTemplate`, `ForkCount`, `ForkOfTemplate`. A fourth field `HotPagesPath` is reserved for the future UFFD prefetch path and is cleared on un-promote.

State derivation in `query.go` returns `StateTemplate` when there's a snapshot, no socket, and `IsTemplate=true`. Template is a derived state alongside Standby/Stopped — no extra files on disk, no separate index.

The running-fork flow (`from_running=true`) skips promotion: it restores the source back to Running afterward, and a template can't wake.

Fork delete decrements the parent's `ForkCount` under the parent's lock after the fork's data is gone. Worst case on partial failure is drift (parent count higher than reality), fixed by a future reconciliation pass.

## Test plan

- [x] `go build ./...` clean
- [x] Existing instance tests pass (`go test ./lib/instances/...`)
- [x] New unit tests in `templates_test.go` cover: deriveState returns Template, Restore refused on Template with forks, Restore demotes Template with zero forks, Delete refused on Template with forks, Delete-fork decrements parent ForkCount
- [x] State-transition tests cover the new Standby<->Template and Template->Stopped edges
- [ ] End-to-end fork-from-Template flow (not exercised in unit tests; needs real snapshot/VMM)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new lifecycle state and new API endpoints that alter instance fork/restore/delete behavior, which could affect VM lifecycle orchestration and safety checks around snapshot sharing. Main risk is incorrect state derivation or fork-counting logic leading to templates being demoted/deleted while forks still exist (or being unnecessarily blocked).
> 
> **Overview**
> Adds a new `Template` state to the instance state machine (derived via `StoredMetadata.IsTemplate`) to represent snapshot-backed, *fork-only* parents.
> 
> Introduces explicit lifecycle operations `PromoteToTemplate` and `DemoteTemplate` (plus `POST /instances/{id}/promote-template` and `POST /instances/{id}/demote-template`) and updates auth scopes and generated OpenAPI client/server code accordingly.
> 
> Updates forking and lifecycle guards so templates can be forked from, forks of templates record `ForkOfTemplate`, and **demote/delete are refused while live forks exist** (counted by scanning metadata); restores and snapshot restores are also **refused on templates** until demoted. Adds focused unit tests covering template state derivation, fork counting, and refusal/idempotency behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c140e4ff0f8dd3b40cfc59a9c8dcf349f9d50e53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/instances/delete.go
```diff
@@ -136,10 +141,41 @@ func (m *manager) deleteInstance(
...
+	}
+
 	log.InfoContext(ctx, "instance deleted successfully", "instance_id", id)
 	return nil
 }
 
+// decrementTemplateForkCount drops the parent template's ForkCount by one under
+// the parent's lock. Logs but does not return errors: the fork is already gone,
+// so the worst case is refcount drift that a future reconciliation pass fixes.
+func (m *manager) decrementTemplateForkCount(ctx context.Context, parentID string) {
```
sjmiller609: should we be tracking fork count in a denormalized way? it seems like the fork count should be based from all the existing instances that are forked from the template, not by tracking how many references there are, as that could become out of sync?

sjmiller609: good call — dropped the denormalized `ForkCount` field and the decrement logic in `DeleteInstance`. fork count is now computed at read time by `countTemplateForks` (scans `listMetadataFiles` for `ForkOfTemplate == templateID`). single source of truth, no drift, no reconciliation pass needed. trade is O(N) on template delete / un-promote (N = instances per host), which is fine for those rare ops.

cmd/api/api/instances.go
```diff
@@ -614,6 +614,80 @@ func (s *ApiService) RestoreInstance(ctx context.Context, request oapi.RestoreIn
 	return oapi.RestoreInstance200JSONResponse(instanceToOAPI(*result)), nil
 }
 
+// PromoteInstanceToTemplate promotes a standby instance into a fork-only template.
+// The id parameter can be an instance ID, name, or ID prefix.
+// Note: Resolution is handled by ResolveResource middleware.
```
sjmiller609, line 619: A template is different than a standby because it cannot be resumed.

Options from the UX perspective:
- template lock as a concept within standby
  - User forks from standby, with an option like "share memory true" or similar
  - This results in the standby being "locked" in standby: seems surprising
- (the option this is doing) User specifically transitions a standby instance into "template" state
  - when you fork from template, it will automatically uses the shared mem optimizations (not included in this PR)
  - you can't un-template until all references are gone, then you could transition it back into a standby or delete it

lib/instances/manager.go
```diff
@@ -454,6 +462,69 @@ func (m *manager) RestoreInstance(ctx context.Context, id string) (*Instance, er
...
+	if inst.State != StateStandby {
+		return nil, fmt.Errorf("%w: cannot promote instance in state %s to template (must be Standby)", ErrInvalidState, inst.State)
+	}
+	meta.IsTemplate = true
+	if err := m.saveMetadata(meta); err != nil {
+		return nil, fmt.Errorf("save metadata after template promote: %w", err)
+	}
+	promoted := m.toInstance(ctx, meta)
+	return &promoted, nil
+}
```
sjmiller609, line 489: a template is the same thing as a standby from the perspective of the VMM. But it just changes what hypeman is allowed to do with it, so all we are doing here is writing the metadata under a lock

lib/instances/query.go
```diff
@@ -822,6 +825,27 @@ func (m *manager) listInstances(ctx context.Context) ([]Instance, error) {
 	return result, nil
 }
 
+// countTemplateForks returns the number of instances with ForkOfTemplate==templateID.
+// Skips metadata files that fail to load (logged elsewhere on read).
+func (m *manager) countTemplateForks(templateID string) (int, error) {
+	files, err := m.listMetadataFiles()
```
sjmiller609, line 833: metadata only scan

lib/instances/query.go
```diff
@@ -822,6 +825,27 @@ func (m *manager) listInstances(ctx context.Context) ([]Instance, error) {
...
+		meta, err := m.loadMetadata(id)
+		if err != nil {
+			continue
+		}
+		if meta.ForkOfTemplate == templateID {
+			count++
+		}
+	}
+	return count, nil
+}
```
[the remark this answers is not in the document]
sjmiller609: fixed in 2738278 — `countTemplateForks` now bubbles `loadMetadata` errors instead of `continue`ing, so demote/delete fail closed if any fork's metadata becomes unreadable rather than risk freeing a template whose pages are still mapped.

hiroTamada (approved):
One thing to tighten before/after merge: `RestoreSnapshot` is still allowed on a promoted `Template` because the in-place restore path only rejects `Running` sources before replacing the instance directory. That bypasses the new delete/demote invariant for templates with live forks.

Concretely, a template `T` with forks where `ForkOfTemplate == T` can still have an existing snapshot restored onto `T`, which rewrites the template payload/metadata while forks still reference it. I’d treat `StateTemplate` like delete/demote here: if `countTemplateForks(id) > 0`, return `ErrInvalidState` before `replaceInstanceWithSnapshotPayload`; optionally require explicit demotion for idle templates depending on intended API semantics.

lib/instances/fork.go
```diff
@@ -286,17 +291,22 @@ func (m *manager) forkInstanceFromStoppedOrStandby(ctx context.Context, id strin
...
-	case StateStopped:
+	} else {
 		forkMeta.Phases.Record(phasetracking.PhaseStopped, now)
 	}
 
+	// Template-only fields don't carry forward to the fork; the fork is a fresh
+	// instance regardless of whether the parent is a template.
+	forkMeta.IsTemplate = false
+	forkMeta.HotPagesPath = ""
+	forkMeta.ForkOfTemplate = ""
```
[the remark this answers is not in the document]
sjmiller609: skipping — the bot's own analysis above concludes "no actual bug here on further analysis." the comment at lines 300-301 is accurate in its narrow sense: forks never *inherit* template fields. the line-349 assignment is a fresh assignment from parent identity, not field inheritance. the fragile-coupling concern is theoretical — an instance with IsTemplate=true whose derived state is not Template is impossible today, since deriveState uses IsTemplate as the discriminator.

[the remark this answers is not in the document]
sjmiller609: same region as the previous comment, same conclusion. the bot agrees logic is correct. the fragile-coupling concern requires IsTemplate=true on an instance whose derived state is not Template, which deriveState rules out. leaving as-is.

lib/instances/delete.go
```diff
@@ -35,6 +35,16 @@ func (m *manager) deleteInstance(
...
 
+	if inst.State == StateTemplate {
+		forks, err := m.countTemplateForks(id)
+		if err != nil {
+			return fmt.Errorf("count forks of template %s: %w", id, err)
+		}
+		if forks > 0 {
+			return fmt.Errorf("%w: cannot delete template %s with %d live fork(s); delete forks first", ErrInvalidState, id, forks)
+		}
+	}
```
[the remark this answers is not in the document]
sjmiller609: skipping. the safety invariant — no deletion while live forks (= reflinks to the template's snapshot extents) exist — is enforced at delete.go:38-46 via countTemplateForks. adding StateTemplate.CanTransitionTo(StateStopped) only for the Template branch would be the sole state-machine check in deleteInstance, which doesn't validate transitions for any other state (it's the burn-it-all-down path). consistent options are "validate every state" or "none" — adding one cosmetically here is cargo-cult. the fork-count guard is the real safety boundary.

PR #230, opened 2026-05-15
Add AWS deployment assets

rgarcia:
## Summary
- Add a maintained AWS deploy path centered on a hosted CloudFormation Launch Stack quickstart.
- Consolidate AWS deploy assets to CloudFormation only by removing the Terraform wrapper, Packer AMI starter, and duplicate standalone validation script.
- Make the Hypeman host a stack-managed AWS::EC2::Instance so deleting the CloudFormation stack deletes the instance and stack-managed volumes.
- Keep a small custom launch-template resource only for the nested-virtualization CPU option because CloudFormation does not expose that EC2 setting on typed resources yet.
- Provision a separate encrypted EBS data volume, format it as XFS, mount it at /var/lib/hypeman, and validate that mount during bootstrap.
- Strip validated registry Authorization headers before forwarding registry requests so authenticated remote pushes reach the embedded registry cleanly.

## Registry auth note
Hypeman has two JWT paths that both remain supported:

- User/API JWTs from hypeman-create-token, used by the CLI for normal remote API access, including hypeman push.
- Registry-scoped JWTs with repo_access or repos plus scope: push, used by the registry auth flow for builder/build access.

The middleware change does not remove registry-scoped push JWT support. For /v2/... requests, JwtAuth first validates registry-scoped tokens with repo/scope checks, then falls back to validating regular user tokens for direct registry access. After either token type is accepted, the request is forwarded without the original Authorization header because Hypeman middleware is the auth boundary; the embedded registry does not understand Hypeman JWTs and can reject requests if the already-validated bearer token is passed through.

## Testing
- cfn-lint deploy/aws/cloudformation/template.yaml
- go test ./deploy/aws/cloudformation
- go test ./lib/middleware
- git diff --check
- AWS e2e in us-east-1 from the PR template: launched a temporary CloudFormation stack, confirmed nested-virtualization launch-template creation, waited for bootstrap, verified /var/lib/hypeman mounted as XFS, generated a JWT, called the public health endpoint with bearer auth, installed/configured the CLI, built and pushed a Claude Code Docker image, waited for image import to reach ready, ran it, executed claude --version successfully, stopped/removed the workload, and deleted the stack.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it introduces new AWS CloudFormation infrastructure/bootstrap logic and changes registry request forwarding by stripping `Authorization` headers, which could affect image push/pull flows if misapplied.
> 
> **Overview**
> Adds a maintained AWS deployment path via a new CloudFormation “Launch Stack” template and documentation, including EC2 bootstrap that enables nested virtualization, provisions an encrypted XFS data volume at `/var/lib/hypeman`, and exposes the API only to a configurable CIDR (optional SSH).
> 
> Introduces a `Deploy Assets` GitHub Actions workflow to `cfn-lint`/unit test the template on PRs and publish the versioned template to S3 on `main`.
> 
> Adjusts registry-path JWT auth handling to forward validated requests **without** the incoming `Authorization` header (helper `requestWithoutAuthorization`), and updates middleware tests to assert the header is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d10fabb676267a336c5dd46b58e6569b48a68b89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


deploy/aws/ami/packer/hypeman.pkr.hcl
```diff
@@ -0,0 +1,66 @@
...
+}
+
+build {
+  sources = ["source.amazon-ebs.hypeman"]
+
+  provisioner "shell" {
+    inline = [
+      "set -euxo pipefail",
+      "sudo apt-get update",
+      "sudo DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates curl docker.io e2fsprogs erofs-utils iproute2 iptables jq openssl qemu-system-x86 qemu-utils tar",
```
sjmiller609: install the native compression libraries used by snapshot compression

deploy/aws/cloudformation/template.yaml
```diff
@@ -0,0 +1,505 @@
...
+                Resource: !GetAtt HypemanInstanceRole.Arn
+
+  LaunchFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      Runtime: python3.12
+      Handler: index.handler
+      Timeout: 900
+      Role: !GetAtt LaunchFunctionRole.Arn
+      Code:
```
sjmiller609, line 224: wait... so why are we deploying an ec2 instance via a lambda function instead of just creating it via cloudformation?

deploy/aws/cloudformation/template.yaml
```diff
@@ -0,0 +1,505 @@
...
+      KeyName: !If [UseSSH, !Ref KeyName, ""]
+      RootVolumeSize: !Ref RootVolumeSize
+      UserData: !Sub |
+        #!/usr/bin/env bash
+        set -euxo pipefail
+        exec > >(tee /var/log/hypeman-bootstrap.log | logger -t hypeman-bootstrap -s 2>/dev/console) 2>&1
+
+        export DEBIAN_FRONTEND=noninteractive
+        apt-get update
+        apt-get install -y ca-certificates curl docker.io e2fsprogs erofs-utils iproute2 iptables jq openssl qemu-system-x86 qemu-utils tar
```
sjmiller609: wait isn't this the stuff that would run inside of packer build? and basically when AMI is build then launched it comes up with hypeman and everything already installed, systemd unit files set up so that when the machine boots it initializes one time stuff like random JWT secret then starts things up? So all that gets set up in the packer build, then the cloudformation template is just launching the AMI that's ready to go

deploy/aws/terraform/main.tf
```diff
@@ -0,0 +1,23 @@
+provider "aws" {
```
sjmiller609: does this terraform just exist so that someone can deploy via terraform? all it's doing is calling the cloudformation

deploy/aws/cloudformation/template.yaml
```diff
@@ -0,0 +1,505 @@
...
+      ApiPort:
+        default: Hypeman API port
+      EnableSSH:
+        default: Enable SSH
+      AllowedSshCidr:
+        default: SSH access CIDR
+      KeyName:
+        default: EC2 key pair
+      RootVolumeSize:
+        default: Root volume size
```
sjmiller609, line 51: Will this be XFS? some of the firecracker forking features will depend on XFS to work optimally. If root devices are not XFS, then we ought to both have increased root block storage and also a separate disk, which in the provisioner script set set up with xfs

sjmiller609 (commented):
I want to sync on packer -> ami versus startup lambda function

sjmiller609 (approved):

PR #232, opened 2026-05-15
Use snapshots as fanout parents

sjmiller609:
## Summary
- record snapshot ids on instances forked from snapshots
- expose `ref_count` on snapshot API responses
- refuse explicit snapshot deletion while direct snapshot forks still exist
- hardlink raw snapshot memory files into snapshot forks and assert the shared inode path in tests
- skip referenced snapshots during scheduled snapshot cleanup

## Tests
- go test ./lib/forkvm/
- go test ./lib/instances/ -run 'Test(StoppedSnapshotLifecycleAndForkAfterSourceDeletion|DeleteSnapshotFailsWhenForkMetadataUnreadable|ForkSnapshotHardlinksRawMemoryFile|ForkSnapshotFromCompressedSourceCopiesRawMemory|RunSnapshotSchedules|SnapshotSchedule)'\n- go test ./lib/snapshot ./cmd/api/api -run 'TestNonexistent'\n\nNote: a full go test ./lib/instances/ run was stopped after it entered slower runtime-adjacent coverage; the targeted snapshot/fork/schedule coverage above passed.

---

> [!NOTE]
> **Medium Risk**
> Changes snapshot lifecycle semantics and the public API schema (new required `ref_count` and `409` on delete), which may affect clients and cleanup behavior. Forking now hardlinks snapshot memory files, so filesystem/linking edge cases could impact fork reliability.
> 
> **Overview**
> **Snapshots now act as fanout parents.** Instances created via `ForkSnapshot` persist the source snapshot ID in metadata (`ForkOfSnapshot`), and snapshot `GET`/`LIST` responses expose a new required `ref_count` field.
> 
> **Deletion semantics tighten.** Explicit snapshot deletion now fails with `409 Conflict` when any direct forks still exist (and scheduled cleanup silently skips those snapshots), and the OpenAPI/client code is updated accordingly.
> 
> **Fork performance/space change.** Snapshot forks avoid copying the raw memory file by skipping it during directory copy and then hardlinking it into the forked instance; tests add coverage for ref counting, unreadable metadata, and inode sharing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05e4a11453b44c2734dae553426a3d25c7396a92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hiroTamada (approved):
Looks good overall. The snapshot fanout direction makes sense to me as the replacement for template-based fanout, and blocking deletion while direct snapshot forks exist is the right shape.

One potential issue worth double-checking: Firecracker snapshot forks hardlink the raw memory file from the central snapshot into the fork. After that fork is restored/running, Firecracker keeps `snapshot-latest` as the retained base. On a later `Running -> Standby`, the retained base is moved back to `snapshot-latest`, `toSnapshotCreateParams` sees `memory` already exists, and sends `snapshot_type: "Diff"` with the same `mem_file_path`.

Firecracker's diff snapshot behavior can write directly on top of an existing memory file. If that retained base memory file is still hardlinked to the central parent snapshot memory, the fork's standby operation could mutate the supposedly immutable parent snapshot memory file through the shared inode.

I would either add a targeted Firecracker lifecycle test for `ForkSnapshot -> Restore/Run -> Standby` asserting the parent snapshot memory inode/content is unchanged, or explicitly break the hardlink before a snapshot-forked Firecracker VM enters standby.

sjmiller609:
I'm holding off on this until we decide to look at UFFD because it's not necessary to hold the snapshots until then

PR #238, opened 2026-05-19
Decouple bandwidth reservations from oversubscription

rgarcia:
## Summary

- decouple default per-VM network and disk I/O reservations from the same oversubscription ratios used for admission
- keep oversubscription ratios as aggregate admission multipliers via resource effective limits
- optionally configure the AWS quickstart gp3 data volume IOPS/throughput, and set `CAPACITY__DISK_IO` only when `DataVolumeThroughput` is provided

## Why

Default per-VM bandwidth reservations previously used the oversubscribed effective capacity as their input. That meant raising the oversubscription ratio increased the aggregate limit and each default VM request by the same factor, so the ratio did not let the host pack more default VMs when that resource was the binding constraint.

Illustrative example for disk I/O on a 32-vCPU host with 1GB/s disk I/O and 1-vCPU VMs:

| ratio | effective disk I/O limit | old default per VM | max default VMs |
|---:|---:|---:|---:|
| 1.0 | 1GB/s | 31.25MB/s | 32 |
| 2.0 | 2GB/s | 62.5MB/s | 32 |
| 4.0 | 4GB/s | 125MB/s | 32 |

Network had the same shape. This change makes the default per-VM reservation proportional to raw capacity, while admission still uses `capacity * oversubscription_ratio`.

## Validation

- `go test ./lib/resources`
- `go test ./...` in `deploy/aws/cloudformation`
- `git diff --check`

A full repo-wide `go test ./...` was not run successfully from this checkout because some test packages require embedded local binaries that are not present in the repository checkout.

### DIFF
--- a/deploy/aws/cloudformation/template.yaml
+++ b/deploy/aws/cloudformation/template.yaml
@@ -17,6 +17,8 @@ Metadata:
           - InstanceType
           - RootVolumeSize
           - DataVolumeSize
+          - DataVolumeIops
+          - DataVolumeThroughput
           - AmiSsmParameter
       - Label:
           default: Access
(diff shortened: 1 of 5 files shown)


sjmiller609 (approved):
makes sense, thanks

PR #239, opened 2026-05-22
Preserve snapshot fork restore paths

sjmiller609:
## Summary
- add a Firecracker regression for snapshot fork -> running fork after deleting the original snapshot source
- retain Firecracker snapshot source aliases when the original source dir is already gone
- preserve retained upstream aliases across later snapshot forks and network rewrites
- add the same warm fork chain coverage for QEMU and Cloud Hypervisor
- persist QEMU restore configs so later standby snapshots do not copy stale source paths
- avoid rotating the restored source CID when the fork target is stopped and no live/snapshotted fork can collide
- refresh stopped fork CIDs after removing their standby snapshot
- return Firecracker source stat errors instead of treating them as usable existing paths

## Test plan
- go test ./lib/hypervisor/firecracker -run 'TestPrepareFork|TestWithSnapshotSourceDirAlias' -count=1 -v\n- go test ./lib/hypervisor/qemu -count=1\n- go test ./lib/hypervisor/qemu ./lib/hypervisor/firecracker -run 'TestPrepareFork|TestWithSnapshotSourceDirAlias' -count=1\n- go test ./lib/instances -run 'TestRotateSourceVsockForRestore|TestResolveForkTargetState' -count=1\n- go test ./lib/instances -run TestApplyForkTargetStateStoppedRefreshesSnapshotForkCID -count=1 -v\n- go test ./lib/instances -run TestFirecrackerWarmForkChain -count=1 -v\n- go test ./lib/instances -run 'TestQEMUWarmForkChain|TestCloudHypervisorWarmForkChain' -count=1 -v (local environment reached Docker Hub anonymous pull limits before guest creation)\n- go test ./lib/instances -run '^$' -count=1\n- go test ./lib/hypervisor/firecracker ./lib/instances -count=1 (local environment did not complete: bridge creation is not permitted here, local disk is below the reservation some tests need, and Docker Hub anonymous pulls hit rate limits)

### DIFF
--- a/lib/instances/fork_test.go
+++ b/lib/instances/fork_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/kernel/hypeman/lib/instances/phasetracking"
 	"github.com/kernel/hypeman/lib/paths"
 	snapshotstore "github.com/kernel/hypeman/lib/snapshot"
+	snapshottest "github.com/kernel/hypeman/lib/snapshot/testsupport"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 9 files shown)


lib/instances/fork_test.go
```diff
@@ -569,6 +570,117 @@ func TestForkCloudHypervisorFromRunningNetwork(t *testing.T) {
...
+		namePrefix: "ch",
+	})
+}
+
+type warmForkChainConfig struct {
+	hypervisor hypervisor.Type
+	namePrefix string
+}
+
+func runWarmForkChain(t *testing.T, mgr *manager, tmpDir string, cfg warmForkChainConfig) {
```
sjmiller609, line 629: This is testing the flow:

Fork from snapshot -> Fork again

rgarcia (approved):
approved — implementation direction looks good. a couple things worth checking:

- `lib/instances/fork.go:71` — skipping `rotateSourceVsockForRestore` for `targetState=Stopped` leaves the stopped fork with the same `VsockCID` as the restored source. `forkInstanceFromStoppedOrStandby` copies the standby source CID into the fork metadata, and the `StateStandby -> StateStopped` transition only deletes the snapshot, so starting that stopped fork later can collide with the still-running source CID. either keep rotating the source or assign the stopped fork a fresh CID before returning it.
- `lib/hypervisor/firecracker/fork.go:57` — consider handling non-`IsNotExist` `os.Stat` errors explicitly; right now permission/io errors are treated as “source exists”, which can set `retainAlias=false` and later clear alias metadata even though the source path may not actually be usable.

PR #251, opened 2026-05-30
Add restore network telemetry

sjmiller609:
## Summary
- add restore_network child spans around allocation, snapshot network patching, and recreate paths
- add network spans for allocation and TAP setup substeps on top of the optimized restore-network path
- preserve behavior; this PR is telemetry-only and now stacks on the host restore network optimization branch

## Tests
- go test ./lib/network -count=1
- go test ./lib/network ./lib/instances -run TestRequiresRestoreConfigDiskRefresh -count=1 *(lib/network passed; lib/instances fails in this checkout because embedded hypervisor and guest-agent binaries are missing)*


sjmiller609:
Closing in favor of #265, which recreates this telemetry PR on top of #264 from current origin/main. Keeping the old branch around for comparison.

PR #252, opened 2026-05-30
Optimize restore network setup

sjmiller609:
## Summary
- collapse network allocation checks onto one allocation scan for name, IP, and MAC selection
- replace bridge isolation shell-out with netlink
- reuse the TAP link returned by netlink instead of looking it up after creation

## Tests
- go test ./lib/network -count=1


sjmiller609:
Closing in favor of #264, which recreates this restore-network optimization from current origin/main. Keeping the old branch around for comparison.

PR #253, opened 2026-05-30
Add mailbox resume network handoff

sjmiller609:
## Summary
- add guest-agent ReconfigureNetwork RPC backed by in-process netlink, with host fallback to the existing shell reconfigure path for older guests
- add a VMGenID-triggered resume-network mailbox so fork restores can patch guest network config before resume and avoid host-initiated guest contact on the fast path
- re-include the public `wait_for_network` option for running fork results; it defaults to true and waits for a guest UDP applied ack, while `false` returns after resume with mailbox reconfigure continuing asynchronously
- document the runtime handoff behavior in `lib/forkvm/README.md`
- keep the change focused by omitting the arm RPC, polling/fetch server, vsock timing instrumentation, and perf harness from the earlier exploratory PRs

## Tests
- `go test ./lib/guest ./lib/system/guest_agent -count=1`
- `go test ./lib/oapi -count=1`
- `go test ./cmd/api/api -run 'TestForkInstance_Success|TestForkSnapshotMapsWaitForNetwork' -count=1` *(fails in this checkout because embedded hypervisor, guest-agent, and caddy binaries are missing)*
- `go test ./lib/instances -run 'TestGuestNetworkReconfigureConfig|TestGuestNetworkReconfigureCommand|TestPatchGuestResumeNetworkMailbox' -count=1` *(fails in this checkout because embedded hypervisor and guest-agent binaries are missing)*


sjmiller609:
Closing in favor of #260, which recreates the mailbox resume network handoff from current origin/main. Keeping the old branch around for comparison.

PR #259, opened 2026-06-01
Optimize UFFD pager cache concurrency

sjmiller609:
## Summary
- shard the UFFD pager page cache to reduce burst contention
- add an immutable borrowed cache-hit path for UFFDIO_COPY
- expose pager timing counters for cache lookup, backing reads, page faults, and copy latency
- add a regression test for aligned page offset distribution across shards

## Test Plan
- go test ./lib/uffdpager ./lib/hypervisor/firecracker
- remote browser burst benchmarks on deft-kernel-dev with UFFD enabled

### DIFF
--- a/lib/uffdpager/cache.go
+++ b/lib/uffdpager/cache.go
@@ -3,6 +3,8 @@ package uffdpager
 import (
 	"container/list"
 	"sync"
+	"sync/atomic"
+	"time"
 )
 
 type pageKey struct {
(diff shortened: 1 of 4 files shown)


sjmiller609:
Closing in favor of #262, which folds this UFFD pager cache concurrency work into the new UFFD pager PR. Keeping the old branch around for comparison.

PR #261, opened 2026-06-01
Optimize Firecracker snapshot resume

sjmiller609:
## Summary
- Uses Firecracker snapshot `resume_vm` support so restored guests can resume during `/snapshot/load` instead of requiring a separate post-load `Resume` call.
- Tracks whether a restored hypervisor has already resumed through `hypervisor.RestoredResumed`, and skips the extra resume step only in that case.
- Speeds up Firecracker API socket readiness detection with inotify on Linux, while keeping polling fallback and non-Linux polling behavior.
- Adds tests for resume-on-load config, restored-resumed propagation, socket readiness waiting, and tracing wrapper behavior.

## Tests
- `go test ./lib/hypervisor ./lib/hypervisor/firecracker ./lib/mailbox ./lib/guest ./lib/system/guest_agent ./lib/oapi -count=1`
- `git diff --check`


lib/hypervisor/firecracker/socket_wait_linux.go
```diff
@@ -0,0 +1,96 @@
...
+	"golang.org/x/sys/unix"
+)
+
+func waitForSocket(path string, timeout time.Duration) error {
+	if err := tryDialUnixSocket(path); err == nil {
+		return nil
+	}
+
+	parent := filepath.Dir(path)
+	fd, err := unix.InotifyInit1(unix.IN_CLOEXEC | unix.IN_NONBLOCK)
```
sjmiller609, line 20: we attempt to not wait via polling, then we fall back to polling

sjmiller609:
closing this one intentionally. resume-on-load saves a small Firecracker API step, but it complicates the restore invariant: when a temporary snapshot source alias is needed, `resume_vm: true` can let guest execution begin before alias cleanup. The child work does not fundamentally require this optimization, so we are dropping #261 from the production stack and rebasing the UFFD / fork-concurrency PRs directly onto main.

sjmiller609:
Closing intentionally; see latest comment. The dependent PRs are being rebased directly onto main without the resume-on-load layer.

PR #262, opened 2026-06-01
Add UFFD snapshot pager

sjmiller609:
## Summary
- adds a config-gated Firecracker UFFD memory backend for snapshot restore
- starts a per-restore UFFD pager session backed by the snapshot memory file and an optional shared page cache
- applies resume-network mailbox updates through UFFD overlay pages so restore does not mutate the backing memory file
- shards the pager cache and tracks pager timing counters for page-fault, lookup, backing-read, and copy latency

## Tests
- go test ./lib/hypervisor/firecracker ./lib/uffdpager ./lib/mailbox ./lib/guest ./lib/system/guest_agent ./lib/oapi -count=1

---

> [!NOTE]
> **High Risk**
> Changes Firecracker snapshot restore and host VM memory handling, depends on a correctly installed/versioned systemd pager, and misconfiguration or pager failure can leave restored VMs unhealthy until recycled.
> 
> **Overview**
> Adds an **opt-in Firecracker UFFD snapshot memory backend** (`hypervisor.firecracker_snapshot_memory_backend=uffd`, default `file`) with a bounded shared page cache (`hypervisor.firecracker_uffd_cache_max_bytes`).
> 
> On Linux, enabling UFFD starts a **versioned `hypeman-uffd-pager`** via the `hypeman-uffd@<version>.service` systemd template; Hypeman wires the Firecracker starter to create per-restore pager sessions and passes a **`mem_backend` of type `Uffd`** (per-session socket) instead of mmap’ing the snapshot `memory` file. Instance metadata tracks session/cache keys, restores go through new **`RestoreOptions`**, and stop/delete/standby/restore failure paths **close UFFD sessions**; state queries can mark instances unknown if the pager is unhealthy.
> 
> Also ships the pager in **build/release/install** (`cmd/uffd-pager`, Makefile, GoReleaser, install scripts) and adds CI **`scripts/check-uffd-version.sh`** so runtime pager changes require bumping `lib/uffdpager/VERSION`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fb4e46da9acde6a481fccb2e3f4f268a45d5bbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/api/config/config.go
```diff
@@ -404,9 +406,11 @@ func defaultConfig() *Config {
...
 
 		Hypervisor: HypervisorConfig{
-			Default:                       "cloud-hypervisor",
-			CloudHypervisorDefaultVersion: "",
-			FirecrackerBinaryPath:         "",
+			Default:                          "cloud-hypervisor",
+			CloudHypervisorDefaultVersion:    "",
+			FirecrackerBinaryPath:            "",
+			FirecrackerSnapshotMemoryBackend: "file",
+			FirecrackerUFFDCacheMaxBytes:     "4294967296",
```
sjmiller609, line 413: I'm thinking about 1 UFFD per snapshot, at 1 GB cache versus 4GB shared across all?

I'm also wondering about possibly having a limit in terms of the total sum of all UFFDs' memory, and a way to signal from hypeman to the UFFD to decrease its cache.. or is that too much? also should we be expiring old cache rather than only evicting after full?

cmd/api/config/config.go
```diff
@@ -618,6 +622,19 @@ func (c *Config) Validate() error {
 	if c.Hypervisor.Memory.KernelPageInitMode != "performance" && c.Hypervisor.Memory.KernelPageInitMode != "hardened" {
 		return fmt.Errorf("hypervisor.memory.kernel_page_init_mode must be one of {performance,hardened}, got %q", c.Hypervisor.Memory.KernelPageInitMode)
 	}
+	backend := strings.ToLower(strings.TrimSpace(c.Hypervisor.FirecrackerSnapshotMemoryBackend))
+	if backend == "" {
+		backend = "file"
+	}
+	switch backend {
+	case "file", "uffd":
+		c.Hypervisor.FirecrackerSnapshotMemoryBackend = backend
```
sjmiller609, line 631: defaults to file based, opt into uffd. what happens on existing server when opt in and restart hypeman?

lib/instances/firecracker_uffd.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/kernel/hypeman/lib/hypervisor/firecracker"
+	"github.com/kernel/hypeman/lib/logger"
+	"github.com/kernel/hypeman/lib/uffdpager"
+)
+
+func (m *manager) useFirecrackerUFFD(stored *StoredMetadata) bool {
+	return stored != nil &&
+		stored.HypervisorType == hypervisor.TypeFirecracker &&
+		m.firecrackerSnapshotMemoryBackend == uffdpager.BackendUFFD
+}
```
sjmiller609, line 23: this seems like a bunch of firecracker specific stuff in front of the hypervisor interface. why not live behind the interface?

lib/instances/guest_resume_network.go
```diff
@@ -164,50 +167,109 @@ func (m *manager) waitForGuestResumeNetworkUDPAck(ctx context.Context, waiter *g
 }
 
 func patchGuestResumeNetworkMailbox(snapshotDir, token string, payload *mailbox.Payload) error {
-	payloadBytes, err := mailbox.MarshalPayload(payload)
+	payloadBytes, idx, file, err := prepareGuestResumeNetworkMailbox(snapshotDir, token, payload, os.O_RDWR)
```
sjmiller609: explain why mailbox stuff changing in the context of uffd server change

lib/uffdpager/cache.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func mixPageHash(hash uint64) uint64 {
+	hash ^= hash >> 33
+	hash *= 0xff51afd7ed558ccd
+	hash ^= hash >> 33
+	hash *= 0xc4ceb9fe1a85ec53
+	hash ^= hash >> 33
+	return hash
+}
```
sjmiller609, line 238: explain why custom implementation

lib/uffdpager/server_linux.go
```diff
@@ -0,0 +1,740 @@
...
+	router.Post("/sessions/{id}/close", s.handleCloseSession)
+	router.Post("/drain", s.handleDrain)
+
+	s.httpServer = &http.Server{Handler: router}
+	err = s.httpServer.Serve(listener)
+	if errors.Is(err, http.ErrServerClosed) {
+		return nil
+	}
+	return err
+}
```
sjmiller609, line 121: currently we are locking VMs to a particular UFFD server that's running. I'm wondering if we could run this like a normal horizontally scalable service, in a way where when we have a version update, we could do a rollout of the UFFD to existing VMs. I wouldn't want updates to keep ending up with more and more cache dedicated total, and it's harder to reason about many concurrently running version.

can we do something like that?

I'm also curious about a benefit to doing UFFD per snapshot, as this seems like it would get a better hit rate, or better performance perhaps. but it would be simplest to just have 1 deployment running maybe? what do you think?

lib/uffdpager/server_linux.go
```diff
@@ -0,0 +1,740 @@
...
+	return uint64(b[0]) | uint64(b[1])<<8 | uint64(b[2])<<16 | uint64(b[3])<<24 |
+		uint64(b[4])<<32 | uint64(b[5])<<40 | uint64(b[6])<<48 | uint64(b[7])<<56
+}
+
+type bigEndian struct{}
+
+func (bigEndian) Uint64(b []byte) uint64 {
+	return uint64(b[7]) | uint64(b[6])<<8 | uint64(b[5])<<16 | uint64(b[4])<<24 |
+		uint64(b[3])<<32 | uint64(b[2])<<40 | uint64(b[1])<<48 | uint64(b[0])<<56
+}
```
sjmiller609: big single file, worth splitting up?

lib/uffdpager/supervisor_linux.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	return fmt.Errorf("uffd pager did not become healthy at %s", s.controlSocket)
+}
+
+func (s *Supervisor) isHealthy(ctx context.Context, versionKey string) bool {
+	var health HealthResponse
+	return s.doJSON(ctx, versionKey, http.MethodGet, "/health", nil, &health) == nil
+}
+
+func (s *Supervisor) drainOlderPagers(ctx context.Context) {
```
sjmiller609, line 179: I am wondering about possible more standard way to do deployment-like behavior with rolling deploys instead of the whole supervisor concept?

cmd/api/config/config.go
sjmiller609, line 413: ought we be introducing an external dependency such as a cache system like redis or no?

sjmiller609: Added `lib/uffdpager/README.md` to spell out the cache model. This is one local pager process with per-restore sessions and a shared bounded page cache keyed by snapshot cache key + page offset, not one independent cache per snapshot. I kept the default at file-backed restore and left the UFFD cache size configurable.

cmd/api/config/config.go
sjmiller609: Documented this in `lib/uffdpager/README.md`: the default remains `file`; opting into `uffd` affects future Firecracker snapshot restores only. Existing running VMs are unaffected, and sessions restored through UFFD stay pinned to their pager session until the instance lifecycle closes it.

lib/instances/firecracker_uffd.go
sjmiller609: Agreed. I moved the Firecracker snapshot-memory backend decision into the restore interface boundary: `RestoreVM` now takes generic restore options, and `firecracker.Starter.RestoreVM` owns creating the UFFD session and passing the UFFD socket to Firecracker. `instances` no longer writes Firecracker restore metadata directly.

lib/instances/guest_resume_network.go
sjmiller609: Added an inline explanation and `lib/uffdpager/README.md` coverage. With UFFD we should not mutate the shared backing snapshot memory file, so the resume-network mailbox patch is represented as a per-session overlay page served by the pager.

lib/uffdpager/cache.go
sjmiller609: Added README context for why this stays local/in-process: page faults are latency-sensitive and UFFD uses local sockets, so an external cache like Redis would add operational and latency risk. The implementation is intentionally a small bounded page cache keyed by snapshot + offset.

lib/uffdpager/server_linux.go
sjmiller609: For this PR I am leaving the current local versioned pager model in place and documenting the session/cache behavior. The end-game rollout model should be a follow-up, likely a node-local managed service with versioned drain, but I am intentionally not changing supervisor/rollout mechanics in this PR.

lib/uffdpager/server_linux.go
sjmiller609: Split the large Linux pager server file into entrypoint/types, HTTP handlers, session lifecycle, and page-fault/UFFD helpers. This is intended as a mechanical maintainability cleanup without behavior changes.

lib/uffdpager/supervisor_linux.go
sjmiller609: Agreed this is worth a follow-up, but I am not changing supervisor/rollout mechanics in this PR. The likely direction is a node-local managed service with versioned drain so active sessions stay pinned while new sessions move to the new version.

cmd/api/config/config.go
sjmiller609: I do not think we should introduce Redis or another external cache here. Added README context: UFFD page faults are local and latency-sensitive, and the pager keeps a bounded local page cache shared across sessions by snapshot cache key.

scripts/install.sh
```diff
@@ -586,6 +586,29 @@ ReadWritePaths=${DATA_DIR}
 
 [Install]
 WantedBy=multi-user.target
+EOF
+
+    $SUDO tee "${SYSTEMD_DIR}/${SERVICE_NAME}-uffd@.service" > /dev/null << EOF
```
sjmiller609, line 606: we install one systemd template: hypeman-uffd@.service. hypeman starts versioned instances like hypeman-uffd@0.1.0.service, where 0.1.0 comes from lib/uffdpager/VERSION. if hypeman changes but UFFD does not, it keeps using the same running pager. if UFFD changes, hypeman starts a new versioned pager and drains the old one: old sessions keep working, new sessions go to the new version, and the old pager exits once empty.

lib/uffdpager/README.md
```diff
@@ -0,0 +1,26 @@
+# UFFD Snapshot Pager
```
hiroTamada, line 1: nit: would be helpful to add a "control http api" section here — GET /health, GET /stats, POST /sessions, POST /sessions/{id}/close, POST /drain on {dataDir}/uffd/{version}/control.sock. firecracker uses a separate per-session unix socket (not http). routes live in server_linux.go but there's no single obvious place to discover them today.

lib/uffdpager/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+The UFFD pager lets Firecracker restore snapshot memory lazily. A restored VM
+gets a per-session UFFD socket, and page faults are served from the snapshot
+memory file by a local pager process.
+
+The pager is local to the host. It is not backed by Redis or another external
+cache because page faults are latency-sensitive and the kernel-facing UFFD
+socket is local. The process keeps one shared in-memory page cache, bounded by
+`hypervisor.firecracker_uffd_cache_max_bytes`. Cache entries are keyed by a
+snapshot cache key plus page offset, so multiple restore sessions from the same
```
hiroTamada: this reads like multiple vms restoring the same snapshot share cache, but the default cache_key hashes stored.Id (see firecrackerSnapshotCacheKey) so different instances won't share pages. might be worth tweaking the wording or the key formula depending on what you want.

lib/instances/restore.go
```diff
@@ -279,6 +286,7 @@ func (m *manager) restoreInstance(
 		log.ErrorContext(ctx, "failed to resume VM", "instance_id", id, "error", err)
 		// Cleanup on failure
 		hv.Shutdown(ctx)
+		m.closeFirecrackerUFFDSession(ctx, stored)
```
hiroTamada, line 289: restore/resume failures close the uffd session here, but the reconfigureGuestNetwork failure path a few lines below (~313) doesn't — worth calling closeFirecrackerUFFDSession there too before return (+1 on bugbot).

lib/instances/firecracker_uffd.go
```diff
@@ -0,0 +1,147 @@
...
+	if m.firecrackerUFFDPager == nil {
+		return fmt.Errorf("firecracker uffd session %s has no configured pager", stored.FirecrackerUFFDSessionID)
+	}
+	version := stored.FirecrackerUFFDPagerVersion
+	if version == "" {
+		version = m.firecrackerUFFDPager.VersionKey()
+	}
+	healthCtx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
+	defer cancel()
+	if _, err := m.firecrackerUFFDPager.HealthVersion(healthCtx, version); err != nil {
```
hiroTamada, line 112: this only checks pager /health for the version, not whether this instance's session still exists. after an unplanned pager restart, health can pass while the vm's uffd session is gone. fine for v1 but worth a readme/runbook note or a follow-up session-level check.

hiroTamada (approved):
reviewed the uffd pager slice — architecture looks solid (separate pager process, opt-in file backend, versioned systemd + drain). left a few nits on docs, restore cleanup, cache key wording, and pager vs session health. nice work.

lib/instances/fork.go
```diff
@@ -291,6 +291,11 @@ func (m *manager) forkInstanceFromStoppedOrStandby(ctx context.Context, id strin
 	forkMeta.ExitCode = nil
 	forkMeta.ExitMessage = ""
 	forkMeta.RestartStatus = restartpolicy.Status{}
+	forkMeta.FirecrackerUFFDSessionID = ""
+	forkMeta.FirecrackerUFFDPagerVersion = ""
+	if source.State != StateStandby {
+		forkMeta.FirecrackerSnapshotCacheKey = ""
+	}
```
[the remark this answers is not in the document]
sjmiller609: I think this is intentional and safe for this PR. For standby forks we first prepare the source snapshot memory, then CopyGuestDirectory creates the fork guest dir via reflink/sparse copy. The fork backing memory path may be a different inode/path, but at restore time it represents the same immutable snapshot bytes as the source generation. Keeping the propagated FirecrackerSnapshotCacheKey is what lets multiple standby forks share UFFD cached pages by snapshot generation.

The key would need to change once the fork produces a new standby snapshot or otherwise mutates its snapshot generation; that path refreshes/clears the Firecracker cache key. UFFD restore itself reads from the backing file and does not mutate it, so shared cache entries for the same snapshot generation should not serve divergent bytes here.

PR #263, opened 2026-06-01
Improve Firecracker fork concurrency

sjmiller609:
## Summary
- rewrites Firecracker snapshot-state data-dir paths for forks when an equal-length replacement is possible, including CRC recomputation
- avoids the temporary source-dir alias on restore when the snapshot state can be rewritten directly
- narrows alias locking to only the restore windows that still need a temporary source-dir alias
- gates concurrent stopped/standby fork preparation behind `SupportsConcurrentForkPrepare`, with Firecracker opting in and other hypervisors keeping the previous exclusive source lock behavior

## Tests
- go test ./lib/hypervisor/firecracker ./lib/hypervisor ./lib/mailbox ./lib/guest ./lib/system/guest_agent ./lib/oapi -count=1
- go test ./lib/instances -run 'TestForkInstanceStoppedSourceUsesReadLock|TestForkInstanceStoppedSourceRequiresWriteLockWithoutConcurrentPrepare' -count=1\n\n## Notes\n- the targeted `./lib/instances` tests required local placeholder files for embedded test binaries that are not present in this checkout.

---

> [!NOTE]
> **Medium Risk**
> Changes fork, restore, and snapshot locking paths plus shared restore admission; behavior is well covered by tests but mistakes could cause rare restore/fork races or overload.
> 
> **Overview**
> This PR improves **Firecracker fork/restore concurrency** and surfaces **resource exhaustion** more clearly on fork/restore APIs.
> 
> **Firecracker fork prepare** can now **rewrite snapshot state file paths** (with CRC64 validation) when source and target data-dir strings are the same length, so many forks avoid the temporary **source data-dir symlink alias** during restore. When rewrite is not possible, prepare reports **`RequiresSnapshotSourceAlias`** and restore still uses a **narrow, shared alias lock** (readers block on metadata/copy; writers only during the alias mutation window).
> 
> **Instance fork** uses a **source read lock** for stopped/standby sources when the hypervisor advertises **`SupportsConcurrentForkPrepare`** (Firecracker enables it); other hypervisors keep the exclusive source lock. Fork metadata admission is **serialized** with a final name check before save, and **snapshot memory prepare** is **per snapshot directory**.
> 
> **Restore capacity** is capped via configurable **`firecracker_max_concurrent_restores`** (default 32), with a **shared slot pool** across managers for that hypervisor/limit.
> 
> **API** maps **`ErrInsufficientResources`** to **409** with code **`insufficient_resources`** on restore, fork instance, and fork snapshot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb2602ee05984574758321f8967086fe515a7c38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/hypervisor/firecracker/snapshot_state.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+const firecrackerSnapshotCRC64JonesReversedPoly uint64 = 0x95ac9329ac4bc9b5
+
+type snapshotStatePathRewrite struct {
+	Source string
+	Target string
+}
+
+func rewriteSnapshotStatePathsForFork(statePath string, rewrites []snapshotStatePathRewrite) (bool, error) {
```
sjmiller609, line 18: > the snapshot rewrite is needed because Firecracker snapshot state can contain absolute paths to the source instance’s data dir.
> 
> when a fork restores from that snapshot, Firecracker may still try to open the old source paths. without rewriting those paths, we have to temporarily make the source path point at the fork path with a symlink alias during `/snapshot/load`.
> 
> that alias is process-global dangerous state, so we serialize it:
> 
> ```text
> fork A: alias source -> fork A, /snapshot/load, undo alias
> fork B: waits
> fork C: waits
> ```
> 
> the rewrite avoids that by changing the path inside the snapshot state up front:
> 
> ```text
> snapshot points directly at fork A path
> no alias needed
> no global restore lock
> ```
> 
> ways to avoid the rewrite without serializing forks:
> 
> 1. **per-source alias locking**  
>    Better than global, but forks from the same source still serialize, which is the common burst case.
> 
> 2. **unique stable path aliases per fork before snapshot creation**  
>    Maybe possible, but it means changing how snapshots are created so Firecracker records fork-safe paths from the beginning. Bigger design change.
> 
> 3. **copy/mount layout where old source path is valid per fork namespace**  
>    Would need mount namespace / bind mount isolation around Firecracker. Also bigger and more operationally sharp.
> 
> 4. **ask Firecracker for path remapping support**  
>    Ideal, but not available today.
> 
> so for this PR, snapshot rewrite is the smallest way to remove the alias lock from the hot restore path.

lib/hypervisor/firecracker/snapshot_state.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const firecrackerSnapshotCRC64JonesReversedPoly uint64 = 0x95ac9329ac4bc9b5
```
sjmiller609, line 11: I was trying to find if we should avoid rewriting this, but it's actually not that bad. This file is rewriting the firecracker snapshot metadata file, in order to rewrite which paths are being referenced, which allows for better concurrency (see below comment in this same file). When we re-write it, we have to recalculate the checksum.

lib/hypervisor/firecracker/snapshot_state_fork.go
```diff
@@ -0,0 +1,91 @@
+package firecracker
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func prepareSnapshotDataDirForFork(snapshotConfigPath string, meta *restoreMetadata, sourceDir, targetDir string) (bool, error) {
```
sjmiller609, line 9: keep in mind that all fork tests would fail if this didn't work. So this is kind of a hard fail or works situation, it seems to me

lib/hypervisor/hypervisor.go
```diff
@@ -234,6 +234,10 @@ type Capabilities struct {
 	// on-disk base across restore/standby cycles.
 	SupportsSnapshotBaseReuse bool
 
+	// SupportsConcurrentForkPrepare indicates stopped/standby forks can prepare
+	// separate target snapshots concurrently from the same source.
+	SupportsConcurrentForkPrepare bool
```
sjmiller609, line 256: I didn't want to assume that all hypervisors will work with concurrently running forks, and I'm only focused on firecracker right now.

config.example.yaml
```diff
@@ -26,6 +26,8 @@ data_dir: /var/lib/hypeman
 #   default: cloud-hypervisor
 #   # Optional: use a custom Firecracker binary path instead of the embedded one.
 #   # firecracker_binary_path: /usr/local/bin/firecracker
+#   # Caps concurrent Firecracker snapshot restore/resume work to avoid host overload.
+#   # firecracker_max_concurrent_restores: 32
```
sjmiller609, line 30: I actually tested this as working up to 200 so far, but we can leave the default down for now

lib/instances/manager.go
```diff
@@ -81,17 +81,33 @@ type ResourceLimits struct {
 
 // ManagerConfig holds non-resource manager behavior settings.
 type ManagerConfig struct {
-	LifecycleEventBufferSize int
+	LifecycleEventBufferSize         int
+	FirecrackerMaxConcurrentRestores int
 }
 
+const defaultManagerFirecrackerMaxConcurrentRestores = 4
```
sjmiller609: the default config is 32

rgarcia (commented):
Did a read-through focused on the fork/restore concurrency changes. Overall the capability-gated approach is clean and the snapshot-state path rewrite is a nice way to avoid the symlink dance. Three things I'd like to discuss before approving — none are blockers on their own, but #1 and #2 are about long-term operability.

**1. The symlink-alias fallback is load-bearing but invisible, and gives us two failure modes to reason about.**

For the common instance→instance fork the rewrite is guaranteed to work (IDs are fixed-length `cuid2`, so `guests/<id>` source and target are always equal length). The fallback in `recordSnapshotSourceAliasFallback` really only matters for (a) symlink-resolved paths from `EvalSymlinks` and (b) diff-snapshot chains that retain an upstream base path (`RetainSnapshotSourceDataDirAlias`), where the alias is genuinely required for correctness.

My concern is operational: the alias path is slower and globally serialized, and right now it's entered *silently*. During an incident I don't want to have to reason about whether a host is hitting the rewrite path or the alias path. Two asks:
- Emit a metric/log counter whenever the alias fallback fires, so "are we on the slow path?" is a one-query answer.
- Consider making the non-rewritable case a hard error *unless* it's the legitimate retained-base diff-snapshot case. That way "alias path" only ever means the one well-understood scenario, instead of a catch-all.

If there's a path to giving diff-snapshot base reuse a stable fixed-length canonical path, we could eventually drop the alias entirely — but I don't think that needs to happen in this PR.

**2. Firecracker-specific concept leaking into the generic instances manager.**

This PR introduces the clean pattern for hypervisor-specific behavior — a capability plus `supportsConcurrentForkPrepare(hvType)` — and then doesn't use it for the restore limiter. `acquireFirecrackerRestoreSlot` hard-codes `stored.HypervisorType != hypervisor.TypeFirecracker`, and we now have a `firecrackerRestoreSlots` field on the generic `manager`, a `firecrackerRestoreSlotPools` package global, and `FirecrackerMaxConcurrentRestores` in the manager config. To stay consistent with the capability approach a few lines up, I'd rather see the concurrency limit advertised by the hypervisor (e.g. a `MaxConcurrentRestores` on `Capabilities` or typed config keyed by `hypervisor.Type`) and the manager keep a generic `map[hypervisor.Type]chan struct{}`, so `manager.go`/`restore.go` never name Firecracker directly. To be clear, the *existence* of a restore limiter is fine and probably inherent to Firecracker's restore cost — it's the naming and the hard-coded type switch in the generic layer I'd like to clean up.

**3. `ForkInstance`'s read-lock logic needs explanation.**

The `useReadLock` block is a double-checked-locking pattern (load metadata under `RLock`, drop it, re-acquire and re-load to re-validate) and its correctness isn't self-evident. The escalation condition (`useReadLock && sourceState == StateStandby && targetState == StateRunning && targetRestoreNeedsSourceLock`) is also a dense boolean. Please add a comment explaining why we re-validate under the second lock and what each escalation condition guards against — this is exactly the kind of concurrency code the next person will be afraid to touch without context.

Also +1 on getting the Bugbot duplicate-name finding resolved (concurrent same-name forks both passing `instanceNameExists`/`NameExists` under the read lock) before merge, unless the serialized name-admission already covers it — if so, a comment pointing at where that's enforced would help.

lib/instances/fork.go
```diff
@@ -259,17 +258,8 @@ func (m *manager) forkInstanceFromStoppedOrStandby(ctx context.Context, id strin
...
-			return nil, fmt.Errorf("prepare standby snapshot for fork: %w", err)
-		}
-	}
-
-	if err := forkvm.CopyGuestDirectory(srcDir, dstDir); err != nil {
-		if errors.Is(err, forkvm.ErrSparseCopyUnsupported) {
-			return nil, fmt.Errorf("fork requires sparse-capable filesystem (SEEK_DATA/SEEK_HOLE unsupported): %w", err)
-		}
-		return nil, fmt.Errorf("clone guest directory: %w", err)
+	if err := m.copyForkSourceGuestDirectory(ctx, source.State, id, stored, srcDir, dstDir); err != nil {
```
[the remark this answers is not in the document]
sjmiller609: Addressed in 32b62db5. The early instance/network name checks are only advisory; final fork metadata admission is serialized through saveForkMetadata under forkMetadataMu, and that function now has an explicit comment documenting that concurrent forks re-check names immediately before saving.

rgarcia (approved):
Approving — the review feedback is well addressed:

- **Hypervisor leakage**: clean refactor. `acquireRestoreSlot(ctx, hvType)` + `restoreSlotsByHypervisor map[hypervisor.Type]chan struct{}` + `MaxConcurrentRestoresByHypervisor`, with the only Firecracker-specific wiring now living in `providers.go`. Exactly the capability-consistent shape I was hoping for. (Tiny nit, non-blocking: `defaultManagerFirecrackerMaxConcurrentRestores` is now applied as the generic per-hypervisor default, so the name reads a little oddly.)
- **Alias fallback visibility**: the `WarnContext` lines in both fork paths are enough to answer "are we on the slow path?" during an incident. A counter metric would be a nice future add but I don't consider it blocking.
- **`useReadLock` docs**: the three added comments make the double-checked-lock + escalation logic readable. Thank you.
- **Duplicate fork names**: agreed that `saveForkMetadata` under `forkMetadataMu` with the re-check before save is the right serialized admission point.

Two open Bugbot threads still look valid to me and don't appear to be touched by 32b62db. Approving so as not to block, but I'd like these resolved (or explicitly dispositioned) before merge:

1. **VsockCID collision on concurrent standby forks.** `fork.go` still assigns standby forks the source's CID (`forkMeta.VsockCID = stored.VsockCID`) while stopped forks get a fresh `generateVsockCID`. With the new read-lock concurrency, multiple standby forks of one source can be prepared in parallel and then started to running, all sharing the source CID → vsock routing / guest-agent breakage. The comment says rewriting CID in restored snapshots isn't reliable across hypervisors, which I get — but then we likely need to either reallocate the CID at restore time for forks or serialize "running" admission per source CID. Can you confirm what prevents two running guests with the same CID here?

2. **Stale source state in the escalated write-lock restore.** `applyForkTargetStateWithSourceLock` re-takes the source write lock, but the `targetRestoreNeedsSourceLock` / `sourceState == StateStandby` decision was captured under an earlier RLock that was already released. Nothing re-validates the source state under the write lock before the aliased restore runs. If the source was restored to running in that gap, we could alias its data dir while its VMM is live. Does holding the source write lock actually guarantee the source isn't running by this point, or should we re-load + re-check source state under the write lock before aliasing?

Neither is a correctness concern for the common (rewrite-succeeds, single-fork) path, so I'm comfortable approving — but I'd treat #2 as the higher-risk of the two given it's the data-dir-corruption path.

PR #267, opened 2026-06-01
Add fork mailbox payload API

sjmiller609:
## Summary
- add optional named JSON mailbox payloads to instance and snapshot fork requests
- patch matching mailbox markers in standby snapshot memory before the fork resumes
- optionally wait for guest UDP acknowledgements after resume when a mailbox requests it
- share mailbox marker/layout helpers through lib/mailbox so host and guest-side integrations use the same contract
- preserve existing fork network readiness behavior: running forks wait for network readiness, with resume-network fallback paths unchanged

## Tests
- make oapi-generate
- git diff --check origin/main --
- go test ./lib/mailbox ./lib/guest ./lib/system/guest_agent ./lib/oapi -count=1
- go test ./cmd/api/api ./lib/instances -run 'Test(ForkInstanceSuccess|ForkSnapshotSuccess|PatchForkMailbox|ForkMailboxPayloadWithAckPort|ForkReturnReadinessDoesNotUseMailboxEligibility)' -count=1 (with local placeholder embedded binaries)\n\nFull cmd/api/api and lib/instances package test runs still require embedded hypervisor, guest-agent, and caddy binaries in this checkout.

---

> [!NOTE]
> **High Risk**
> Changes fork/restore orchestration and direct snapshot memory patching; incorrect markers or ack handling can fail forks or leave guests misconfigured, and the feature is limited to Firecracker standby→Running paths.
> 
> **Overview**
> Adds optional **named JSON mailboxes** to instance and snapshot fork APIs so callers can inject per-fork configuration into guest memory before a forked standby VM resumes.
> 
> Fork requests accept a `mailboxes` array (name, guest-provided token, JSON payload, optional `wait_for_ack` / `ack_timeout_ms`). The API layer maps these into domain types; **Firecracker standby forks** that target **Running** validate payloads (limits, duplicate names, JSON object shape) and patch matching markers in snapshot **memory** before restore. When `wait_for_ack` is set, the host injects `ack_port` and fails restore if the guest does not send a UDP `stage=applied` ack for that mailbox in time.
> 
> **`lib/mailbox`** gains fork-specific layout/marker helpers and shared `FindMarker` / `WritePayloadAt` used by resume-network and fork paths. Restore runs **resume-network** and **fork-mailbox** handoffs through a common post-resume hook; UDP ack parsing is stricter (field-based matching).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f763c6ef1bb8405e5c59aa3142886e4cbcff940c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/instances/fork_mailbox.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func validateForkMailboxes(mailboxes []ForkMailboxPayload) error {
+	if len(mailboxes) > 16 {
+		return fmt.Errorf("%w: at most 16 mailboxes can be patched for a fork", ErrInvalidRequest)
+	}
+	seen := make(map[string]struct{}, len(mailboxes))
+	for _, mailbox := range mailboxes {
+		name := strings.TrimSpace(mailbox.Name)
+		if !mailboxpkg.ValidForkMailboxName(name) {
```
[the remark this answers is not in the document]
sjmiller609: fixed in 3107262 by validating the exact mailbox name that will be used for patching, instead of trimming during validation. added a regression test for padded mailbox names.

rgarcia (commented):
blocking on a few maintainability/design issues:

- `lib/instances/fork.go:50` and `lib/instances/snapshot.go:390` only gate mailbox usage on standby + running target, but `lib/instances/fork_mailbox.go:225` assumes a Firecracker-style raw `memory` snapshot file and fixed mailbox offsets. Non-Firecracker standby forks can accept mailbox requests and fail later inside restore. Please make this an explicit capability/hypervisor check before fork work begins, so the API rejects unsupported mailbox forks at the boundary.

- `lib/instances/fork_mailbox.go:209` duplicates the open/stat/mmap/find/write/cache logic already in `lib/instances/guest_resume_network.go:186`. This is the same mailbox-frame operation with a different layout. Please push the common primitive into `lib/mailbox` with layout config, instead of maintaining two snapshot-memory patchers that can drift. Ideally the fork path should also preflight all requested markers before writing any payloads, so multi-mailbox patching is structurally all-or-nothing rather than relying on outer fork cleanup.

- `lib/instances/guest_resume_network.go:143` adds mailbox ACK matching as raw substring checks. `mailbox=foo` can match `mailbox=foobar`, and any free-form UDP text containing `stage=applied` is accepted. Since this ACK now gates successful fork restore, please parse the key/value payload and require exact fields for `stage` and `mailbox`.

openapi.yaml
```diff
@@ -690,6 +690,48 @@ components:
...
+          description: Guest mailbox name.
+          pattern: ^[A-Za-z0-9._:-]+$
+          minLength: 1
+          maxLength: 64
+          example: kernel.identity.v1
+        token:
+          type: string
+          description: Per-template mailbox token used to identify the guest memory marker.
+          minLength: 1
+          maxLength: 128
```
rgarcia, line 719: Discoverability gap — there's no way for a fork caller to enumerate which mailbox names/tokens exist in a given snapshot. A wrong/typo'd token just surfaces as a generic "marker not found"-style failure at fork time. Feels strange for the token to be an out-of-band contract that assumes knowledge of how the guest sets this token?

openapi.yaml
```diff
@@ -875,6 +917,15 @@ components:
             Optional hypervisor override. Allowed only when forking from a Stopped snapshot.
             Standby snapshots must fork with their original hypervisor.
           example: cloud-hypervisor
+        mailboxes:
+          type: array
+          description: |
+            Optional JSON mailbox payloads to patch into a standby snapshot before resuming the fork.
+            Each mailbox must correspond to a guest-side mailbox marker that was present when the
+            source snapshot was captured. Mailboxes are only supported for forks that restore from a
+            standby snapshot into Running state.
```
rgarcia: a reader could reasonably expect a snapshot-side API given the phrasing here. Consider adding a sentence like "The marker is written by guest software before the standby snapshot is captured; Hypeman does not create or return these tokens" if we're doing out-of-band contract

rgarcia (commented):
1. Suggestion, maybe too early: Unify the two handoffs into one interface + slice (main ask).
After this PR, restoreInstance prepares two handoffs back-to-back (restore.go:263 and :271) that now implement the identical trio — prepare… / Close() / AfterResume(ctx) error — and the two post-resume teardown blocks (restore.go:332-341 and :343-349) are line-for-line the same recovery (hv.Shutdown → rollbackAdmissionAllocationActive → releaseNetwork → wrapped return). That's the setup for:

```
type resumeHandoff interface {
    AfterResume(ctx context.Context) error
    Close()
}
handoffs := []resumeHandoff{resumeNetworkHandoff, forkMailboxHandoff}
for _, h := range handoffs { defer h.Close() }
// after resume, in order:
for _, h := range handoffs {
    if err := h.AfterResume(ctx); err != nil { /* one shared teardown */ }
}
```
Order is preserved (network up before the guest can UDP-ack), so the slice is safe. Collapses the duplicated teardown and makes a third handoff one line. No guest changes needed.

2. Light suggestion: Collapse the duplicated UDP-wait machinery.
WaitApplied and WaitMailboxApplied (guest_resume_network.go:119 and :140) are methods on the same waiter, reading the same channel, with identical select loops — they differ only in the match predicate. Suggest one wait(ctx, match func(fields map[string]string) bool) and route both through parseUDPAckFields. That also removes the case-handling mismatch between them (one lowercases the whole string, the other only the keys).

3. WriteForkMailboxPayloadAt is a thin wrapper used only by its own test.
mailbox.go:110 just calls WritePayloadAt(w, ForkLayout, …); production calls WritePayloadAt directly (fork_mailbox.go:272), and the only caller of the wrapper is TestWriteForkMailboxPayloadAt. Suggest deleting it and pointing the test at WritePayloadAt.

4. AckTimeout validation is asymmetric + message is off.
fork_mailbox.go:99-104: the < 0 check is gated on WaitForAck but the > 30s check isn't, and "must be positive" is misleading since 0 is valid (maps to the 2s default). Suggest aligning the gating and changing the message to "must not be negative."

5. Question on wait_for_ack failure semantics.
On a missing ack, a fork that resumed successfully gets torn down (restore.go:343). Two asymmetries vs the resume-network path worth confirming are intentional: that path falls back to host-initiated reconfigure (resume_network_handoff.go:81) rather than aborting, and its default timeout is 5s vs 2s here. Since the guest-side consumer of a fork mailbox is caller-implemented, wait_for_ack=true against a guest that doesn't ack will reliably destroy the fork after 2s — fine if that's the intended contract, but worth a note on the API field and a deliberate choice on the 2s vs 5s default.

openapi.yaml
sjmiller609: documented the out-of-band contract in `f763c6e`: guest software writes the mailbox marker/token before standby capture, and Hypeman does not create, return, or enumerate those tokens.

I am intentionally not adding a discovery API in this PR. The snapshot-memory patching path is expected to evolve with the UFFD pager work, and if that introduces a natural metadata registry for patch targets we can expose discovery there. For #267, the API contract is explicit that mailbox names/tokens are coordinated with guest-side code.

openapi.yaml
sjmiller609: added that sentence in `f763c6e` to both fork mailbox request descriptions, and mirrored the same behavior-level wording in `lib/mailbox/README.md`.

lib/instances/restore.go
```diff
@@ -248,13 +265,23 @@ func (m *manager) restoreInstance(
...
 		releaseNetwork()
 		return nil, fmt.Errorf("prepare guest resume network handoff: %w", err)
 	}
-	defer resumeNetworkHandoff.Close()
+	handoffs = append(handoffs, resumeNetworkHandoff)
+
+	forkMailboxHandoff, err := m.prepareForkMailboxHandoff(ctx, stored, snapshotDir, opts.Mailboxes)
+	if err != nil {
+		releaseNetwork()
+		return nil, fmt.Errorf("prepare fork mailbox handoff: %w", err)
```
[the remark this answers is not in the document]
sjmiller609: fixed in `8145ab0`: restore handoff cleanup is now registered immediately after the handoff slice is created, so an already-prepared resume-network handoff is closed if later fork mailbox preparation fails.

sjmiller609:
Closing this for now because we are dropping the mailbox direction. Leaving the branch intact for reference.

PR #270, opened 2026-06-03
Use UFFD only for initial snapshot fork restores

sjmiller609:
## Summary
- add a one-shot Firecracker UFFD restore metadata flag
- arm UFFD only for the first restore of a new fork from standby snapshot state
- clear the flag after restore/standby/stop so later direct resumes use file-backed restore

## Validation
- go test ./lib/instances ./lib/hypervisor/firecracker ./lib/uffdpager
- validated on deft with local UFFD + Firecracker concurrency merge:
  - single browser fork -> CDP, standby, file-backed restore -> CDP succeeds
  - 25-way Running fork burst: 25/25 succeeded; all requests started together; no source-state/vsock copy races

Note: the runtime validation used the Firecracker concurrency branch merged locally, since that branch removes the source alias/fork race this behavior relies on for burst fanout.

---

> [!NOTE]
> **High Risk**
> Changes Firecracker snapshot memory, fork fanout, and UFFD/systemd integration on a critical VM lifecycle path; incorrect backing or locking could affect restore correctness or concurrent forks.
> 
> **Overview**
> This PR limits **Firecracker UFFD** to a **one-shot** path on **standby snapshot fanout**, instead of using UFFD on every restore when the backend is configured for UFFD.
> 
> New forks from standby (instance or snapshot) can **skip copying** `snapshot-latest/memory`, record **deferred backing** and **`FirecrackerUseUFFDOnNextRestore`**, and restore once via the pager against shared source memory. **Standby** then **materializes** local `memory` via `SnapshotOptions.DeferredMemoryBackingPath` before diff snapshot; later restores use **file-backed** memory. Metadata and lifecycle paths clear the one-shot flag and UFFD session state after restore, standby, and stop.
> 
> **Fork/copy** gains optional path skips, `CopyRegularFile`, and stricter skipping of stale `*.sock` artifacts. **Restore** accepts `SnapshotMemoryBackingPath`; the UFFD supervisor moves env to `/run/hypeman/uffd` with CI-scoped systemd instance names. **CI/Makefile** install the pager unit template, pass pager binary env vars, and add an integration test for the full UFFD lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c74c539c6edce656456360cbfb1cde48b80470d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/test.yml
```diff
@@ -95,6 +95,29 @@ jobs:
       - name: Build
         run: make build
 
+      - name: Install UFFD pager systemd template
```
sjmiller609, line 98: basically this is ci setup so e2e can run UFFD based testing

lib/forkvm/copy.go
```diff
@@ -33,12 +33,22 @@ type copyState struct {
...
+	SkipRelativePaths map[string]struct{}
+}
+
 // CopyGuestDirectory recursively copies a guest directory to a new destination.
 // Regular files are cloned via reflink (FICLONE) when the underlying filesystem
 // supports it; otherwise we fall back to a sparse extent copy
 // (SEEK_DATA/SEEK_HOLE). Runtime sockets and logs are skipped because they are
 // host-runtime artifacts.
 func CopyGuestDirectory(srcDir, dstDir string) error {
+	return CopyGuestDirectoryWithOptions(srcDir, dstDir, CopyOptions{})
```
sjmiller609, line 47: this empty wrapper makes sense to avoid refactoring everywhere calling without copy options

lib/hypervisor/firecracker/firecracker.go
```diff
@@ -112,17 +113,76 @@ func (f *Firecracker) Resume(ctx context.Context) error {
...
+		return err
+	}
 	params := toSnapshotCreateParams(destPath)
 	if _, err := f.do(ctx, http.MethodPut, "/snapshot/create", params, http.StatusNoContent); err != nil {
 		return fmt.Errorf("create snapshot: %w", err)
 	}
 	return nil
 }
 
+func materializeDeferredSnapshotMemory(destPath, sourcePath string) error {
```
sjmiller609, line 130: One of the main changes: we do the reflink copy of the snapshot into the guest directory, only just before doing the standby. this is because if we do it during fanout, then bursts are serializing on the  reflink copy.

lib/instances/firecracker_test.go
```diff
@@ -648,6 +664,195 @@ func TestFirecrackerWarmForkChain(t *testing.T) {
 	require.NoError(t, mgr.DeleteSnapshot(ctx, snapshot.Id))
 }
 
+func TestFCUFFDOneShotLifecycle(t *testing.T) {
```
sjmiller609, line 667: One of the main changes: adding a UFFD lifecycle test

sjmiller609: > `TestFCUFFDOneShotLifecycle` now verifies the full one-shot UFFD lifecycle plus guest state preservation.
> 
> Flow:
> 
> 1. **Create source VM**
>    - Starts Firecracker with `sleep infinity`.
>    - Waits until guest exec works.
>    - Writes markers:
>      - disk: `/root/uffd-lifecycle/source`
>      - memory/tmpfs: `/dev/shm/uffd-lifecycle/source`
> 
> 2. **Create standby snapshot resource from source**
>    - Confirms snapshot memory file exists.
>    - Restores source and confirms both source markers still exist.
> 
> 3. **Fork snapshot into running parent**
>    - Parent is restored through the one-shot UFFD path.
>    - Confirms source disk and memory markers are present.
>    - Writes new parent markers after UFFD restore:
>      - disk: `/root/uffd-lifecycle/parent-after-uffd`
>      - memory/tmpfs: `/dev/shm/uffd-lifecycle/parent-after-uffd`
>    - Confirms parent has no local `snapshot-latest/memory` yet, because memory copy is deferred.
> 
> 4. **Standby parent, then resume parent file-backed**
>    - Standby materializes a normal snapshot memory file.
>    - Metadata clears UFFD one-shot/session/deferred path.
>    - File-backed resume succeeds.
>    - Confirms both source markers and parent-after-UFFD markers still exist, including the tmpfs memory marker.
> 
> 5. **Fork from running parent into running child**
>    - Forking from running parent first standbys parent, then restores parent and child.
>    - Confirms child inherits source and parent markers.
>    - Writes child-only disk and tmpfs markers.
>    - Confirms parent does **not** see child-only markers.
> 
> 6. **Standby child, then resume child file-backed**
>    - Confirms child’s disk and tmpfs markers survive the child’s own standby/resume.
>    - Then standbys child again and cleans everything up.
> 
> The key thing it proves: data written after UFFD restore, including tmpfs-backed memory state, survives standby and later file-backed resume. That directly guards against accidentally reverting to the original base memory.

lib/instances/stop.go
```diff
@@ -307,6 +307,8 @@ func (m *manager) stopInstance(
 	stored.ProgramStartedAt = nil
 	stored.GuestAgentReadyAt = nil
 	stored.FirecrackerSnapshotCacheKey = ""
+	stored.FirecrackerUseUFFDOnNextRestore = false
+	stored.FirecrackerDeferredSnapshotMemoryPath = ""
```
hiroTamada: nit: the one-shot uffd metadata gets cleared in ~6 spots now (stop, standby, restore, snapshot restore, and two places in fork), and each clears a slightly different subset of {FirecrackerSnapshotCacheKey, FirecrackerUseUFFDOnNextRestore, FirecrackerUFFDSessionID, FirecrackerUFFDPagerVersion, FirecrackerDeferredSnapshotMemoryPath}. given the snapshot-isolation blast radius here, consider a single `clearFirecrackerUFFDOneShotState(stored)` helper that resets all of them, and have each lifecycle site call it (plus any path-specific extras). keeps a future lifecycle path from silently forgetting a field.

lib/hypervisor/firecracker/firecracker.go
```diff
@@ -112,17 +113,76 @@ func (f *Firecracker) Resume(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	if err := forkvm.CopyRegularFile(resolvedSourcePath, targetPath); err != nil {
+		return fmt.Errorf("materialize deferred snapshot memory: %w", err)
+	}
+	return nil
+}
+
+func resolveDeferredSnapshotMemorySourcePath(sourcePath string) (string, error) {
```
hiroTamada, line 151: nit: `resolveDeferredSnapshotMemorySourcePath` + `alternateRetainedSnapshotMemoryPath` here are near-identical twins of `resolveFirecrackerSnapshotMemoryBackingPath` + `alternateFirecrackerRetainedSnapshotMemoryPath` in lib/instances/firecracker_uffd.go (same latest↔base swap + stat fallback). presumably duplicated to avoid an instances→firecracker import cycle, which is fair — but two copies of subtle path-resolution logic will drift. consider a shared helper in a neutral leaf package (alongside forkvm, or a small snapshotpaths util), or at minimum a cross-reference comment noting they must stay in sync.

scripts/install.sh
```diff
@@ -612,8 +612,12 @@ After=network.target
 [Service]
 Type=simple
 Environment="HOME=${DATA_DIR}"
-EnvironmentFile=${DATA_DIR}/uffd/%i/pager.env
-ExecStart=${INSTALL_DIR}/${UFFD_PAGER_BINARY_NAME} --data-dir ${DATA_DIR} --version-key %i --cache-max-bytes \${HYPEMAN_UFFD_CACHE_MAX_BYTES}
+Environment="HYPEMAN_UFFD_BINARY=${INSTALL_DIR}/${UFFD_PAGER_BINARY_NAME}"
+Environment="HYPEMAN_UFFD_DATA_DIR=${DATA_DIR}"
+Environment="HYPEMAN_UFFD_VERSION_KEY=%i"
+Environment="HYPEMAN_UFFD_CACHE_MAX_BYTES=4294967296"
+EnvironmentFile=-/run/hypeman/uffd/%i.env
```
hiroTamada, line 619: deploy-ordering question: the old unit used a non-optional `EnvironmentFile=${DATA_DIR}/uffd/%i/pager.env` and the supervisor wrote there; the new supervisor writes to `/run/hypeman/uffd/%i.env` instead. so a host that upgrades the hypeman binary (new supervisor → writes /run/...) but still has the *old* template installed would fail to start the unit (old non-optional EnvironmentFile path no longer produced). this requires install.sh (template refresh + daemon-reload) to land together with / before the binary upgrade. can you confirm the rollout always runs install.sh in that order?

lib/instances/manager.go
```diff
@@ -143,6 +143,7 @@ type manager struct {
 	resourceValidator         ResourceValidator // Optional validator for aggregate resource limits
 	instanceLocks             sync.Map          // map[string]*sync.RWMutex - per-instance locks
 	forkMetadataMu            sync.Mutex
+	snapshotSourceLocks       sync.Map      // map[string]*sync.Mutex - Firecracker snapshot source alias locks
```
hiroTamada: nit: keys here are added via LoadOrStore and never deleted, so the map grows unbounded over the host's lifetime. each entry is a tiny *sync.Mutex and distinct guest dirs are bounded by instances-ever-created, so probably fine — but on a long-lived high-churn host it's a slow leak. worth a comment acknowledging it (or opportunistic cleanup on instance delete).

lib/instances/fork.go
```diff
@@ -478,6 +536,10 @@ func (m *manager) applyForkTargetState(ctx context.Context, forkID string, targe
 				return nil, fmt.Errorf("load stopped fork metadata: %w", err)
 			}
 			meta.StoredMetadata.VsockCID = generateVsockCID(forkID)
+			meta.StoredMetadata.FirecrackerSnapshotCacheKey = ""
+			meta.StoredMetadata.FirecrackerUseUFFDOnNextRestore = false
+			meta.StoredMetadata.FirecrackerUFFDSessionID = ""
+			meta.StoredMetadata.FirecrackerUFFDPagerVersion = ""
```
hiroTamada: nit: this stopped-target branch clears SnapshotCacheKey / UseUFFDOnNextRestore / UFFDSessionID / UFFDPagerVersion but not FirecrackerDeferredSnapshotMemoryPath. in practice a stopped-target fork never sets it (useFirecrackerUFFDOnNextRestore is false for StateStopped), so it's already empty — but clearing it here too would be consistent belt-and-suspenders (and falls out naturally if the clearing helper suggestion above is adopted).

hiroTamada (approved):
nice work — the one-shot design cleanly confines the fragile pager/backing dependency to a single post-fork restore, and the lifecycle clearing + the `TestFCUFFDOneShotLifecycle` integration test (disk + /dev/shm files across the full fork→standby→restore→chained-fork chain) give good confidence on the snapshot-isolation risk.

both bugbot findings look already resolved on this head: the chained-fork backing path via the `firecrackerDeferredSnapshotMemoryPath` helper (regression-tested at firecracker_test.go:817), and the session close on the network-reconfigure failure path (restore.go:337).

a few non-blocking nits below, mostly around maintainability. approving.

lib/instances/stop.go
sjmiller609: Done in c74c539c via clearFirecrackerUFFDRestoreState. I kept FirecrackerSnapshotCacheKey path-specific because some flows intentionally preserve or refresh the shared snapshot cache key, but the one-shot/session/deferred fields now share one clearing helper.

lib/hypervisor/firecracker/firecracker.go
sjmiller609: Agree these helpers need to stay conceptually aligned. I am leaving the code as-is in this PR to keep the change small and avoid introducing a new neutral package just for this path fallback; we can extract once this code settles or if another caller appears.

scripts/install.sh
sjmiller609: I checked the infra deploy path. The hypeman deploy workflow checks out the target hypeman ref and runs Ansible with --tags build,deploy; the hypeman_install role downloads/runs that ref scripts/install.sh. The installer stops hypeman, installs the new API/UFFD binaries, writes the hypeman and hypeman-uffd@.service unit templates, runs systemctl daemon-reload, then starts hypeman. So the normal rollout lands the updated unit template before the new API binary starts.

lib/instances/manager.go
sjmiller609: Added the comment in c74c539c. I kept this as process-lifetime retention rather than delete-time cleanup because entries are tiny and bounded by distinct snapshot source paths seen by the host.

lib/instances/fork.go
sjmiller609: Covered by c74c539c as part of clearFirecrackerUFFDRestoreState; the stopped-target branch now clears FirecrackerDeferredSnapshotMemoryPath too.

PR #271, opened 2026-06-05
ingress: allow per-port listen-address override (tailnet-only ingresses)

rgarcia:
## Use case

Hypeman's ingress (Caddy) currently binds every listen port to a single
global address (`caddy.listen_address`, default `0.0.0.0`), so all ingress
ports are exposed on the public NIC.

We want the CDP (`9222`) and ChromeDriver (`9224`) ingresses to be reachable
**only on a non-public interface** (the host's Tailscale IP) so they are never
served on the public NIC — defense-in-depth at the listener, complementing the
host firewall (which is also being restricted for 9222/9224). The browser API
(`444`) and VNC (`443`) ingresses must stay public.

## Change

Add an optional config map under `caddy` that overrides the listen address for
specific listen ports. Ports not present in the map fall back to the global
`listen_address`.

```yaml
caddy:
  listen_address: 0.0.0.0        # global default, unchanged
  port_listen_addresses:
    9222: "100.107.186.40"       # CDP        -> tailnet only
    9224: "100.107.186.40"       # ChromeDriver -> tailnet only
```

Caddy already supports a per-server `listen` array, so this is a natural
extension: `buildConfig` builds one server per listen port, and now resolves
the bind address per port via `listenAddressForPort(port)` instead of always
using the global address.

The field is threaded through:
`CaddyConfig.PortListenAddresses` (cmd/api/config) → `ingress.Config.PortListenAddresses`
→ `CaddyConfigGenerator` → per-server `listen` in `buildConfig`.

This is config-level (not a per-Ingress API field) so it fits the
Ansible-managed host config and the e2e API client does not need to know the
Tailscale IP.

## Backward compatibility

- `caddy.listen_address` default stays `0.0.0.0`.
- When `port_listen_addresses` is unset or empty, every server binds the global
  listen address exactly as before — no behavior change.
- An empty string value for a port also falls back to the global address.

## Tests

- `lib/ingress` table-driven generator tests: override applied (CDP/ChromeDriver
  on the tailscale IP, 443/444 still on `0.0.0.0`), nil map, empty map, and
  empty-value fall-through.
- `cmd/api/config` tests: YAML loads the `port_listen_addresses` map while
  `listen_address` stays `0.0.0.0`; default config has no overrides.
- `go build ./...`, `go vet ./...`, and `go test ./lib/ingress/... ./cmd/api/config/...` all pass locally.

### DIFF
--- a/cmd/api/config/config.go
+++ b/cmd/api/config/config.go
@@ -76,6 +76,20 @@ type CaddyConfig struct {
 	AdminPort       int    `koanf:"admin_port"`
 	InternalDNSPort int    `koanf:"internal_dns_port"`
 	StopOnShutdown  bool   `koanf:"stop_on_shutdown"`
+
+	// PortListenAddresses overrides the listen address for specific listen ports.
+	// Keyed by listen port, value is the bind address (e.g. a Tailscale IP).
+	// Ports not present in the map fall back to ListenAddress (default "0.0.0.0").
+	//
+	// Use case: bind CDP/ChromeDriver ingresses to a non-public (e.g. Tailscale)
+	// interface so they are not reachable on the public NIC, while keeping
+	// 443/444 public. Example (YAML):
+	//
+	//   caddy:
+	//     port_listen_addresses:
+	//       9222: "100.107.186.40"
+	//       9224: "100.107.186.40"
+	PortListenAddresses map[int]string `koanf:"port_listen_addresses"`
 }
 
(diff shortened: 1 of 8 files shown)


sjmiller609:
closing as stale: the immediate 9222/9224 exposure was addressed with tailnet-only firewall rules in kernel/infra#590, and this draft now conflicts with current ingress code.

PR #276, opened 2026-06-07
Add APFS clonefile fast path for macOS VM disk forks

rgarcia:
## Summary

Wires Apple's **APFS `clonefile(2)`** into hypeman's `forkvm` reflink slot on macOS, so VM disk forks become near-instant, copy-on-write, and space-efficient on Apple Silicon instead of falling back to the slow `SEEK_DATA`/`SEEK_HOLE` byte copy. Includes the original design proposal under `docs/proposals/`.

## What's in this PR

- **Implementation** (`lib/forkvm`): `copy_reflink_darwin.go` implements `copyRegularFileReflink` via `unix.Clonefile` (`CLONE_NOFOLLOW|CLONE_NOOWNERCOPY`; removes any stale destination since `clonefile` requires the dst not to exist; then `chmod`s to honor the perms contract), with a darwin errno ladder: only `ENOTSUP`/`EOPNOTSUPP`/`EXDEV` map to `ErrReflinkUnsupported` (sparse fallback). `EEXIST` is owned by the pre-clone `os.Remove`, and `EINVAL`/`ENOTDIR` are programming/path errors, so those plus real errors (`EIO`/`ENOSPC`/`EACCES`) propagate rather than silently fall back. The `!linux` stub is retagged `!linux && !darwin`. No call-site changes — it plugs into the existing reflink→sparse ladder.
- **No dependency change**: `golang.org/x/sys v0.38.0` already exports `unix.Clonefile` for darwin (hidden from the linux-default `go doc`), so no `x/sys` bump, cgo shim, or raw syscall.
- **Tests**: darwin clone-correctness, stale-destination, CoW write-isolation, best-effort CoW block-sharing, and errno-mapping tests, plus the `BenchmarkForkGuestDisk` micro-benchmark and the end-to-end `TestVZForkSpeed`.
- **Design doc**: `docs/proposals/apfs-clonefile-forks.md`, comparing against Tart (open-source prior art) and the cross-volume `EXDEV` constraint.

## Benchmark — macOS (Apple M1, self-hosted CI runner)

**End-to-end VM fork** — `ForkInstance` copies the guest disk, rewrites the snapshot manifest, and materializes a new (stopped) instance. Source given a dense **2 GiB** overlay (3 forks each):

| fork (2 GiB-disk VM) | latency |
|---|---|
| **clonefile** (this PR) | **~1.1 ms** |
| sparse (previous behavior) | **~768 ms** |

**≈ 680× faster.** Fork latency is dominated by the disk copy, so clonefile collapses it to roughly the fixed fork overhead (~1 ms), and the fork is copy-on-write — ~no extra disk until the guest writes diverging pages.

<details>
<summary>Raw disk-copy micro-benchmark (single dense 2 GiB file)</summary>

Isolates the copy primitive itself:

```
goos: darwin
goarch: arm64
BenchmarkForkGuestDisk/clonefile-10    5      197892 ns/op   10851818.11 MB/s    1635 B/op   13 allocs/op
BenchmarkForkGuestDisk/sparse-10       5   420316492 ns/op       5109.21 MB/s 1050635 B/op   17 allocs/op
```

clonefile ~0.2 ms vs sparse ~420 ms = **~2,100×** (`clonefile` is metadata-only, so its MB/s reflects that no data is moved).
</details>

Reproduce — e2e fork: `HYPEMAN_FORK_BENCH=1 go test -run TestVZForkSpeed -v -timeout=20m ./lib/instances/`; disk-copy: `go test -run='^$' -bench=BenchmarkForkGuestDisk ./lib/forkvm/`.

## CLI end-to-end QA (Apple M1, APFS)

Verified the fast path through the actual CLI (`hypeman fork`) against a local server, complementing the server-side benchmarks above. Same host, same recipe, run on **`main` (sparse copy, "before")** and on **this PR (clonefile, "after")**.

**Behavior** (`run → standby → fork → restore`): the fork boots, gets a distinct id, and execs correctly; writes in the fork stay isolated from the source and the restored source shows no leak (copy-on-write); and a fork of a source carrying a dense 2 GiB overlay has its file intact at exactly `2147483648` bytes — the clone preserves content.

**End-to-end fork latency** — wall-clock of `hypeman fork <src> <name> --target-state Stopped` (3 runs each). This is the *full* round-trip (HTTP + snapshot-manifest rewrite + instance materialization), so it's larger than the server-side ~1 ms copy primitive:

| source overlay (allocated) | before — `main` (sparse) | after — this PR (clonefile) |
|---|---|---|
| ~empty (69–151 MiB) | 75 / 83 / 109 ms | 16 / 17 / 20 ms |
| 2.1 GiB (dense) | 1296 / 1512 / 2398 ms | 17 / 18 / 23 ms |

On `main` the sparse copy **scales with allocated data** (~90 ms at 69 MiB → ~1.3–2.4 s at 2.1 GiB). With clonefile, fork latency is **flat at ~17–23 ms regardless of overlay size** — a ~14× larger overlay forks in the same time, i.e. the metadata-only CoW clone confirmed end-to-end. For the 2 GiB fork that's **≈70–140× faster** through the CLI (the server-side copy primitive alone is ~680×; the e2e ratio is lower because CLI wall-time also includes fixed API + materialization overhead that clonefile doesn't change). Overlays were inflated by writing 2 GiB into the guest rootfs; `du` confirmed `overlay.raw` grew to 2.1 GiB allocated in both runs.

---

> [!NOTE]
> **Medium Risk**
> Touches the core fork disk-copy path on darwin; incorrect errno→fallback mapping could silently slow forks or surface wrong errors, though Linux is unchanged and fallback behavior is heavily tested.
> 
> **Overview**
> Adds an **APFS `clonefile(2)` fast path** on macOS by implementing `copyRegularFileReflink` in a new darwin build (`copy_reflink_darwin.go`) via `unix.Clonefile`, so VM guest disk copies during fork use copy-on-write instead of always hitting the sparse byte-copy fallback. The non-Linux stub is narrowed to `!linux && !darwin`; **no public API or `copy.go` changes**—the existing reflink→sparse ladder and `reflinkDead` short-circuit stay as-is.
> 
> Darwin-specific behavior: remove any existing destination before clone (unlike Linux FICLONE), apply `CLONE_NOFOLLOW|CLONE_NOOWNERCOPY`, then `chmod` for perms; only `ENOTSUP`/`EOPNOTSUPP`/`EXDEV` map to `ErrReflinkUnsupported` for sparse fallback. A test hook `cloneFileFn` injects clone failures.
> 
> Also adds the design RFC (`docs/proposals/apfs-clonefile-forks.md`), darwin unit tests (correctness, stale dst, CoW isolation, errno mapping, injected fallback), `BenchmarkForkGuestDisk`, and opt-in `TestVZForkSpeed` for end-to-end fork latency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5850635031dffa4f612d156d6a2a0ad8c7e4a3d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -211,6 +211,19 @@ jobs:
           export HYPEMAN_TEST_PREWARM_DIR="$HOME/.cache/hypeman-ci/darwin-arm64"
           go run ./cmd/test-prewarm
 
+      # TEMP(ci): one-off end-to-end VM fork-speed measurement (clonefile vs
+      # sparse) for the PR description. Remove after capturing the numbers.
+      - name: TEMP benchmark e2e fork speed
+        env:
+          HYPEMAN_FORK_BENCH: "1"
+          DEFAULT_HYPERVISOR: vz
+          JWT_SECRET: ci-test-secret
+          HYPEMAN_TEST_PREWARM_STRICT: "1"
+          HYPEMAN_TEST_REGISTRY: 127.0.0.1:5001
+        run: |
+          export HYPEMAN_TEST_PREWARM_DIR="$HOME/.cache/hypeman-ci/darwin-arm64"
+          make test-darwin TEST='TestVZForkSpeed' VERBOSE=1 TEST_TIMEOUT=20m 2>&1 | tee /tmp/forkspeed.txt
+
       - name: Check gofmt
         run: |
           set -euo pipefail
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -236,7 +236,7 @@ jobs:
           HYPEMAN_TEST_REGISTRY: 127.0.0.1:5001
         run: |
           export HYPEMAN_TEST_PREWARM_DIR="$HOME/.cache/hypeman-ci/darwin-arm64"
-          make test TEST='Reflink|CopyGuestDirectory|CopyRegularFile|Sparse|Clone'
+          make test TEST="'Reflink|CopyGuestDirectory|CopyRegularFile|Sparse|Clone'"
       - name: Cleanup
         if: always()
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,8 +5,6 @@ on:
 
 jobs:
   test:
-    # TEMP(ci): disabled on this branch to focus signal on the forkvm darwin feature.
-    if: false
     runs-on: [self-hosted, linux, x64, kvm]
     steps:
       - uses: actions/checkout@v4
@@ -234,14 +232,9 @@ jobs:
           JWT_SECRET: ci-test-secret
           HYPEMAN_TEST_PREWARM_STRICT: "1"
           HYPEMAN_TEST_REGISTRY: 127.0.0.1:5001
-          # TEMP(ci): VERBOSE surfaces per-test PASS/SKIP so the narrowed run
-          # proves the clone tests actually executed; HYPEMAN_REQUIRE_APFS_CLONE
-          # turns an unexpected non-APFS skip into a hard failure.
-          VERBOSE: "1"
-          HYPEMAN_REQUIRE_APFS_CLONE: "1"
         run: |
           export HYPEMAN_TEST_PREWARM_DIR="$HOME/.cache/hypeman-ci/darwin-arm64"
-          make test TEST="'Reflink|CopyGuestDirectory|CopyRegularFile|Sparse|Clone'"
+          make test
       - name: Cleanup
         if: always()
         run: |
@@ -250,8 +243,6 @@ jobs:
           make clean
 
   e2e-install:
-    # TEMP(ci): disabled on this branch to focus signal on the forkvm darwin feature.
-    if: false
     runs-on: [self-hosted, macos, arm64]
     needs: test-darwin
     concurrency:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,8 @@ on:
 
 jobs:
   test:
+    # TEMP(ci): disabled on this branch to focus signal on the forkvm darwin feature.
+    if: false
     runs-on: [self-hosted, linux, x64, kvm]
     steps:
       - uses: actions/checkout@v4
@@ -234,7 +236,7 @@ jobs:
           HYPEMAN_TEST_REGISTRY: 127.0.0.1:5001
         run: |
           export HYPEMAN_TEST_PREWARM_DIR="$HOME/.cache/hypeman-ci/darwin-arm64"
-          make test
+          make test TEST='Reflink|CopyGuestDirectory|CopyRegularFile|Sparse|Clone'
       - name: Cleanup
         if: always()
         run: |
@@ -243,6 +245,8 @@ jobs:
           make clean
 
   e2e-install:
+    # TEMP(ci): disabled on this branch to focus signal on the forkvm darwin feature.
+    if: false
     runs-on: [self-hosted, macos, arm64]
     needs: test-darwin
     concurrency:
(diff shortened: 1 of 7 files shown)


hiroTamada (approved):
reviewed — solid PR overall, ~789 LOC scoped cleanly to `lib/forkvm` + tests + an RFC. a few things worth a look:

### PR description vs. code
- the PR body claims `EEXIST` and `ENOTDIR` map to `ErrReflinkUnsupported` fallback, but `lib/forkvm/copy_reflink_darwin.go:67-73` (deliberately, per the in-code comment) does not. tests at `lib/forkvm/copy_reflink_darwin_test.go:154-156` explicitly assert the actual behavior. code + tests are right; PR description is wrong.

### Test coverage
- no automated test of CoW *write isolation* (write to clone → source unchanged). the PR body claims one ("best-effort CoW block-sharing"), but I only see it verified in the manual CLI QA section. worth a small addition to `copy_reflink_darwin_test.go` since write isolation is the headline correctness property of CoW.

### Test hygiene
- `lib/instances/fork_speed_darwin_test.go:111-131` — `forkvm.SetReflinkDisabled` is a global flag, reset in two places but leaks if `mgr.DeleteInstance` at line 126 fails. one `t.Cleanup(func() { forkvm.SetReflinkDisabled(false) })` near the top would simplify. low-risk since env-gated.
- `lib/forkvm/copy_bench_darwin_test.go` — single 2 GiB source shared across `clonefile` and `sparse` sub-benchmarks means the sparse path runs cache-warm. ratio still valid for the comparison, but worth a line in the leading comment so future readers don't read the absolute sparse number as cold-cache.

### Doc
- `docs/proposals/apfs-clonefile-forks.md:264` — claims `unix.Clonefile` is gated `//go:build darwin && arm64`. if true, Intel-mac builds wouldn't link the wrapper. worth a quick `GOOS=darwin GOARCH=amd64 go doc golang.org/x/sys/unix Clonefile` check before merging.

### Nits
- `lib/forkvm/copy_reflink_darwin_test.go:55-58` — `defer listener.Close()` could be `t.Cleanup`, minor consistency thing.

rgarcia:
Thanks for the review @hiroTamada — all addressed in 5850635 (non-blocking, but worth tidying before merge):

- **errno ladder (description vs code):** you're right, the code+tests are correct (`EEXIST`/`ENOTDIR` propagate; only `ENOTSUP`/`EOPNOTSUPP`/`EXDEV` fall back). Fixed the PR description to match.
- **CoW write-isolation test:** added `TestCopyRegularFileReflink_WriteIsolation` — write to the clone, assert the source is byte-for-byte unchanged. This was the one correctness property only covered by the manual QA before.
- **`SetReflinkDisabled` leak:** moved the reset to `t.Cleanup` in `fork_speed_darwin_test.go`, so a `require` failure mid-loop can't leak the flag into other tests; dropped the now-redundant manual resets.
- **bench cache-warm:** added a note to `BenchmarkForkGuestDisk` that the shared source means the sparse run reads cache-warm (ratio valid; absolute sparse number is a warm-cache best case).
- **`listener.Close` nit:** → `t.Cleanup`.
- **`unix.Clonefile` build tag:** ran your check — `GOOS=darwin GOARCH=amd64 go doc golang.org/x/sys/unix Clonefile` resolves fine (it's generated per-arch in `zsyscall_darwin_arm64.go` *and* `zsyscall_darwin_amd64.go`), so Intel builds link it. Corrected the proposal wording (it had implied an arm64 gate).

One I intentionally left: the `reflinkDisabled` global. It predates this PR (`lib/forkvm/copy.go` on `main`) and existing tests (`copy_reflink_test.go`, `copy_sparse_unix_test.go`) rely on it to exercise the sparse fallback deterministically on a reflink-capable host — so removing it is out of scope here.

`go test ./lib/forkvm/` (darwin, APFS) green incl. the new write-isolation test.

PR #277, opened 2026-06-07
standby/restore (vz): document the clonefile win + fix restore state reporting

rgarcia:
> **Point-in-time note (rewritten):** this PR started as "compress machine-state + overlap the save with the disk copy." Investigation showed **neither mechanism helps on vz today** — the real standby/fork speedup already shipped in #276 (APFS `clonefile` CoW disk copy). So this PR was slimmed to an RFC documenting that finding plus the small fixes that survived. Net diff: **4 files, +307/−8.**

## What actually made vz standby/fork fast: #276 (clonefile)
APFS `clonefile` makes the guest-disk copy a copy-on-write, size-independent ~1ms operation. That — not anything in this PR — is the win. See #276 for before/after fork timings.

## Why the two originally-proposed mechanisms are no-ops on vz
**A — compress `machine-state.vzm`:** a no-op. Apple's `SaveMachineStateToPath` already omits zero/free guest pages, so the artifact is a high-entropy residual. Measured: zstd compresses it **~0%**, lz4 nets **larger**. (This is distinct from Tart, whose beneficial compression is on the OCI *disk image* — hypeman's equivalent is the overlay, handled by #276.)

**B — overlap save ∥ disk-copy in the paused window:** no measurable speedup now that the copy is ~1ms via clonefile. The paused window is dominated by the machine-state save, and the disk copy that B would hide behind it is already negligible. Not worth the concurrency/complexity on vz.

Neither A nor B is implemented here. Both still matter for the **cross-volume / sparse-copy fallback** path and for the **firecracker/cloud-hypervisor** backends, so the RFC keeps them on the table for those cases.

## Measured CLI timings (Apple Silicon, vz, 4 GB guest)
End-to-end wall-clock of the actual CLI commands: `standby` timed until the instance reaches `Standby` (machine state fully saved + disk cloned + VM stopped); `restore` timed until `Running` **and** an immediate `exec` succeeds. Median of 3, configured RAM fixed at 4 GB; the variable is how much guest RAM was *dirtied* before standby.

| dirtied RAM | `standby` → Standby | `restore` → Running+exec | `machine-state.vzm` |
|------------:|--------------------:|-------------------------:|--------------------:|
|       64 MB |              1666ms |                    761ms |               164 MB |
|      256 MB |              1813ms |                   1111ms |               360 MB |
|      512 MB |              1839ms |                   1685ms |               624 MB |
|    1024 MB |              2232ms |                   2814ms |              1120 MB |
|    2048 MB |              2641ms |                   5365ms |              2048 MB |

Takeaways:
- **State size tracks *dirtied* RAM, not configured RAM** (a 4 GB guest with 2 GB touched saves a ~2 GB `.vzm`, not 4 GB) — direct confirmation that vz omits zero/free pages, which is why mechanism A can't compress it.
- **standby/restore time is dominated by the machine-state save/load**, proportional to the live working set. The guest-disk copy is ~1ms (clonefile) and contributes nothing measurable — which is precisely why overlapping it (mechanism B) is a no-op.

## What this PR lands (the keepers)
- **`docs/proposals/faster-standby-restore.md`** — the RFC above, with the measurements.
- **`lib/hypervisor/vz/client.go`** — `Resume()` now polls `vm.info` until the shim reports raw state `Running` (10s bound, 25ms poll) instead of being fire-and-forget. So `restore` reports `Running` and an immediate `exec` succeeds. This replaces a generic hypervisor-state prime that had leaked vz-specific semantics up into `lib/instances`.
- **`Makefile`** — drop a dead duplicate `run` target; silence duplicate-library linker warnings on darwin builds.
- **`lib/instances/...darwin_test.go`** — serial running-source snapshot→fork→restore roundtrip e2e regression test.

## Verification
Live-tested on Apple Silicon (vz): standby→restore reports `(state: Running)` with immediate `exec` succeeding and uptime preserved; fork-from-running boots, sees source state, stays CoW-isolated (no write leak back to source). Timings above captured on the same setup.

---

> [!NOTE]
> **Low Risk**
> Behavior change is localized to vz Resume polling (adds up to ~10s wait on failure); restore path is more correct for immediate exec. Docs and test-only additions otherwise.
> 
> **Overview**
> Adds **`docs/proposals/faster-standby-restore.md`**, an RFC that records why vz standby/fork latency is dominated by **`machine-state.vzm` save/load** (with #276’s APFS **`clonefile`** disk copy already ~instant), and why **compressing `machine-state.vzm`** and **overlapping save with disk copy** were measured as no-ops on vz and not shipped.
> 
> **`lib/hypervisor/vz/client.go`**: **`Resume()`** no longer treats the shim’s fire-and-forget resume PUT as done—it **polls raw `vm.info` until state is `Running`** (10s cap, 25ms interval) so restore callers see a real **Running** post-condition before **`exec`**.
> 
> **`lib/instances/snapshot_running_source_darwin_test.go`**: new darwin e2e **`TestVZRunningSnapshotRoundTrip`** (running source → standby snapshot → fork → restore) with guest marker verification.
> 
> **`Makefile`**: removes an older duplicate **`run`** recipe; adds **`DARWIN_LDFLAGS`** to silence Xcode 15+ duplicate **`-lobjc`** linker warnings on **`vz-shim`** and **`hypeman`** darwin builds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 021b4f467e6cd8239b38ae1a0d8864d955b1f69d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,11 +263,6 @@ $(BIN_DIR)/hypeman-uffd-pager: | $(BIN_DIR)
 # Build all binaries
 build-all: build
 
-# Run without live reload (build once and run)
-run: build
-	sudo setcap cap_net_admin,cap_net_bind_service=+eip $(BIN_DIR)/hypeman
-	$(BIN_DIR)/hypeman
-
 # Run in development mode with hot reload
 # On macOS, redirects to dev-darwin which uses vz instead of cloud-hypervisor
 dev:
@@ -410,10 +405,15 @@ ENTITLEMENTS_FILE ?= vz.entitlements
 
 # Build vz-shim (subprocess that hosts vz VMs)
 # Also copies to embed directory so it gets embedded in the hypeman binary
+# DARWIN_LDFLAGS silences the macOS (Xcode 15+) linker's "ignoring duplicate
+# libraries: '-lobjc'" warning: cgo deps that link Virtualization.framework
+# specify -lobjc more than once, which the new linker flags. Harmless; suppress.
+DARWIN_LDFLAGS := -ldflags=-extldflags=-Wl,-no_warn_duplicate_libraries
+
 .PHONY: build-vz-shim
 build-vz-shim: | $(BIN_DIR)
 	@echo "Building vz-shim for macOS..."
-	go build -o $(BIN_DIR)/vz-shim ./cmd/vz-shim
+	go build $(DARWIN_LDFLAGS) -o $(BIN_DIR)/vz-shim ./cmd/vz-shim
 	mkdir -p lib/hypervisor/vz/vz-shim
 	cp $(BIN_DIR)/vz-shim lib/hypervisor/vz/vz-shim/vz-shim
 	@echo "Build complete: $(BIN_DIR)/vz-shim"
@@ -431,7 +431,7 @@ sign-vz-shim: build-vz-shim
 .PHONY: build-darwin
 build-darwin: build-embedded build-vz-shim | $(BIN_DIR)
 	@echo "Building hypeman for macOS with vz support..."
-	go build -tags containers_image_openpgp -o $(BIN_DIR)/hypeman ./cmd/api
+	go build -tags containers_image_openpgp $(DARWIN_LDFLAGS) -o $(BIN_DIR)/hypeman ./cmd/api
 	@echo "Build complete: $(BIN_DIR)/hypeman"
 
 # Sign the binary with entitlements (required for Virtualization.framework)
--- a/docs/proposals/faster-standby-restore.md
+++ b/docs/proposals/faster-standby-restore.md
@@ -1,0 +1,98 @@
+# RFC: Faster macOS VM standby/restore (snapshot save/restore)
+
+## Summary
+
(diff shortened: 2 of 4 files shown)


pbjaru (approved):
LGTM! reviewed — overall solid. the resume fix is correct and well-targeted, the proposal doc is accurate, and the new test adds real running-source coverage. a couple small things:

### Doc
- `docs/proposals/faster-standby-restore.md:80` — there's no `macOSAvailable(14)` in the code; the macOS-14 check that exists is the test helper `isMacOS14OrLater`, and production save/restore is gated by the `darwin && arm64` build tags + `ValidateSaveRestoreSupport()`. consider renaming to match.

### Nit
- `lib/hypervisor/vz/client.go:270` — `waitForRawState` takes `timeout`/`interval` params but has one caller passing constants and no unit test. either inline the constants, or add a small unit test exercising the success/timeout branches with a fake state source.

PR #279, opened 2026-06-07
Rosetta x86-64 emulation and multi-platform image pull for vz guests

rgarcia:
## What this does

Runs **x86-64 (amd64) container images on Apple-Silicon `vz` microVMs via Rosetta**, and adds a Docker-style **`platform`** field so callers can ask for a specific architecture. On an arm64 host, `hypeman run --platform linux/amd64 …` boots the amd64 image and its binaries execute through Apple's Rosetta translator inside the arm64 guest.

Two capabilities, threaded end-to-end:

1. **Multi-platform image pull** — `platform` (`os/arch[/variant]`, e.g. `linux/amd64`) on image + instance create, modeled on `docker --platform`.
2. **Rosetta emulation** — host attaches a Rosetta virtio-fs share; guest `init` registers a `binfmt_misc` handler so amd64 ELFs dispatch through Rosetta. Enabled **automatically** when the resolved image arch differs from the host (no user flag), mirroring Docker Desktop.

## Key design decisions (the non-obvious bits worth a reviewer's attention)

- **Rosetta is derived, not a flag.** There is no `enable_rosetta` API field. `createInstance` compares the resolved image arch to the host and turns Rosetta on for emulated images on `vz`/Apple-silicon; any other host rejects an emulated image with a clear error. `deriveEnableRosetta` is a pure, unit-tested function.
- **`platform` is recorded from the image config, not the request.** The request is validated against the manifest; a mismatch fails the build. Today only `linux` / `amd64|arm64` are accepted (validated up front).
- **A no-platform run resolves the *host* arch, never a drifted tag.** Tags are mutable (`last-pull-wins`), so `createInstance` resolves the host variant explicitly instead of following the tag symlink — it will not silently boot an emulated image.
- **Instance image is decoupled: display vs. boot.** `instance.image` stays the **caller-supplied reference** (the tag) — unchanged from today for API consumers. A separate internal `ResolvedImage` holds the digest-pinned ref that start/restore use to locate the rootfs, so a flipped tag can't change a running instance's arch across restarts. (This replaced an earlier approach that exposed the digest on `instance.image`; see commit `instances: keep caller image ref for display…`.)
- **Errors map to real HTTP codes.** Unavailable platform → `404 platform_not_available`, registry rate limit → `429 rate_limited`, missing repo/digest → `404 not_found`, bad platform string → `400 invalid_platform`. No client/registry condition returns 500 on these paths.

## How to review

Suggested order (build context first, then the mechanical bits last):

1. **`docs/proposals/rosetta-x86.md`** — the design + rationale; orients everything else.
2. **`lib/images/platform.go`** — the `Platform` type, parsing/validation, host/emulation helpers. Single source of truth.
3. **`lib/images/{oci,manager,reference,errors}.go`** — manifest/digest resolution, tag repoint, registry-error classification.
4. **`lib/instances/{create,create_image,rosetta,restore,start}.go`** — host-default resolution, `deriveEnableRosetta`, `ResolvedImage`/`bootImageRef`, restore image pre-check.
5. **Rosetta plumbing** — `cmd/vz-shim/rosetta_*.go` (host device), `lib/system/init/rosetta.go` (guest binfmt), config threading (`hypervisor.VMConfig` → `shimconfig.ShimConfig` → `vmconfig.Config`).
6. **API surface** — `cmd/api/api/{images,instances}.go` + `openapi.yaml`.
7. **Skim / mechanical:** `lib/oapi/oapi.go` is **generated** from `openapi.yaml` (don't read line-by-line); the `Architecture`→`Platform` rename across image types is mechanical.

## Testing

- **Unit (Linux CI `test`):** platform parse/validate, manifest-platform recording, `deriveEnableRosetta`, registry-error classification, tag-repoint, digest resolution, instance-platform echo, restore-missing-image mapping.
- **E2E (CI `test-darwin`, Apple silicon + Rosetta):** `TestVZRosettaImageX86` boots a real amd64 alpine and execs an amd64 ELF through Rosetta.
- **5 rounds of local QA** (parallel agents driving the CLI against a live server): verified amd64 exec via real ELF bytes, native-arm64 no-regression, digest pinning/dedup, snapshot/restore/fork arch parity, and every error path. All fixes from QA are in this PR; CI is green (`test`, `test-darwin`, scan, Bugbot).

## Out of scope (pre-existing, not touched by this PR)

Image-delete handler edge cases (delete-by-digest, not-found→500, multi-arch tag orphaning), non-Docker-Hub registry error mapping (gcr/quay/DNS→500), snapshot-fork-of-deleted-image→500, and a `cp`-to-dir nesting quirk are **pre-existing** (0 lines in this diff) and intentionally left for a separate change.

## Risk

Touches core image resolution and instance create/start/restore + vz device config. Main mitigations: host-default resolution (no silent emulation), digest-pinned `ResolvedImage` (no arch drift on restart), and snapshot device-parity preserved by replaying `EnableRosetta` from the manifest. One field note: under heavy *concurrent* vz snapshot/restore load, the framework can intermittently fail a machine-state restore (`VZError Code=12`); restores are reliable in isolation — not introduced here, flagged for ops awareness.

---

> [!NOTE]
> **High Risk**
> Changes core image resolution, instance create/start/restore paths, and vz VM device configuration; mistakes could cause wrong-arch boots, broken restores, or misclassified API errors.
> 
> **Overview**
> Adds **Docker-style `platform`** on image/instance create and echoes resolved `platform` on responses. Image pull resolves multi-arch manifests (including digest pins), records platform from the manifest, applies **last-pull-wins** tag repointing, and classifies registry failures into **400/404/429** instead of generic 500s.
> 
> **Instance lifecycle:** create resolves images per platform (default = host arch, not a drifted tag), stores an internal **digest-pinned `ResolvedImage`** for boot/start/restore, and rejects emulated images on non-capable hosts. Missing images on start/restore/fork/snapshot restore map to **404**.
> 
> On **Apple Silicon `vz`**, when the image arch differs from the host, **Rosetta is enabled automatically** (no API flag): `vz-shim` attaches a Rosetta virtio-fs device; guest `init` mounts it and registers **binfmt_misc** for amd64 ELFs.
> 
> Also: config **`~/` path expansion**, macOS **Rosetta** docs + **signed `vz-shim` tests**, prewarm **linux/amd64** mirror for E2E, and **`mkfs.ext4`** lookup for Homebrew e2fsprogs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4bf8093c4eb110c09c29bd6ba8157c3ae9c82f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


openapi.yaml
```diff
@@ -539,6 +539,10 @@ components:
           type: string
           description: OCI image reference
           example: docker.io/library/alpine:latest
+        platform:
+          type: string
+          description: 'Target platform as os/arch[/variant] (e.g. "linux/amd64"), matching Docker --platform. Omit for the host platform. Only os "linux" is supported today; arch must be amd64 or arm64.'
```
hiroTamada: enum maybe? It is little confusing as a user here

hiroTamada (approved):
reviewed end-to-end — strong PR overall. Pull-by-digest in `buildImage`, the manifest-as-source-of-truth platform recording, and the pure-function `deriveEnableRosetta` are the highlights. The guest-side binfmt setup is well thought-through (F-flag rationale, systemd `binfmt.d` drop-in for flush survival).

Approving. A handful of things worth a look before/after merge, none blocking:

### Behavior worth confirming
- `lib/instances/create_image.go:21-38` — when `platform` is non-empty, the resolver always calls `CreateImage` and skips the local `GetImage` fast path. If `CreateImage` is cheap when the image already exists, fine; otherwise every explicit-platform create pays a re-resolve. Worth double-checking.
- `lib/images/manager.go:148-151,289-295,355` — non-host pulls intentionally skip tag-symlink create/update so a `linux/amd64` pull on arm64 doesn't clobber the host-arch tag. Subsequent tag-only resolves on the same host will pull host-arch fresh. Worth a doc/changelog note for users.
- `lib/system/init/rosetta.go:97-99` — `registerBinfmt` silently tolerates EEXIST. If an image ships its own `:rosetta:` rule pointing at a different interpreter, we'd run that instead. Consider unregister-then-register, or at minimum log when an existing rule was found.

### Questions
- `lib/images/manager.go:131` — 5s manifest-resolve timeout is unchanged but multi-arch manifest-list resolution may need two registry round-trips. Still comfortable for slow registries?
- `lib/images/platform.go:140-152` — `resolveManifestPlatform` trusts the requested platform when the manifest omits architecture (synthetic/locally-built images). Worth a code-side comment acknowledging this trust path.
- PR description reads "non-host platform pins the ref to its manifest digest"; code in `create_image.go:21-38` pins for any explicit platform incl. host-matching. Align wording or code.

### Scope creep
- `cmd/api/config/config.go:507-526` — `~/` path expansion is orthogonal to Rosetta. Either split into its own PR or call it out in the description.

### Robustness
- `lib/images/mirror.go:43-56` — mirror passes platform to `remote.WithPlatform` but does not verify the pulled image config matches. For single-manifest source tags the request is silently ignored. The prewarm comment already explains the workaround on the consumer side; consider mirroring the `manager.go` post-pull guard for symmetry.
- `cmd/test-prewarm/main.go:30-49` — the "don't double-mirror a source:tag" invariant is comment-only. A duplicate-detection pre-flight in `main()` would turn a future silently-broken E2E into a clear error.
- `cmd/api/config/config.go:511-517` — path expansion applies to 5 named fields; future path additions will be silently missed. Consider a reflection walk or a clear anchor comment.

### Coverage gaps
- `lib/images/manager.go:148-151,289-295,355` — `shouldUpdateTagFor*` (skip tag symlink for non-host platforms) is one of the more user-visible behavior changes and has no direct unit test that I saw.
- `lib/system/init/rosetta.go:97-99` — `alreadyRegistered` EEXIST tolerance untested.

### Nits
- `cmd/api/api/images.go:43-45` vs `cmd/api/api/instances.go:309` — inconsistent optional-string deref style (manual if/deref vs `derefString`). Pick one.
- `lib/instances/create_image.go:24` — wrap drops the underlying sentinel (`errors.Is(ErrNotFound)`) for the platform path; no-platform path preserves it. Mild asymmetry.
- `lib/instances/create_image.go:100` — `storedImageNameForCreate` re-parses what `resolveImageForCreate` already computed.
- `lib/instances/create_image.go:117` — `ErrInvalidPlatform` reused for "stored platform unparseable" — that's a manager-corruption case, not user input. Could be a separate internal error.
- `lib/instances/create_image.go:82` — empty-digest wraps `ErrImageNotReady`; "no digest" isn't really "not ready". A more specific error would read better.
- `lib/instances/create.go:117-121` — comment is slightly redundant with the immediately following code.
- `lib/instances/rosetta.go:24-25` — error wording doesn't distinguish "wrong host" from "right host, wrong hypervisor" (darwin/arm64 + qemu falls through this branch).
- `lib/instances/rosetta.go:30-35` — `emulatedArchName` silently degrades if a third GOARCH ever lands; document the assumption.
- `lib/images/manager.go:225,230` — `tags.Clone(req.Tags)` done into both `storedReq.Tags` and `meta.Tags`. One clone would do.
- `lib/images/manager.go:133` — "resolve manifest for platform <host>" is slightly noisy in the empty-platform default case.
- `lib/images/manager.go:377-381` — `shouldUpdateTagForRequest` swallows parse errors. `ImportLocalImage` doesn't pre-validate, so a bad `meta.Request.Platform` would silently disable tag symlinking. Consider logging.
- `lib/images/reference.go:163-170` — `Resolve` and `ResolveForPlatform` are near-duplicates; consider having `Resolve` call `ResolveForPlatform` with `vmPlatform()`.
- `lib/images/oci_public.go:46-49` — `InspectManifestForLinux` now redundant; remove if no remaining callers.
- `lib/images/disk.go:285-290` — fallback returns the bare `"mkfs.ext4"` name; a clearer up-front error ("install e2fsprogs") would beat the eventual exec failure.
- `lib/images/platform.go:178-189` — `ImageNeedsHostEmulation` silently treats unparseable platform as host-native; consider warn-logging the parse failure.
- `cmd/vz-shim/rosetta_arm64.go:39-43` — `SetDirectorySharingDevicesVirtualMachineConfiguration` replaces the entire list. Add a note so a future contributor doesn't accidentally clobber by calling it again.
- `cmd/vz-shim/rosetta_arm64.go:22` — empty `case Installed:` fall-through is idiomatic but a one-word `// available, proceed` comment would help.
- `lib/system/init/rosetta.go:69` — 2s `waitForDevice` timeout could be tight on slow boots; failures here silently break amd64 execution since the caller logs and continues.
- `lib/system/init/rosetta.go:46-50` — "C applies the target's credentials" is ambiguous between binary and interpreter.
- `openapi.yaml:540,1245` — duplicated platform description; consider a shared schema component.

openapi.yaml
rgarcia: Good call on the confusion. I kept it a string rather than an `enum`: the value is the Docker `os/arch[/variant]` grammar (optional variant + the os×arch matrix), which a strict OpenAPI enum can't cleanly express, and we validate it server-side (bad values → `400 invalid_platform`). I clarified the description to say exactly that and added a light `pattern: ^[a-z0-9]+/[a-z0-9]+(/[a-z0-9.]+)?$` for documentation/early validation (commit 04cbf73). I considered extracting a shared `Platform` schema component to de-dup the description, but oapi-codegen would generate it as a named defined type and flip the request fields from `*string` to `*Platform`, rippling into the handlers — not worth it for a doc-only win, so I left the fields inline.

rgarcia:
### Review follow-ups addressed

Thanks @hiroTamada for the review 🙏 Pushed a few small cleanup commits (`e005fca`, `04cbf73`) — all no-behavior-change except one CLI fix noted below.

**Your inline comment (platform `enum`):** replied in-thread — kept it a `string` (Docker `os/arch[/variant]` grammar can't be a clean enum; validated server-side → `400 invalid_platform`), but clarified the description and added a light `pattern`.

**Other points from your review:**
- **`~/` path-expansion scope creep** — agreed it's orthogonal to Rosetta; it landed earlier in `334d64f` and I've called it out in the PR description so it's not a surprise. Also simplified `expandHomePath` per the nit.
- **binfmt existing-rule handling** — `init` now *logs* when an existing `:rosetta:` handler is found instead of silently tolerating it. (EEXIST tolerance is already covered by `TestRosettaAlreadyRegistered`.)
- **Comment verbosity / dedup** — consolidated the `last-pull-wins` rationale onto `createTagSymlink` (call sites are now one-liners), trimmed the over-long `resolveManifestPlatform` / empty==host comments, and de-duped the test-prewarm constraint note.
- **`ParsePlatform`** — collapsed the three per-arity empty-component checks into one rule.
- **Stale testing plan** — marked the snapshot/restore E2E and kernel-config assertion in the proposal as deferred follow-ups (not implemented here).

**Cursor Bugbot:** all six findings were already fixed at HEAD (instance-ignores-platform, wait-wrong-tag, start-resolves-tag→`ResolvedImage`, legacy-host-resolve, binfmt EEXIST, scratch `TODO.md`).

**Deliberately deferred** (non-blocking, can be follow-ups): the larger refactors — unifying `Resolve`/`ResolveForPlatform`, threading the parsed platform into `createAndQueueImage`, and the optional test-helper consolidations. Happy to do any of these if you'd prefer them in-PR.

Testing unchanged: unit + `test-darwin` E2E green, plus 5 local QA rounds. CI re-running on the cleanup commits.

PR #289, opened 2026-06-12
Fix tc class/filter leaks on the bridge and reconcile them periodically

hiroTamada:
## Summary

Per-VM HTB classes and `basic` filters leak on the bridge on busy hosts. Since the bridge filter list is evaluated **linearly for every VM-egress packet**, accumulated leaks degrade network throughput for every VM on the host (observed: ~1,880 stale class/filter pairs vs ~6 live VMs on one production host).

### Leak mechanisms fixed

1. **Wrong-ID deletes.** Cleanup paths that have no persisted class ID (periodic TAP reaper, orphaned-TAP cleanup, recreate-existing-TAP, TAP-create failure) fell back to the hash-derived class ID. Whenever the original `tc class add` had been collision-probed to `hash+N`, these paths deleted the wrong ID and leaked the real class *and* its filter — a feedback loop, since more stale classes cause more probing. `removeVMClass` now resolves the class from the bridge filter table via the TAP's ifindex (`meta(rt_iif eq N)` is the authoritative tap→class link). When neither a persisted ID nor a filter resolves, it deletes nothing and defers to the reconciler instead of guessing.

2. **Silent delete failures.** `removeVMClass` ran every `tc` delete best-effort with no logging. Failures are now logged and counted via `hypeman_network_tc_cleanup_failures_total{operation}`.

3. **Reconcile never ran and missed filters.** `CleanupOrphanedClasses` only ran at process startup and iterated classes only, so dangling filters (class deleted, filter left, or TAP gone) were invisible to it. It now:
   - runs every TAP GC tick (60s), not just at startup
   - holds the tc mutex so it can't observe a half-applied rate limit
   - sweeps filters not anchored to a live TAP ifindex, then classes not referenced by a live TAP's filter/derived ID or a stored allocation classid
   - bails out if filters exist but no `rt_iif` ematch parses, so an unexpected `tc` output format can't cause a sweep of live state
   - counts removals via `hypeman_network_tc_orphans_swept_total{kind}`

4. **Double-release noise.** The persisted classid file is cleared after an instance-scoped release, so the routine stop→delete double release no longer re-deletes the class.

### Out-of-band recovery

`scripts/sweep-stale-tc.sh` applies the same sweep logic standalone (dry-run by default, `--apply` to delete) to recover already-polluted hosts immediately, without waiting for a deploy/restart. Candidates are snapshotted before keep-sets so anything allocated mid-run is never considered stale.

## Test plan

- [x] `go build` / `go vet` on `lib/network` and `lib/instances`
- [x] `go test ./lib/network/` (includes new `parseBridgeFilters` unit tests)
- [x] `bash -n` + dry-run of `scripts/sweep-stale-tc.sh`
- [ ] Full `make test-linux` integration suite — **not run locally** (requires KVM + HTB kernel modules not available in my environment); needs CI or a metal host
- [ ] Validate on a polluted staging/prod host: run the sweep script dry-run, compare counts against `tc class show dev vmbr0 | wc -l`, then `--apply` and confirm `hypeman_network_tc_class_collisions_total` and softirq CPU drop

---

> [!NOTE]
> **Medium Risk**
> Changes live Linux bridge traffic-control delete and reconciliation logic under `tcMu`; incorrect sweeps could affect VM upload shaping, though multiple safety bail-outs limit that risk.
> 
> **Overview**
> Fixes **leaked per-VM HTB classes and `basic` filters** on the bridge, which linearly slow every VM’s egress when they accumulate.
> 
> **Release and delete paths** now resolve the real class via the bridge filter table (`meta(rt_iif eq N)`) instead of guessing the hash-derived ID after collision probing; they skip deleting when a persisted ID appears owned by another live TAP. Failed `tc` deletes are logged and counted. Instance-scoped release **clears the persisted `classid` file** after a successful delete so stop→delete doesn’t repeat cleanup.
> 
> **`CleanupOrphanedClasses`** is extended to sweep dangling filters first, then unreferenced classes (live TAP ifindexes, filter flowids, and stored allocation class IDs), holds **`tcMu`** to avoid racing async rate-limit setup, and bails if filter output can’t be parsed safely. It runs on **every 60s TAP GC tick** (even when the TAP pass is skipped), not only at startup.
> 
> Adds **`hypeman_network_tc_cleanup_failures_total`** and **`hypeman_network_tc_orphans_swept_total`**, unit tests for **`parseBridgeFilters`**, and **`scripts/sweep-stale-tc.sh`** for dry-run / `--apply` recovery on already-polluted hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a7d1938a4757cd2e9c2a514a531433512549fb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sjmiller609:
unrequesting review to clear queue, please re-request if still applicable

PR #309, opened 2026-07-15
Reduce test wall-clock time without weakening coverage

IlyaasK:
## What

Reduce the Hypeman test suite wall-clock time by removing measured lifecycle overhead:

- cancel the full native compression process group when its context expires
- use immediate VM shutdown only in test teardown while retaining the normal cleanup path
- prepare VM-ready test images once per run instead of repeatedly unpacking layers
- use 512 MiB guests for lifecycle tests that do not exercise memory capacity

## Why

Wall-clock profiling showed that the suite was spending substantial time in repeated image preparation, graceful shutdown timeouts, oversized lifecycle guests, and a cancellation path that killed the shell but left its child process holding output pipes open.

These changes address those root causes directly. They do not shard the suite or add a test matrix.

## Result

Linux Run tests step:

- before: 259 seconds
- after: 190 seconds
- reduction: 69 seconds, or 26.6 percent

The full Linux job, macOS job, and install checks passed.

## Test strength

No tests, assertions, VMM backends, or integration scenarios are removed.

Production DeleteInstance behavior remains unchanged. Immediate shutdown is restricted to test cleanup and terminal test teardown. Dedicated memory and resource-capacity tests retain their larger guest sizes.

The compression cancellation regression test also asserts that cancellation returns within one second, so the original 30-second child-process leak cannot silently return.

## Verification

- Before: https://github.com/kernel/hypeman/actions/runs/29444740345
- After: https://github.com/kernel/hypeman/actions/runs/29450412921

---

> [!NOTE]
> **Medium Risk**
> Production snapshot compression cancellation behavior changes (process-group kill); delete graceful-shutdown skipping is test-only but lives on the shared delete path.
> 
> **Overview**
> Cuts integration test wall-clock by attacking repeated setup and teardown overhead, without dropping scenarios or assertions.
> 
> **Prewarm builds VM-ready images once per CI run:** `test-prewarm` now runs `CreateImage` + `WaitForReady` for each mirrored image, and `EnsureImageReady` prefers `HYPEMAN_TEST_PREWARM_DIR` so tests seed from that cache instead of rebuilding layers. Linux CI **chowns** the prewarm directory so the job user can write under `/ci/prewarm`.
> 
> **Faster test VM teardown:** `deleteInstanceWithOptions` gains `skipGracefulShutdown`; `deleteTestInstanceNow` uses it for cleanups while public `DeleteInstance` behavior is unchanged.
> 
> **Smaller default lifecycle guests:** many integration tests use `lifecycleTestMemorySize` (512 MiB) where memory size is not the subject; capacity-focused tests keep larger sizes.
> 
> **Compression cancel fix (production):** native zstd/lz4 helpers run in a process group and get **SIGKILL on the whole group** when the context is canceled, so `Wait` does not hang on orphaned children; the regression test now requires cancellation within one second.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de5fe4fece127e37ab21cd73732f3c79cec01627. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -121,6 +121,7 @@ jobs:
           # cross-filesystem safe. Keeping both on /ci avoids that.
           export HYPEMAN_TEST_PREWARM_DIR="/ci/prewarm/linux-amd64"
           mkdir -p "$HYPEMAN_TEST_PREWARM_DIR"
+          sudo chown -R "$(id -u):$(id -g)" "$HYPEMAN_TEST_PREWARM_DIR"
           go run ./cmd/test-prewarm
 
       - name: Check gofmt
--- a/cmd/test-prewarm/main.go
+++ b/cmd/test-prewarm/main.go
@@ -106,6 +106,11 @@ func main() {
 		PrewarmDir: prewarmDir,
 		Images:     make([]manifestImage, 0, len(imagesToWarm)),
 	}
+	p := paths.New(prewarmDir)
+	readyImageManager, err := images.NewManager(p, 1, nil)
+	if err != nil {
+		fatalf("create ready image manager: %v", err)
+	}
 
 	for _, img := range imagesToWarm {
 		entry, err := ensureMirroredImage(ctx, inspectClient, registry, img)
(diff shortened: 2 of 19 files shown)


sjmiller609 (approved):
nice, thanks

PR #320, opened 2026-07-28
Require QEMU for vGPU instances

yummybomb:
## summary

- reject vGPU creates that select, or default to, a non-QEMU hypervisor
- enforce the same invariant when restarting stopped vGPU instances
- cover explicit and default hypervisor selection with parallel pure tests

## why

QEMU is the supported vGPU path. Rejecting unsupported hypervisors before image resolution or device allocation prevents an invalid VM configuration from reaching the host.

## tests

- `go test -race ./lib/instances -run '^TestResolveCreateHypervisorForVGPU$' -count=1`
- `go test ./lib/instances -run '^$' -count=1`
- `go vet ./lib/instances`
- CI Linux and Darwin test suites

The full local VM integration suite requires `mkfs.erofs` and privileged networking; the equivalent CI suite passed.

### DIFF
--- a/lib/instances/create.go
+++ b/lib/instances/create.go
@@ -88,9 +88,10 @@ func (m *manager) createInstance(
 		log.ErrorContext(ctx, "invalid create request", "error", err)
 		return nil, err
 	}
-	hvType := req.Hypervisor
-	if hvType == "" {
-		hvType = m.defaultHypervisor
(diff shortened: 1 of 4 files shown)


yummybomb:
Closing because this invariant is already enforced by the caller and does not need a Hypeman change.

PR #326, opened 2026-07-30
Queue image rebuilds behind in-flight builds instead of dropping them

chruffins:
## Summary

Fixes an image rebuild race where deleting and recreating an image after its ready notification but before its queue slot is released left the recreated image pending forever.

## Root cause

Image builds notify readiness before the queue wrapper releases the active key. A delete/recreate in that window writes fresh pending metadata, then the shared queue deduplicates the new request against the still-active digest and drops it. The original build completes with no successor to start.

A related stale-result race allowed an older build to write terminal metadata or notify readiness after a newer build had replaced its metadata for the same digest.

## Fix

- Add `Queue.EnqueueSuccessor` for callers that intentionally replace work while the previous job is still active.
- Replace an already-pending successor with the newest build closure so repeated recreates cannot retain stale work.
- Keep normal active-key deduplication unchanged for image pushes and other queue users.
- Skip pending successors whose key is still active when promoting queued work.
- Report a pending position when a key has both an active job and a successor.
- Attach a unique build ID to image metadata and reject stale status/finalization writes and notifications.
- Use successor enqueueing for image creation and retry paths.

## Tests

- Queue successor execution and pending successor replacement.
- Promotion skips a successor whose active predecessor has not completed.
- End-to-end local OCI image delete/recreate during the build tail.
- Stale build updates cannot change replacement metadata or unblock `WaitForReady`.
- Repeated delete/recreate requests run the latest pending successor.

Verification:

- `go test -race ./lib/queue`
- `go test -race -run '^TestDeleteAndRecreateDuringBuildTail$' ./lib/images`
- `go test -race ./lib/images` remains for CI; the local environment lacks `mkfs.erofs`, and the development-host rerun was unavailable during this update

---

> [!NOTE]
> **Medium Risk**
> Touches core image build scheduling, metadata writes, and readiness notifications; behavior is constrained by BuildID checks and new queue paths, with broad test coverage for the race scenarios.
> 
> **Overview**
> Fixes a race where **delete + recreate** during the window after an image reports ready but before its digest leaves the active queue slot left the new image **pending forever**, because the queue deduped the recreate against the still-active digest.
> 
> **Queue:** Adds `EnqueueSuccessor` so a replacement job can sit behind an active job with the same key; a pending successor is **replaced** on repeat recreates, promotion **skips** successors whose key is still active, and `GetPosition` reports queue depth when a successor is waiting.
> 
> **Image manager:** Persists a per-build **`BuildID`** and routes new creates through `EnqueueSuccessor`. `updateStatusByDigest` and `finalizeImage` no-op when metadata’s `BuildID` no longer matches, so an older goroutine cannot mark failed/ready or notify `WaitForReady` after a newer build took over. Borrowed-auth lookup takes the expected inflight handle and returns **stale** when credentials were re-registered for a superseded pull.
> 
> **Tests** cover successor queue behavior, end-to-end delete/recreate during the build tail, and stale-build rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10d0b1946e8c8c3193be83ee82aec750426de900. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/images/credentials_test.go
+++ b/lib/images/credentials_test.go
@@ -113,9 +113,28 @@ func TestBorrowedCredentialsExpireWhileQueued(t *testing.T) {
 		return m.inflightPulls[digest].credentials == nil
 	}, time.Second, time.Millisecond)
 
-	credentials, _, expired := m.borrowedAuth(digest)
+	credentials, _, expired, stale := m.borrowedAuth(digest, inflight)
 	assert.True(t, expired)
+	assert.False(t, stale)
 	assert.Nil(t, credentials)
+}
+
+func TestBorrowedAuthRejectsReplacedInflightPull(t *testing.T) {
+	m := &manager{inflightPulls: make(map[string]*inflightImagePull)}
+	const digest = "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+	first := m.registerInflightPull(digest, &authn.AuthConfig{Username: "first"})
+	second := m.registerInflightPull(digest, &authn.AuthConfig{Username: "second"})
+	defer m.releaseInflightPull(digest, second)()
+
+	credentials, _, expired, stale := m.borrowedAuth(digest, first)
+	assert.Nil(t, credentials)
+	assert.False(t, expired)
+	assert.True(t, stale)
+
+	credentials, _, expired, stale = m.borrowedAuth(digest, second)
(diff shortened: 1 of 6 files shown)


sjmiller609 (approved):
- check credential ownership: should each queued closure should read its captured inflight record, not the digest’s newest record?
- Strengthen queue test by asserting successor remains blocked until its same-key predecessor is released.

PR #327, opened 2026-07-30
Parse creatable_vgpu_types in the format the driver emits

yummybomb:
## Summary

Stacked on #321. Validated against an L40S host running vGPU Manager 580.159.01 on Ubuntu 24.04 / kernel 6.8, where the vendor VFIO framework is active and `/sys/class/mdev_bus` has no parents.

`creatable_vgpu_types` is a table with a header row and the type ID first:

```
ID    : vGPU Name
1145  : NVIDIA L40S-1B
1147  : NVIDIA L40S-1Q
...
```

The parser expected the name first and the ID last, so every line failed `strconv.Atoi` on the trailing word:

```
ListGPUProfilesWithVFs: n=0 err=parse vGPU type ID "Name": strconv.Atoi: parsing "Name": invalid syntax
```

`listProfiles` propagates that error, `getVGPUStatus` swallows it, and the host reports vGPU mode with 64 slots and zero profiles — so no vGPU placement can ever succeed.

Parsing on the `:` separator and skipping the header row fixes it. On the same host, profile enumeration now returns all 30 L40S types with correct framebuffer sizes.

There is a second case worth locking in: once a profile consumes a GPU's entire framebuffer, the remaining VFs on that GPU print the header and nothing else. That is a legitimately empty list, not a malformed file, so it now yields zero profiles rather than an error — otherwise a single 48Q allocation would take the whole GPU's remaining capacity out of the status response.

## Tests

- Updated the fixture to the real sysfs format and added a header-only case.
- `go test ./lib/devices/...` passes.
- Verified live: 30 profiles enumerated, availability drops from 64 to 32 slots after allocating a 48Q on one GPU, and back to 64 after release.

### DIFF
--- a/lib/devices/vendor_vfio_linux.go
+++ b/lib/devices/vendor_vfio_linux.go
@@ -331,21 +331,28 @@ func (s vendorVFIOSysfs) vfioDeviceInUse(vfAddress string) (bool, error) {
 	return false, nil
 }
 
+// parseCreatableVGPUTypes reads the driver's creatable_vgpu_types listing:
+//
+//	ID    : vGPU Name
+//	1147  : NVIDIA L40S-1Q
 func parseCreatableVGPUTypes(value string) ([]profileMetadata, error) {
 	profiles := make([]profileMetadata, 0)
 	for lineNumber, line := range strings.Split(value, "\n") {
-		fields := strings.Fields(line)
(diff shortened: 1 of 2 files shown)


yummybomb:
Folded into #321 directly — the parser fix and its live validation are documented in that PR's description.

PR #339, opened 2026-08-03
builder-agent: use a pre-mounted /var/lib/buildkit and bound BuildKit GC

rgarcia:
Extract ensureBuildkitRoot: when the host has already attached a disk at
/var/lib/buildkit the agent uses it directly; otherwise it mounts the same
3 GB tmpfs as before with identical command and error behavior. Mountpoint
detection reads /proc/self/mounts, which reflects the guest's own mount
namespace. If persistent-cache GC bounds are configured, a missing mount
fails fast instead of applying large-disk thresholds to the tmpfs fallback.

Add optional GC bounds to the guest BuildConfig (unused by the host for
now). When both bounds are set, buildkitd.toml gains an OCI worker gcpolicy
with gc=true and quoted human-readable reservedSpace/maxUsedSpace sizes.
Bounds must both be at least 1 MiB, and reservedSpace must be less than
maxUsedSpace. The deprecated gckeepstorage is not emitted: it only maps to
reservedSpace, so nothing would reclaim above the floor and a fixed-size
disk could fill to ENOSPC. Sizes are quoted because BuildKit decodes them
with units.RAMInBytes; a bare integer would be read as bytes.

Tests cover the mount decision (mounted, unmounted, check and mount
failures), /proc/self/mounts parsing, a TOML golden for the worker
section, and a decode regression proving the emitted sizes parse back to
the intended byte counts via docker/go-units (the same decoder BuildKit
uses).

---

---

> [!NOTE]
> **Medium Risk**
> Changes how BuildKit stores cache data and when builds fail (persistent mount required with GC bounds); host wiring is not in this diff, but misconfigured bounds or shared volumes could cause failed builds or ENOSPC without GC.
> 
> **Overview**
> The builder agent no longer always mounts a 3 GB tmpfs at `/var/lib/buildkit`. **`ensureBuildkitRoot`** reuses an existing mount (e.g. host-attached persistent disk) when `/proc/self/mounts` shows one; otherwise it keeps the same tmpfs behavior. If **GC byte bounds** are set but the path is not mounted, the build **fails fast** instead of applying disk-sized GC to tmpfs.
> 
> **`CacheGCReservedBytes`** and **`CacheGCMaxUsedBytes`** are added on guest and host **`BuildConfig`** (host still leaves them zero). When both are set, **`buildkitd.toml`** gets an OCI worker **`gcpolicy`** with quoted `reservedSpace` / `maxUsedSpace` (BuildKit **`units.RAMInBytes`** decoding); validation enforces ≥1 MiB and reserved &lt; max, and **`gckeepstorage`** is omitted on purpose.
> 
> Tests cover mount decisions, mount table parsing, GC TOML output, and size decode regression. **`docker/go-units`** is a direct dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02706f7dccd43ac9b989c10a19f1c4d667fc12cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/digitalocean/go-qemu v0.0.0-20250212194115-ee9b0668d242
 	github.com/distribution/reference v0.6.0
 	github.com/docker/distribution v2.8.3+incompatible
+	github.com/docker/go-units v0.5.0
 	github.com/getkin/kin-openapi v0.133.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-chi/chi/v5 v5.3.0
@@ -85,7 +86,6 @@ require (
 	github.com/docker/docker v28.5.1+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.9.3 // indirect
 	github.com/docker/go-connections v0.6.0 // indirect
-	github.com/docker/go-units v0.5.0 // indirect
 	github.com/dustin/go-humanize v1.0.1 // indirect
 	github.com/fatih/structs v1.1.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
--- a/lib/builds/builder_agent/buildkitroot.go
+++ b/lib/builds/builder_agent/buildkitroot.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// ensureBuildkitRoot prepares BuildKit's data directory. If root is already a
+// mountpoint — e.g. a persistent disk attached by the host — it is used
+// directly. Otherwise a tmpfs is mounted there: the VM rootfs is an overlayfs
+// (read-only ext4 + writable ext4 upper layer) and BuildKit's native overlayfs
+// snapshotter creates char device 0:0 for whiteout markers, but mknod(char 0:0)
+// fails on an overlayfs mount because the kernel treats it as an overlayfs
+// whiteout rather than a regular device node. tmpfs avoids this
+// nested-overlayfs conflict.
+//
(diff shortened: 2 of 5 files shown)


hiroTamada (approved):
approved — the mount detection and tests look solid. two non-blocking edge cases worth considering before the host enables persistent caches:

- `lib/builds/builder_agent/main.go:886-889` — if GC bounds are configured but the persistent mount is missing, this silently falls back to a 3 GB tmpfs with limits intended for a larger disk, which can reach ENOSPC before GC starts. consider failing fast when nonzero bounds imply that the mount must exist.
- `lib/builds/builder_agent/main.go:769-775` — consider rejecting `reservedBytes >= maxUsedBytes` and positive values below 1 MiB; the current conversion can emit a nonsensical policy, including `"0MB"`, from positive inputs.

sjmiller609 (approved):
**builder_agent/main.go**: crosses 1k lines by end of stack, optionally extract some code to buildkitroot.go.

Otherwise, no objections, looks good.

rgarcia:
@hiroTamada addressed both edge cases in 69a4179: configured GC bounds now require an existing `/var/lib/buildkit` mount instead of falling back to tmpfs, and malformed bounds are rejected (both must be at least 1 MiB; reserved must be below max). Added focused mount and validation regressions; `go test -race ./lib/builds/builder_agent -count=3` passes.

rgarcia:
@sjmiller609 took the optional split in 02706f7: BuildKit root setup and mount detection now live in `buildkitroot.go`. The upper Builder integration keeps its version-stamp/reset behavior in that file as well, reducing `main.go` by the root-lifecycle block and keeping that behavior isolated. Focused tests, race tests, vet, and build pass.

PR #341, opened 2026-08-03
api: add /builders CRUD and prune endpoints

rgarcia:
Expose the Builder resource over the API. POST /builders creates a
builder with an optional caller-supplied ID (validated; replay conflicts), optional
non-unique name, tags,
and disk size; the cache disk is provisioned eagerly and 201 returns the
builder. GET /builders lists with tag filtering and always returns a
non-nil array; GET /builders/{id} returns one builder; DELETE
/builders/{id} removes builder and disk (409 while in use); POST
/builders/{id}/prune resets the cache disk asynchronously and returns
202 with status pruning. Responses carry id, name, disk_size_gb, status,
tags, created_at, and last_used_at only. Every builder route documents
the 401 bearer-auth failure response.

Builders resolve by opaque ID through the resource resolver middleware;
builder:read/write/delete scopes gate the routes. The builders manager
is wired through config (builders.max_count, default_disk_size_gb,
max_disk_size_gb, idle_ttl), providers, and the generated injector, and
started with the server so startup reconciliation runs. OpenAPI pins generated enum constant names so adding BuilderStatus does not
rename the existing ImageStatus or BuildStatus API. `stainless.yaml` maps every
Builder endpoint and model into SDK previews.

Handler tests cover create (defaults, positive-size enforcement, custom ID,
invalid ID, conflict replay), get, list with tag filtering, delete, deterministic
prune acceptance, and 409s while a build holds the builder. Config validation
parses idle TTL once, tests negative limits, and startup now propagates
reconciliation failures.

---

---

> [!NOTE]
> **Medium Risk**
> New API manages persistent disks and background prune/delete behavior; misconfiguration of idle_ttl or quota limits could surprise operators, but patterns match existing volume/build resources.
> 
> **Overview**
> Adds a **Builder** HTTP API for persistent build-cache disks: **list/create**, **get/delete by opaque ID**, and **POST prune** (202, async disk reset). Handlers delegate to `builders.Manager`, map domain errors to 400/409/404, and use tag filtering on list like other resources.
> 
> **Wiring:** `BuilderManager` is injected into `ApiService`, Wire, and `ProvideBuilderManager` (limits from new `builders` config: max count, default/max disk GB, optional destructive `idle_ttl`). Startup calls `BuilderManager.Start` for reconciliation/reaper. **Middleware** resolves `/builders/{id}` via `BuilderResolver`; **scopes** `builder:read/write/delete` gate routes.
> 
> OpenAPI/Stainless/oapi codegen adds models, routes, and clients; enum `x-enum-varnames` avoids renaming existing image/build status constants. Config validation and handler/integration tests cover quotas, in-use conflicts, and races with the idle reaper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a41df037b33b2a782482859a16dc05ba25607243. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/api/api/builders.go
```diff
@@ -0,0 +1,208 @@
...
+				Code:    "conflict",
+				Message: "builder is in use by a build",
+			}, nil
+		}
+		log.ErrorContext(ctx, "failed to prune builder", "error", err)
+		return oapi.PruneBuilder500JSONResponse{
+			Code:    "internal_error",
+			Message: "failed to prune builder",
+		}, nil
+	}
```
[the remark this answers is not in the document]
rgarcia: Fixed in 939e6a8 (`api: return 404 from prune when the builder is deleted mid-request`). `PruneBuilder` now maps `builders.ErrNotFound` from `ResetDisk` to the documented `PruneBuilder404JSONResponse`, matching `DeleteBuilder`. Regression test: `TestPruneBuilder_NotFoundAfterResolution` deletes the builder after middleware resolution and asserts a 404. Passing under `go test -race ./cmd/api/api/ -run Builder -count=3`.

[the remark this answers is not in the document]
rgarcia: Follow-up: after the cascade rebase this commit is now efd278a (identical diff).

cmd/api/api/builders.go
```diff
@@ -0,0 +1,215 @@
...
+				Code:    "invalid_request",
+				Message: err.Error(),
+			}, nil
+		}
+		log.ErrorContext(ctx, "failed to create builder", "error", err)
+		return oapi.CreateBuilder500JSONResponse{
+			Code:    "internal_error",
+			Message: "failed to create builder",
+		}, nil
+	}
```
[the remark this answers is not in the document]
rgarcia: Fixed in fb3c065 (`api: return 400 when builder disk size exceeds the configured max`) on top of 79b990b, which makes the manager return a wrapped `builders.ErrDiskSizeExceeded` sentinel. `CreateBuilder` now maps it to `CreateBuilder400JSONResponse`. Regression test: `TestCreateBuilder_DiskSizeExceeded` (400 with the limit in the message); domain-side assertion added to `TestCreateBuilder_Quotas` (`ErrorIs ErrDiskSizeExceeded`). Passing under `go test -race ./cmd/api/api/ -run Builder -count=3` and `go test -race ./lib/builders/`.

cmd/api/api/builders.go
```diff
@@ -0,0 +1,220 @@
...
+			return oapi.DeleteBuilder404JSONResponse{
+				Code:    "not_found",
+				Message: "builder not found",
+			}, nil
+		}
+		if errors.Is(err, builders.ErrInUse) {
+			return oapi.DeleteBuilder409JSONResponse{
+				Code:    "conflict",
+				Message: "builder is in use by a build",
+			}, nil
```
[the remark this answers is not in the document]
rgarcia: Fixed in 00abce8 (`api: use a generic in-use message for builder conflicts`). Both delete and prune 409s now return "builder is in use" since `builders.ErrInUse` also covers pruning, deleting, and an attached disk; the openapi.yaml conflict descriptions were updated to match and oapi.go regenerated. `go test ./cmd/api/api/ -run Builder` passes.

hiroTamada (commented):
reviewed — overall this looks good. a few suggestions worth considering:

### API shape

- `openapi.yaml:1836-1888` — consider modeling this as a build-cache resource (`/build-caches`, `cache_id`) since the durable resource is the cache disk while builder VMs are ephemeral. alternatively, nesting the cache configuration under `cache` would leave room for future builder settings.

### Implementation

- `openapi.yaml:1882-1885` / `cmd/api/api/builders.go:51-52` — an explicit `disk_size_gb` of zero or a negative value is treated as omission and replaced with the default. consider adding `minimum: 1` and returning `400` for non-positive values.
- `cmd/api/api/builders.go:156-178` — `ResetDisk` starts its background worker before the handler re-reads the builder, so the worker may finish first and the `202` response may contain `ready`. returning the accepted transition snapshot directly would make the response and test deterministic.
- `openapi.yaml:1836` / `lib/oapi/oapi.go:145-150` — adding `BuilderStatus` renames the existing `ImageStatusReady`, `ImageStatusFailed`, etc. constants to package-global `Ready`, `Failed`, and so on. consider pinning stable enum names so adding this schema does not change the existing generated API.

### Nits

- `cmd/api/api/api_test.go:137-146` — the `ctxWithVolume` comment now documents `ctxWithBuilder`; add the builder comment and move the volume comment back.
- `cmd/api/config/config_test.go:549-555` — consider adding a negative `MaxDiskSizeGb` case for the corresponding validation branch.

openapi.yaml
```diff
@@ -1833,6 +1833,59 @@ components:
...
+      enum: [ready, pruning, deleting, error]
+      description: Builder lifecycle status
+
+    Builder:
+      type: object
+      required: [id, disk_size_gb, status, created_at]
+      properties:
+        id:
+          type: string
+          description: Unique builder identifier (opaque; the only selector)
```
sjmiller609:
```suggestion
          description: Builder identifier
```

openapi.yaml
```diff
@@ -1833,6 +1833,59 @@ components:
...
+          example: "bldr-abc123"
+        name:
+          type: string
+          description: Optional non-unique display name
+          example: team-cache
+        disk_size_gb:
+          type: integer
+          description: Persistent BuildKit cache disk size in gigabytes (immutable)
```
sjmiller609:
```suggestion
          description: Persistent builder cache disk size in gigabytes. Cannot be changed after creation.
```

openapi.yaml
```diff
@@ -1833,6 +1833,59 @@ components:
...
+          format: date-time
+          description: When a build last ran on this builder
+          nullable: true
+
+    CreateBuilderRequest:
+      type: object
+      properties:
+        id:
+          type: string
+          description: Optional caller-supplied identifier for control-plane idempotency (auto-generated if not provided)
```
sjmiller609:
```suggestion
          description: Optional caller-supplied identifier for idempotency, auto-generated if not provided
```
also note need to make it actually idempotent if going with this description

openapi.yaml
```diff
@@ -1833,6 +1833,59 @@ components:
...
+          example: "team-cache-1"
+        name:
+          type: string
+          description: Optional non-unique display name
+          example: team-cache
+        disk_size_gb:
+          type: integer
+          description: Cache disk size in gigabytes (default from server config; immutable after creation)
```
sjmiller609:
```suggestion
          description: Cache disk size in gigabytes
```

openapi.yaml
```diff
@@ -4333,3 +4386,207 @@ paths:
...
+                $ref: "#/components/schemas/Error"
+        500:
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+    post:
+      summary: Create builder
+      description: Creates a builder and eagerly provisions its persistent BuildKit cache disk. One build at a time runs on a builder.
```
sjmiller609:
```suggestion
      description: Creates a builder its cache disk. One build at a time runs per builder.
```

openapi.yaml
```diff
@@ -4333,3 +4386,207 @@ paths:
...
+                $ref: "#/components/schemas/Error"
+        500:
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+    delete:
+      summary: Delete builder
+      description: Deletes a builder and its cache disk. Deletion is total and irreversible.
```
sjmiller609:
```suggestion
      description: Permanently deletes a builder and its cache disk.
```

openapi.yaml
```diff
@@ -4333,3 +4386,207 @@ paths:
...
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /builders/{id}/prune:
+    post:
+      summary: Prune builder cache
+      description: Resets the builder's cache by asynchronously recreating its disk. The builder transitions to pruning, then ready. Builder identity is preserved.
```
sjmiller609:
```suggestion
      description: Resets the builder's cache disk.
```

sjmiller609 (commented):
Some in-line comments for clean up of wording in API

### bugs
- `stainless.yaml` — never updated; `/builders` endpoints missing from SDK mapping stack-wide
- `builders.go:58` — oversize `disk_size_gb` returns 500; needs sentinel error + 400 arm (mentioned in #340 already)

### abstract builder tooling
- optionally add `type` field (default `"buildkit"`) now
- schema descriptions hardcode "BuildKit cache disk"; keep wording generic

### findings
- `openapi.yaml` — `id` claims "idempotency" but retry returns 409
- `builders.go:161` — prune re-reads builder; `ResetDisk` should return it (race → spurious 500)

### nits
- `openapi.yaml` — `disk_size_gb` lacks `minimum: 1`; could protect against in schema
- 409 message "in use by a build" wrong when builder is mid-prune (`ErrInUse` conflation)
- `id` example `"bldr-abc123"` — generated IDs have no prefix, make example look like it actually looks
- generic ptr helpers (`valueOrEmpty` etc.) stranded in `builders.go`
- `idle_ttl` parsed twice (Validate + provider); can drift
- `BuilderManager.Start` error branch is dead code; reconcile failures swallowed
- default disk size defaulted in two places (config + manager)

openapi.yaml
rgarcia: Removed the idempotency claim in d69245d. This remains an optional caller-supplied identifier, and replaying an existing ID intentionally returns 409.

openapi.yaml
rgarcia: Updated in d69245d: the response now describes the persistent builder cache disk and explicitly states that its size is immutable.

openapi.yaml
rgarcia: Updated the identifier description and example in d69245d.

openapi.yaml
rgarcia: Updated in d69245d and added `minimum: 1`; omission uses the server default.

openapi.yaml
rgarcia: Updated the delete description in d69245d.

openapi.yaml
rgarcia: Updated the prune description in d69245d while retaining the lifecycle transition detail.

openapi.yaml
rgarcia: Updated the create description in d69245d.

rgarcia:
@hiroTamada @sjmiller609 addressed the API/config feedback in d69245d:

- added all Builder endpoints/models to `stainless.yaml`
- rejects explicit zero/negative `disk_size_gb`, adds `minimum: 1`, and preserves the typed oversized-disk 400 path
- makes `ResetDisk` return the accepted `pruning` snapshot, removing the handler re-read race
- pins existing and new generated enum constant names so adding `BuilderStatus` cannot rename `ImageStatus` or `BuildStatus`
- removes the incorrect idempotency claim and uses a generated-ID-shaped example
- makes Builder/cache descriptions backend-neutral and applies the inline wording cleanups
- fixes the test-helper comment and adds negative max-disk coverage
- parses idle TTL once during config validation and propagates reconciliation failures through `BuilderManager.Start`
- removes the file-local pointer helpers

I kept the `Builder` resource shape rather than renaming it to build cache: the resource is the stable build-execution/cache identity even though its current VMs are disposable. I did not add a `type` field with only one legal value; a backend discriminator can be added compatibly when a second implementation exists. The manager retains its own default as a safe domain-level default for non-API callers, while API config controls the deployed value.

sjmiller609 (approved):

PR #342, opened 2026-08-03
builds: run builds on Builders with persistent cache disks

rgarcia:
Add optional builder_id to CreateBuild and the Build response; it is the
only builder selector. A build with builder_id acquires the builder
before attaching, attaches its disk read-write at the exact
/var/lib/buildkit path through the internal AllowSystemVolumeMounts
carveout (now also covering the mount path), boots the disposable builder
VM, and releases the builder after VM execution ends and before the
image-conversion wait, stamping last_used_at even on failed builds.
Builds for one builder are serialized by builder ID in the queue: a
waiting same-builder build stays pending without consuming a global
concurrency slot. Scheduling may skip a serially blocked build, while reported
queue positions remain physical submission order. Builds without builder_id
keep the exact tmpfs and queue behavior.

The builder API responses gain max_concurrency (fixed at 1),
active_build_id, and queued_builds computed from the queue, and builder
delete/prune now also reject while builds for the builder are queued or
running, via an activity checker the build manager registers. The host
computes BuildKit GC bounds from the disk size into the guest config
(reservedSpace 70%, maxUsedSpace 90%). The builder agent stamps the
persistent root with the buildkitd version and empties the root on
incompatibility before buildkitd starts; a fresh disk (no stamp, nothing
but lost+found) is stamped without a wipe. Cache contents are
best-effort, and no secrets or source/config volumes live on the builder
disk.

Stale holder recovery happens at acquire: DeleteInstance's volume detach
is warn-only, so a crashed build can leak an attachment record and leave
its builder VM holding the disk. AcquireForBuild clears those stale
attachments before marking the builder held (deleting the stale VM,
tolerating its absence, and detaching surviving records), so the next
build self-heals instead of failing until a restart runs reconciliation.
The builds tests wire the builds and builders managers to one shared
volume manager, matching production.

Tests cover the disk attach lifecycle and cleanup, the unchanged
no-builder path asserted exactly, create-time builder validation, GC
bound computation, queue serialization and introspection, stale holder
recovery success and loud failure through the real acquire path,
delete/prune rejection while builds are queued or running, and the
version stamp transitions.

Add `lib/builders/README.md` to document the user-visible cache identity,
build execution, concurrency, lifecycle, recovery, and durability model.

### Follow-up (not blocking)

Add a CI fixture that boots the generic Builder image in a real VM and proves
cache reuse across two disposable VMs plus stale-holder recovery after a VM
crash. The current suite covers these host/agent lifecycle invariants under the
race detector but does not execute BuildKit end to end.

---

---

> [!NOTE]
> **Medium Risk**
> Touches build scheduling, VM volume attachment, and builder lifecycle; mistakes could leak holds, corrupt cache, or allow unsafe delete/prune, but behavior is heavily tested and fails closed on activity checks.
> 
> **Overview**
> Adds optional **`builder_id`** on create-build and **`builder_id`** on build responses so jobs can reuse a Builder’s persistent cache disk instead of ephemeral tmpfs. Builds without it keep the previous behavior.
> 
> **Execution and queue:** Same-builder jobs are serialized via queue **serial keys** (pending work does not consume a global slot while waiting). After VM work finishes, **`ReleaseSerialKey`** lets the next job start while post-build steps run. The host acquires the builder, mounts its disk at `/var/lib/buildkit` (with system-mount carveout), computes BuildKit GC bounds from disk size, and releases with retries; stale VM/disk holders are cleared on acquire.
> 
> **API and lifecycle:** Builder payloads expose **`max_concurrency`** (1), **`active_build_id`**, and **`queued_builds`**. Create-build returns **404/409** for missing or busy builders. Delete/prune/idle-reaper consult a **build activity checker** so queued or disk-pending work blocks teardown. The builder agent **version-stamps** persistent roots and wipes cache on BuildKit mismatch.
> 
> Wiring passes **`builders.Manager`** into the build manager and registers the activity checker. Adds **`lib/builders/README.md`** and broad tests for queue, disk lifecycle, and recovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 124184a801ff2351998f83d9bd6027738e0a30d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/builds/manager.go
```diff
@@ -530,6 +563,39 @@ func (m *manager) CreateBuild(ctx context.Context, req CreateBuildRequest, sourc
...
+// builder, oldest first.
+func (m *manager) QueuedBuildsForBuilder(builderID string) []string {
+	return m.queue.PendingBuildsForSerialKey(builderID)
+}
+
+// BuilderHasBuilds reports whether any build targeting the builder is
+// queued or running.
+func (m *manager) BuilderHasBuilds(builderID string) bool {
+	return m.ActiveBuildForBuilder(builderID) != nil || len(m.QueuedBuildsForBuilder(builderID)) > 0
+}
```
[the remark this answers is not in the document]
rgarcia: Fixed in 6e8347d (`builds: harden builder release and pre-recovery activity checks`). `BuilderHasBuilds` now also matches persisted pending builds (`StatusQueued`/`StatusBuilding`/`StatusPushing` with a `builder_id`) until startup recovery completes, and fails closed (reports activity) when the listing errors, so delete/prune/idle-reaper can't remove a builder out from under disk-pending builds. `RecoverPendingBuilds` sets the recovered flag only after a successful re-enqueue pass. Regression test: `TestBuilderHasBuilds_CountsDiskPendingBeforeRecovery`.

lib/builds/manager.go
```diff
@@ -564,6 +630,11 @@ func (m *manager) runBuild(ctx context.Context, id string, req CreateBuildReques
 	// Run the build in a builder VM
 	result, err := m.executeBuild(buildCtx, id, req, policy)
 
+	// executeBuild's defers release the builder acquisition when it returns,
+	// so same-builder builds may start while this build waits for image
+	// conversion and re-tagging below.
+	m.queue.ReleaseSerialKey(id)
```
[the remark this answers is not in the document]
rgarcia: Fixed in 6e8347d (`builds: harden builder release and pre-recovery activity checks`). The deferred release in `executeBuild` now retries `ReleaseBuild` up to 5 times (1s apart) before giving up: `ReleaseBuild` keeps the hold on persist failure precisely so the caller can retry, and without the retry a transient metadata write error leaked the acquisition and bricked later builds for that builder until restart. Regression test: `TestExecuteBuild_ReleaseRetriedOnPersistFailure` occupies the metadata temp path for the first attempts and asserts the retried release clears the hold and stamps last_used_at.

lib/builds/manager.go
```diff
@@ -564,6 +661,11 @@ func (m *manager) runBuild(ctx context.Context, id string, req CreateBuildReques
...
```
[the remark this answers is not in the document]
rgarcia: Fixed in 783be9b (`builds: start serialized successor on serial key release`). `ReleaseSerialKey` now starts the pending build holding the same serial key directly instead of going through the global-slot gate: the releasing build is past its exclusive VM phase, so the successor overlaps post-build work even at `MaxConcurrentBuilds=1`. Regression test: `TestBuildQueue_ReleaseSerialKeyStartsSuccessorAtFullCapacity` (passes under `-race -count=3`).

openapi.yaml
```diff
@@ -1867,6 +1871,19 @@ components:
           format: date-time
           description: When a build last ran on this builder
           nullable: true
+        max_concurrency:
+          type: integer
+          description: Maximum concurrent builds on this builder (fixed at 1)
```
sjmiller609:
```suggestion
          description: Maximum concurrent builds on this builder
```

sjmiller609: why have this in the API at all if we only support 1?

openapi.yaml
```diff
@@ -1867,6 +1871,19 @@ components:
...
+          example: 1
+        active_build_id:
+          type: string
+          description: ID of the build currently running on this builder
+          nullable: true
+        queued_builds:
+          type: array
+          description: IDs of queued builds waiting for this builder, oldest first
+          items:
+            type: string
```
sjmiller609, line 1890: recommend limit queue size to N to avoid needing to handle issues here (deny admission beyond N)

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
+# Builders
```
sjmiller609, line 1: Suggestions for what could be the e2e tests, I'm thinking one happy-path and one crash-recovery path.

- Watch out for dependency on pulling from dockerhub, address in similar fashion to other tests
- Confirm tests actually run + pass in CI, as that should be possible and is what protects against future regression

**test 1 — cache reuse (happy path)**
1. create Builder
2. build 1 with `RUN head -c16 /dev/urandom | base64 > /stamp`, `builder_id` set — real VM
3. assert success, builder `ready`, `last_used_at` set
4. build 2, identical Dockerfile, same builder
5. assert layer digest identical to 1 (or `CACHED` in logs) — proves cache survived across VMs
6. delete builder, assert volume gone

**test 2 — crash recovery**
1. create Builder
2. start build with `RUN tail -f /dev/null` (blocks forever)
3. on the build's `building` SSE event, delete the VM directly — simulates crash, leaks the attachment
4. run a normal build on the same builder
5. assert it succeeds — proves stale-holder cleanup + dirty-root tolerance

Not sure if you want something for the case of hypeman restarting while builds are occurring + queued. Also, ideally these can all run in parallel to each other, it seems they should be able to in concept.

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Builders give repeated Hypeman builds an explicit, persistent BuildKit cache.
+Without a Builder, every build uses a new 3 GB in-memory BuildKit root and loses
+its local cache when the build VM exits. With a Builder, builds reuse the same
+cache disk across otherwise disposable build VMs.
```
sjmiller609: Should the default situation end up using builders, e.g. like "use the default builder by default" or something like that?

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+intermediate layers, and build metadata, reducing subsequent build time and
+registry traffic.
+
+## Identity and isolation
+
+Each Builder has an opaque ID and owns one cache disk. Pass that ID as
+`builder_id` when creating a build. The ID—not the optional display name, API
+credential, or caller identity—is the cache boundary.
+
+Names and tags are organizational metadata. Names are optional and are not
```
sjmiller609: "organizational metadata" is leaking external use case into hypeman documentation

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+unique, so automation should always store and use the Builder ID.
+
+Omitting `builder_id` preserves the original build behavior: the build runs with
+a temporary 3 GB BuildKit root and does not retain local BuildKit state.
+
+## Build behavior
+
+A build that targets a Builder still runs in a new disposable VM. Hypeman
+attaches the Builder's disk at `/var/lib/buildkit` for the duration of the build,
```
sjmiller609: maybe leave it generic for potential future builder types

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+then detaches it when VM execution ends. Source archives, build secrets, and VM
+configuration are not stored on the Builder disk.
+
+One build runs on a Builder at a time. Additional builds for the same Builder
+wait in submission order without consuming a global build-execution slot. Use
```
sjmiller609: weird phrasing
```suggestion
One build runs on a Builder at a time. Additional builds for the same Builder
wait in submission order. Use
```

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+separate Builders when independent build streams need to run concurrently or
+must not share cache state.
+
+Builder details report the active build, queued build IDs, and
+`max_concurrency`, which is fixed at `1`.
+
+## Cache lifecycle
+
+Creating a Builder eagerly creates its fixed-size ext4 cache disk. Disk size is
+immutable; create a new Builder to use a different size.
```
sjmiller609:
```suggestion
Creating a Builder creates a fixed-size cache disk.
```

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+BuildKit garbage collection is derived from the disk size and reclaims cache
+before the disk fills. Cache contents are an optimization, not durable build
+output: Hypeman may recreate an empty disk after missing storage, interrupted
+cleanup, or an incompatible BuildKit version.
```
sjmiller609: this is hard to understand and also phrased buildkit specific

lib/builders/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Operational model
+
+Builders are local to the Hypeman server that owns their disk. Their metadata
+survives process restarts, and startup recovery reconciles interrupted prune or
+delete operations and stale VM attachments.
+
+The initial implementation deliberately keeps build VMs disposable. It does not
+provide warm Builder VMs, concurrent BuildKit solves on one Builder, cache
+migration between servers, or a durability guarantee for cached data.
```
sjmiller609:
```suggestion

## Hypeman restarts

Builders' metadata survives Hypeman process restarts, and startup recovery 
reconciles interrupted prune or delete operations and stale VM attachments.
Reconciliation may restart some builds that were in progress during the restart.
```

sjmiller609 (commented):
**bugs**
* manager.go:597 — BuilderHasBuilds takes two separate queue locks; pending→active builds invisible mid-transition
* delete/prune guard relies on that racy predicate; builder deletable under a starting build

**findings**
* no e2e test: nothing boots a VM, runs buildkitd on ext4, verifies a cache hit. At least good e2e functional test should exist for each new major feature. Made some suggestions about this in one of the review comments.
* manager.go:407 — builder status unchecked at create; doomed builds 201 then fail async at acquire
* main.go:955 — stamp is full buildkitd --version output; sha-only rebuild wipes all caches (intended?)
* serialization invariant assembled from defer ordering across three functions; extract builderLease
* SetBuildActivityChecker setter-injection dodges dependency cycle; hoist check to handler, delete callback
abstract builder tooling
* /var/lib/buildkit literal in three packages (agent, builds, instances); define once

**nits**
* buildSerialKey returns req.BuilderID verbatim; inline it
* activeSerialKeys keyed backwards; map[key]buildID gives O(1) lookups
* canStartLocked inlines serialKeyFreeLocked's scan; call the helper
* queue positions renumber for plain builds too; note in PR body
* no -race stress test on queue serialization invariant
* mock volume manager re-implements real semantics; builders tests use the real one
* resetRoot deletes lost+found; freshDisk preserves it — asymmetric
* 70%/90% GC split undocumented for users sizing disks
* builder_instance_id vs builder_id adjacent with different meanings; clarify descriptions
* active_build_id/queued_builds are point-in-time snapshots; say so in schema
* delete/prune-blocked test anchors current wiring; moves if check hoists to handler

lib/builders/README.md
rgarcia: Agreed that a real VM cache-reuse/crash-recovery test is valuable. This stack currently covers the host lifecycle, mounted-root agent behavior, serialization, and stale-holder recovery under `-race`, but does not have a CI fixture that boots the generic Builder image and runs BuildKit end to end. I’m keeping that as explicit rollout follow-up rather than adding a test that skips when the image/VM fixture is unavailable.

openapi.yaml
rgarcia: Keeping queue admission unchanged in this stack. An arbitrary per-Builder N would add a new capacity policy and create-time error contract; that should be configured alongside a host-wide pending-build limit rather than embedded here. Queue positions now retain the existing physical submission-order semantics even when scheduling skips a serially blocked build.

lib/builders/README.md
rgarcia: Updated in 156c878. The user-facing build behavior is backend-neutral and no longer exposes the internal mount path.

lib/builders/README.md
rgarcia: Removed “organizational metadata” in 156c878. The README now states the concrete behavior: names are optional/non-unique and tags support filtering.

openapi.yaml
rgarcia: Updated in 156c878. I kept the field as a read-only capability signal so clients do not hardcode today’s limit and can consume a future higher value without an API shape change; the description now says the maximum is currently fixed at 1.

lib/builders/README.md
rgarcia: Applied the simpler wording in 156c878.

lib/builders/README.md
rgarcia: No default Builder in this version. That is now explicit in the README: omitting `builder_id` preserves the existing temporary-cache behavior, while persistent cache ownership always requires an explicit opaque Builder ID.

lib/builders/README.md
rgarcia: Applied in 156c878.

lib/builders/README.md
rgarcia: Applied the restart section in 156c878, including that queued or running builds may be started again during recovery.

lib/builders/README.md
rgarcia: Rewritten in 156c878. The lifecycle section now explains the user-visible 70% protected-cache floor and 90% GC ceiling without exposing BuildKit configuration names.

rgarcia:
@sjmiller609 addressed the new integration review in 156c878 (plus the lifecycle fix in #340 at 3ae855e):

- `BuilderHasBuilds` now checks active and pending serial-key state under one queue lock, eliminating the pending→active deletion window; added a repeated `-race` transition test
- create rejects non-ready Builders before metadata/queue admission
- BuildKit compatibility stamps use the semantic release, so a same-version rebuild SHA does not wipe cache
- cache reset preserves `lost+found`
- queue scheduling still skips blocked serial keys, but reported positions retain physical submission order
- removed the trivial serial-key helper and reused `serialKeyFreeLocked`
- clarified `builder_instance_id` vs `builder_id`, and marked activity/queue fields as point-in-time
- rewrote the README to be backend-neutral, make explicit that there is no default Builder, document the 70%/90% cache policy, and describe restart recovery

I kept `max_concurrency` as a read-only capability signal; clients can consume a future value without a schema change. I also kept `SetBuildActivityChecker`: the domain idle reaper needs queue/disk-pending activity, so checking only in HTTP handlers would leave destructive background deletion racy. The instances package no longer hardcodes `/var/lib/buildkit`; the remaining host/guest literals are on opposite sides of the VM boundary.

A real-VM BuildKit cache-hit/crash-recovery CI fixture is now called out explicitly in the PR follow-up. I did not add an arbitrary per-Builder queue cap in this stack because that needs a host-wide admission policy and a new create-build error contract.

cmd/api/api/builds.go
```diff
@@ -260,6 +271,11 @@ func (s *ApiService) CreateBuild(ctx context.Context, request oapi.CreateBuildRe
 				Code:    "invalid_source",
 				Message: err.Error(),
 			}, nil
+		case errors.Is(err, builders.ErrNotFound):
+			return oapi.CreateBuild404JSONResponse{
+				Code:    "not_found",
+				Message: "builder not found",
+			}, nil
```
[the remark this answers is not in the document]
rgarcia: Fixed in 124184a. Create-build now maps `builders.ErrInUse` to a documented 409 conflict, with a direct multipart handler regression test.

lib/builds/manager.go
```diff
@@ -530,6 +578,51 @@ func (m *manager) CreateBuild(ctx context.Context, req CreateBuildRequest, sourc
...
+		m.logger.Error("list pending builds for builder activity check", "builder", builderID, "error", err)
+		return true
+	}
+	for _, meta := range pending {
+		if meta.Request != nil && meta.Request.BuilderID == builderID {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
rgarcia: This is intentional and safe. `ReleaseSerialKey` happens only after `executeBuild` returns: the disposable VM is deleted, its cache disk is detached, and the Builder acquisition is released. Image conversion/re-tagging operates on the published output image and does not read or write the Builder cache. Releasing here is what allows the next same-Builder VM phase to overlap post-build image work; delete/prune are also safe once the disk is detached.

sjmiller609 (approved):
> A real-VM BuildKit cache-hit/crash-recovery CI fixture is now called out explicitly in the PR follow-up

Ok sounds good, and I do think the follow up is actually warranted

PR #359, opened 2026-08-06
Return invalid request for macOS vGPU

pbjaru:
## Summary

- return HTTP 400 `invalid_request` when a macOS instance request includes a vGPU
- reject the request before checking GPU capacity, so it is not reported as insufficient resources
- leave other GPU creation errors unchanged

## Validation

- Linux and macOS tests pass in CI
- local build, vet, and focused error-response tests pass

## Scope

This only changes instance creation. Starting an existing instance is unchanged.

---

> [!NOTE]
> **Low Risk**
> Scoped to instance-create validation and error mapping on macOS; Linux vGPU behavior unchanged and existing start paths are untouched.
> 
> **Overview**
> **Instance create** now treats vGPU profile requests on hosts without mediated GPU support as a client error instead of a generic or capacity failure.
> 
> Adds `devices.Capabilities()` (Linux reports vGPU support; macOS does not) and moves `ErrVGPUNotSupportedOnMacOS` to a shared sentinel in `devices/errors.go`. **Create** rejects a non-empty GPU profile when `SupportsVGPU` is false **before** `ReserveAllocation`, so macOS no longer surfaces this as insufficient resources. If `CreateMdev` still returns the macOS unsupported error, `wrapCreateMdevErr` wraps it as `ErrInvalidRequest`, which the API maps to HTTP **400** with code `invalid_request` and the message `invalid request: vGPU (mdev) is not supported on macOS`.
> 
> Tests cover platform capabilities, early rejection (no reservation), error wrapping, and the create-instance error mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9034ff4c2ec3131505dac88ca3cdba5c6e496004. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/api/api/instances_test.go
+++ b/cmd/api/api/instances_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/c2h5oh/datasize"
 	"github.com/kernel/hypeman/lib/autostandby"
+	"github.com/kernel/hypeman/lib/devices"
 	"github.com/kernel/hypeman/lib/healthcheck"
 	"github.com/kernel/hypeman/lib/hypervisor"
 	"github.com/kernel/hypeman/lib/images"
@@ -636,7 +637,7 @@ func TestInstanceToOAPI_OmitsPlatformWhenUnset(t *testing.T) {
 }
 
 // errCreateInstanceManager is a fake whose CreateInstance always fails with a
-// preset error, used to assert the handler maps typed image errors to statuses.
+// preset error, used to assert the handler maps typed errors to statuses.
 type errCreateInstanceManager struct {
 	instances.Manager
 	err error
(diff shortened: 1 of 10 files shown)


sjmiller609 (commented):
`lib/instances/create.go:100` — could we model unsupported vGPU as a device/host `Capability` (find existing pattern in codebase) rather than branching on `runtime.GOOS` here? something like `devices.Capabilities().SupportsVGPU`, implemented per platform, is the more standard way to handle platform specific capabilities

pbjaru:
good callout - implemented the existing pattern we rely on in the codebase currently!

sjmiller609 (approved):
perfect, thanks!

PR #364, opened 2026-08-06
Add the vendor VFIO vGPU device backend

yummybomb:
## Summary

Layer 2 of the vendor VFIO vGPU stack (`generalize-vgpu-device` ← `hypervisor-liveness` ← **this** ← `vendor-vfio-vgpu`). Self-contained in `lib/devices` + `lib/resources`; nothing in the instance lifecycle calls it yet (that's the top layer).

Linux 6.8 hosts with NVIDIA R580 drop the mdev interface: vGPUs are assigned by writing a type ID to a VF's `nvidia/current_vgpu_type` and passed to QEMU as a plain VFIO PCI device. This adds that backend behind the framework dispatch introduced in #322:

- **Discovery & placement** — profile discovery from the capacity-dependent `creatable_vgpu_types` catalogs, least-loaded-GPU VF selection, create/verify/rollback. Profile availability counts free VFs currently advertising each type as a best-effort snapshot; creating one assignment may change sibling catalogs.
- **Release guards** — unlike mdev (fresh UUID per assignment), vendor VFIO reuses the same VF path across assignments, so a stale release could clear a later owner's vGPU. Release is guarded by an in-process owner map (covers the window before QEMU opens the device) and an open-VFIO-handle scan (refuses to clear a VF a running VM holds).
- **Reconciliation** — clears orphaned assignments on startup, skips VFs in the caller-supplied protected set, and fails closed (skips vendor VFIO entirely) when that set is unavailable, while mdev reconciliation still runs.
- **Per-VF degradation** — one unreadable VF is skipped with a warning instead of failing discovery or profile listing wholesale, so a flaky sysfs read cannot blank the host's advertised GPU capacity. Only when no VF is readable does discovery fail, so a wholesale outage cannot demote a vGPU host to passthrough while assignments exist. The open-handle probe stays strict on purpose: it authorizes clearing a reused VF path, so an incomplete scan fails the release rather than risk a false "not in use".
- **Integration test** — branched by discovered framework, extended to cover release on stop and reacquisition on start.

## Testing

- `go build ./...`, `go vet` clean
- `go test -race ./lib/devices/ ./lib/resources/` pass

---

> [!NOTE]
> **Medium Risk**
> Changes GPU discovery, capacity reporting, and sysfs assignment/release on the host; vendor VFIO create remains disabled but destroy/reconcile logic is security-sensitive around VF reuse.
> 
> **Overview**
> Adds **NVIDIA vendor VFIO vGPU** support alongside mdev: hosts are classified via `DiscoverVGPU()` (mdev first, then VFs with `nvidia/creatable_vgpu_types` / `current_vgpu_type`). A new Linux backend handles profile listing, least-loaded VF placement, assignment by writing the type ID, guarded release (instance ownership + open VFIO FD scan), and startup reconciliation of orphaned assignments.
> 
> **`CreateVGPU` still rejects vendor VFIO** until instance lifecycle integration lands; destroy/reconcile paths and types (`VGPUFramework`, `GPUDevicePath`, `InstanceID` on assignments) are wired for the next layer. Mdev discovery is tightened (skip unreadable VFs; fail only if none are readable). **`DetectHostGPUMode` is removed**; `/resources` GPU status uses `DiscoverVGPU(ctx)` and avoids mis-reporting passthrough on transient vGPU probe failures.
> 
> The vGPU integration test is framework-aware (sysfs checks for mdev vs vendor VFIO), adds **stop → release** and **start → reacquire**, and skips vendor VFIO until create is enabled. Docs/OpenAPI clarify dual frameworks, lifecycle, QEMU-only vGPU, and vendor VFIO troubleshooting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17fd8e8917b3bc06288aacdd0b8451b76d829da9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


yummybomb:
added 7f5233f: report `GPUProfile.Available` as the count of free VFs advertising the profile type (creatable-instance units), matching mdev's summed `available_instances`, the OpenAPI description, and the integration test's decrement assertion. note the `gpuProfileSlots` metric steps up on vendor VFIO hosts (per-GPU → per-free-VF units). `go test ./lib/devices ./lib/resources` green; the hardware integration test was not run locally.

sjmiller609 (approved):
- mdev discovery now hard-fails on one unreadable mdev_supported_types, zeroing host GPU capacity — vendor VFIO degrades per-VF. intentional asymmetry? Also that is different behavior than before the VFIO changes.
- reconcile silently skips an allocated + in-use + unprotected VF. log at warn level.

PR #373, opened 2026-08-07
imagepush: review fixes for push job manager (#350)

chruffins:
Addresses review findings on #350 in `lib/imagepush`. Stacked on #350 (targets its head branch).

## Changes
- **credentials-blind dedup is now a hard conflict.** `CreatePush` merges a request into an in-flight job only when credential presence matches. A request that borrowed credentials can no longer merge into an anonymous in-flight push (auth silently dropped), and an anonymous request can no longer merge into a credentialed one (silently inherits another caller's login). Mismatch returns new `ErrCredentialConflict`. The manager never stores credential values, so presence is the only comparable signal; two requests borrowing different credentials for the same target remain indistinguishable and merge (documented).
- **dropped the dead orphan-adoption path.** The per-`CreatePush` `PushesDir` scan (`findPendingPush`) was unreachable: startup recovery adopts every pending record before any `CreatePush` runs, and the create lock spans metadata write + inflight registration, so no pending record can exist on disk that isn't tracked. Removing it also removes the O(N) scan performed while holding the create lock. Adjusts/removes the three orphan-adoption tests.
- **corrupt metadata is surfaced, not swallowed.** `listAllPushes` logs a warning for unreadable records instead of silently dropping them from listing and recovery.
- plus: empty-digest guard in `CreatePush` (dedup key would otherwise collide across images), `ctx` cancellation honored in `GetPush`/`ListPushes`, and clarifying comments for the best-effort pushing-status write and the terminal-persist-before-inflight-release invariant.

## Tests
- new `TestCreatePushCredentialConflict` (both directions), `TestCreatePushDedupesConcurrently` (the previously-missing concurrent dedup race), `TestListPushesSkipsUnreadableMetadata`.
- `go test -tags containers_image_openpgp -race -count=2 ./lib/imagepush/...` — pass. `go build -tags containers_image_openpgp ./...` and `go vet` — clean. `lib/registrypush`, `lib/ocicache`, `lib/registry`, `lib/ocicachegc` also pass with `-race`.

Not addressed here (left as follow-ups): consolidating `pushQueue` with the builds/images queues, decomposition of the 449-line manager, and a retention policy for terminal records.

Note: the full lib/images suite has a pre-existing failure (`TestImportLocalImageFromOCICache` times out spinning up a builder VM at the 60s cap); it is untouched by this change.

---

> [!NOTE]
> **Medium Risk**
> Changes registry authentication merge semantics for concurrent pushes (safer but may surface new errors to callers); otherwise localized manager/storage behavior with no credential persistence changes.
> 
> **Overview**
> **Credential-aware in-flight dedup:** `CreatePush` still merges duplicate digest+target pushes, but only when both requests agree on whether credentials were borrowed. A mismatch returns **`ErrCredentialConflict`** instead of silently running under the wrong registry auth. In-flight state tracks credential *presence* only (values are never stored).
> 
> **Simpler create path:** Per-request **orphan adoption** (`findPendingPush` and the O(N) scan under the create lock) is removed as redundant with startup recovery and lock ordering. Related orphan/supersede tests are replaced by **concurrent dedup** and **credential conflict** coverage.
> 
> **Operational hardening:** Ready images without a digest are rejected before dedup; **`GetPush`** / **`ListPushes`** respect context cancellation; corrupt push metadata is **logged and skipped** on list rather than disappearing silently; comments clarify terminal persist vs inflight release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4998f67f1fed62103f1cb9f9f6fbe33e566a91b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/imagepush/imagepush.go
+++ b/lib/imagepush/imagepush.go
@@ -27,6 +27,12 @@ var (
 	ErrNotFound      = errors.New("push not found")
 	ErrImageNotReady = errors.New("image not ready for push")
 	ErrInvalidTarget = errors.New("invalid push target")
+	// ErrCredentialConflict is returned when a push request matches an
+	// in-flight job but its credential intent differs from that job's. The
+	// manager never stores credential values, so it can only detect a
+	// presence mismatch; two requests that both borrow different credentials
+	// for the same target are indistinguishable and merge.
+	ErrCredentialConflict = errors.New("push already in flight with different credentials")
 )
 
 // PushRequest describes a request to push a hypeman image to a remote registry.
@@ -77,7 +83,9 @@ type Manager interface {
 	// CreatePush validates the request, persists a queued job, and enqueues it.
 	// A request that matches an in-flight job (same digest and target) returns
 	// the existing job instead of creating a duplicate; the in-flight job's
-	// credentials remain in effect.
+	// credentials remain in effect. A request whose credential presence
+	// differs from the in-flight job's (one borrowed, one not) returns
+	// ErrCredentialConflict instead of silently merging under the wrong auth.
 	CreatePush(ctx context.Context, req PushRequest) (*Push, error)
 	GetPush(ctx context.Context, id string) (*Push, error)
 	// ListPushes returns all pushes, newest first.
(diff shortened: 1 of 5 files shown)


chruffins:
closing this — the fixes are being applied directly onto #350's branch instead

PR #374, opened 2026-08-07
Profile OCI image conversion phases

IlyaasK:
## Why

Hypeship staging spends about 146 seconds per host converting each new agent image digest, but Hypeman currently exposes only one end-to-end image build duration. That cannot distinguish registry transfer, cached-layer unpacking, filesystem creation, or finalization, so the next optimization would be guesswork.

## What

- Record `hypeman_images_build_phase_duration_seconds` with bounded `phase`, `status`, and `cache_status` attributes.
- Separate `oci_cache_lookup`, `registry_pull`, `metadata_extract`, `layer_unpack`, `filesystem_export`, `finalize`, and `cleanup`.
- Record OCI layer count and compressed layer bytes so phase duration can be correlated with image shape.
- Emit structured per-digest phase logs without putting the digest in metric labels.
- Record total build duration consistently on conversion and finalization failures and after cleanup.

## How

The OCI pull result carries timings collected directly around each operation. The image manager publishes them through OpenTelemetry and adds its own filesystem/finalization/cleanup measurements. Cache hits omit the registry-pull phase, making cached unpack cost directly visible.

This does not change pulling, unpacking, conversion, image metadata, or VM launch behavior.

## Before / after

Before: one total build histogram and a pull counter; a 146-second conversion cannot be decomposed.

After: dashboards and logs can answer whether time is spent downloading layers, unpacking all cached layers, or rebuilding EROFS, separately for cache hits and misses.

## Verification

- `DOCKER_CONFIG=/private/tmp/hypeman-empty-docker-config GOCACHE=/private/tmp/hypeman-go-cache go test -tags containers_image_openpgp ./lib/images -count=1`
- `GOCACHE=/private/tmp/hypeman-go-cache go test ./lib/images -run TestImageBuildPhaseMetrics -race -count=1`
- `GOCACHE=/private/tmp/hypeman-go-cache go vet -tags containers_image_openpgp ./lib/images`
- `git diff --check`

The full CI-tagged image package passed against Docker Hub in 17.6 seconds.

---

> [!NOTE]
> **Low Risk**
> Observability-only refactor with a small `finalizeImage` extraction; no changes to pull, unpack, conversion, or VM launch paths.
> 
> **Overview**
> Adds **per-phase observability** for image builds so long conversions can be broken down instead of relying on a single end-to-end duration.
> 
> The OCI client now times **oci_cache_lookup**, **registry_pull** (skipped on cache hit), **metadata_extract**, and **layer_unpack**, and returns layer count plus compressed byte totals. The manager records **filesystem_export**, **finalize**, and **cleanup**, publishes `hypeman_images_build_phase_duration_seconds` (with `phase`, `status`, `cache_status`), and adds OCI layer-count / compressed-bytes histograms. **Structured slog** logs per digest/phase; digest stays out of metric labels.
> 
> **Total build duration** is recorded via a single deferred path on all exit paths (including after cleanup). Final metadata/tag symlink work is extracted into `finalizeImage` without changing pull/unpack/conversion behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94f4780465ccbc381220042466ae21d03f78e795. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/images/manager.go
+++ b/lib/images/manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"sort"
@@ -277,18 +278,23 @@ func (m *manager) createAndQueueImage(ref *ResolvedRef, req CreateImageRequest, 
 
 func (m *manager) buildImage(ctx context.Context, ref *ResolvedRef) {
 	buildStart := time.Now()
+	buildStatus := "failed"
 	buildDir := m.paths.SystemBuild(ref.String())
 	tempDir := filepath.Join(buildDir, "rootfs")
+	defer func() {
+		m.recordBuildMetrics(ctx, buildStart, buildStatus)
+	}()
 
 	if err := os.MkdirAll(buildDir, 0755); err != nil {
 		m.updateStatusByDigest(ref, StatusFailed, fmt.Errorf("create build dir: %w", err))
-		m.recordBuildMetrics(ctx, buildStart, "failed")
 		return
 	}
 
 	defer func() {
 		// Clean up build directory after completion
-		os.RemoveAll(buildDir)
+		start := time.Now()
+		err := os.RemoveAll(buildDir)
+		m.recordImageBuildPhase(ctx, ref.Digest(), "cleanup", time.Since(start), phaseStatus(err), "not_applicable")
 	}()
 
 	m.updateStatusByDigest(ref, StatusPulling, nil)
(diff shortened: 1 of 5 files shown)


sjmiller609 (commented):
Looks good, one issue

lib/images/metrics.go:52-56 — ociCompressedBytes uses OTel’s default histogram boundaries, which end at 10,000 bytes. Normal images therefore land in +Inf, preventing useful distribution or percentile analysis. Add explicit byte-scale boundaries and assert them in the metric test.

IlyaasK:
> Looks good, one issue
> 
> lib/images/metrics.go:52-56 — ociCompressedBytes uses OTel’s default histogram boundaries, which end at 10,000 bytes. Normal images therefore land in +Inf, preventing useful distribution or percentile analysis. Add explicit byte-scale boundaries and assert them in the metric test.

fixed, and added. thank you :)

sjmiller609 (approved):

PR #375, opened 2026-08-09
Add request header authorization to ingress rules

sjmiller609:
## Summary

- add optional `request_header_auth` metadata with a custom `header` and direct `value` through the OpenAPI contract and generated bindings
- validate dedicated header names and 32–256 byte values without including values in validation, reload, or storage errors
- persist ingress metadata atomically at mode `0600` and render protected Caddy routes that strip the verification header before proxying
- order exact host routes before pattern routes on each port so wildcard routes cannot bypass authorization
- document value rotation through delete/recreate, without coupling ingress configuration to Hypeman's environment or restart lifecycle

## Compatibility

- `request_header_auth` remains optional, and persisted ingress JSON without it remains valid
- unprotected routes retain their existing behavior and route shape
- hostname/port conflicts, TLS, HTTP redirects, dynamic DNS upstreams, streaming, WebSocket/SSE behavior, and catch-all routing are unchanged

## Validation

- `go test -race ./lib/ingress`
- `go test -race ./cmd/api/api -run 'TestIngressRequestHeaderAuth'`
- `go test ./cmd/api/api`
- `go vet ./lib/ingress ./cmd/api/api`
- regenerated `lib/oapi/oapi.go` from `openapi.yaml` and verified reproducible output
- repository CI: Linux, macOS, generation, scanning, and install checks pass

---

> [!NOTE]
> **Medium Risk**
> Changes edge routing and stores verification secrets in API responses and on-disk Caddy config; misconfiguration could lock out traffic or leak credentials if clients log responses carelessly.
> 
> **Overview**
> Adds optional **`request_header_auth`** on ingress rules end-to-end: OpenAPI schema, API mapping helpers, domain validation, and Caddy JSON generation.
> 
> Protected rules require an exact **host + custom header** match before the existing dynamic reverse proxy runs. Caddy **strips the verification header** before upstream proxying, and a sibling **403** route handles the same hostname when the header is missing or wrong (HTTP→HTTPS redirects stay unauthenticated). **Exact host routes are ordered before pattern routes** on a port so a protected literal hostname wins over wildcards.
> 
> Validation rejects reserved hop-by-hop/auth headers and enforces a **32–256 byte** secret shape safe for Caddy matchers; validation and Caddy reload errors **omit secret values**. Ingress metadata is written **atomically at mode 0600**; docs describe rotation via delete/recreate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10ab4ea7c0ed59201c6d0bd4a6b11dd851c646eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (approved):
reviewed — the implementation is well-scoped and generally fails closed, but one authorization bypass needs fixing and one design choice needs clarification.

### bug

- `lib/ingress/config.go:280-282`, `lib/ingress/manager.go:356-366` — agreeing with BugBot: route authorization depends on ingress order. An earlier unprotected `{instance}.example.com` route can proxy `admin.example.com` before a later protected exact route runs. Consider rejecting overlapping hostname/port matches or enforcing deterministic exact-over-wildcard precedence. Please add regression coverage for both creation orders and persisted reload order.

### question

- `lib/ingress/config.go:475-486`, `lib/ingress/README.md:91-93` — is coupling each ingress secret to the Hypeman API launch environment intentional? Credentials must be provisioned before startup, rotation requires changing service configuration and restarting Hypeman, and Caddy inherits the environment at `lib/ingress/daemon.go:149`. If this is the intended initial lifecycle, the docs should say the value remains in the process environment and Caddy config—not only the config. I feel Hypeman could use a generic host secret-source abstraction that VMs, ingress, and egress could pull from.

PR #378, opened 2026-08-10
Report vendor VFIO profile availability per free VF

yummybomb:
## summary

- report vendor VFIO profile availability in creatable-instance units by counting free VFs that advertise each profile
- align vendor VFIO availability with the existing API contract and mdev reporting semantics
- update the unit test to cover multiple free VFs sharing a parent GPU

This changes the `gpuProfileSlots` metric for vendor VFIO hosts from parent-GPU units to free-VF units.

## tests

- `GOCACHE=/tmp/gocache go vet ./lib/devices ./lib/resources`
- `GOCACHE=/tmp/gocache go test ./lib/devices -run 'TestVendorVFIO' -count=1 -v`
- `GOCACHE=/tmp/gocache go test ./lib/devices ./lib/resources -count=1`

The vendor VFIO hardware integration test was not run locally.

### DIFF
--- a/cmd/api/api/instances.go
+++ b/cmd/api/api/instances.go
@@ -346,12 +346,16 @@ func (s *ApiService) CreateInstance(ctx context.Context, request oapi.CreateInst
 		var vgpuPending *instances.VGPUCleanupPendingError
 		switch {
 		// Checked first: it wraps the original create error, so a later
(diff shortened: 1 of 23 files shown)


yummybomb:
folding this into the stack where the bug was introduced: the fix now lives in #364 (hypeship/vendor-vfio-backend) and the stack has been rebased.

PR #379, opened 2026-08-10
Confirm hypervisor identity by process start time before kill

yummybomb:
## summary

- persist the Linux `/proc/<pid>/stat` start time alongside the hypervisor PID
- treat a live stored PID with a matching start time as identity-confirmed before destructive cleanup
- keep the existing socket ownership checks unchanged for metadata without a token or with a mismatched token
- capture and clear the PID/start-time pair together throughout create, restore, refresh, stop, standby, fork, and snapshot lifecycles

## why

QEMU can close its QMP listener before the process exits. During that window, socket ownership lookup returns `ErrNoOwningProcess`, so delete or stop refuses to kill the still-live process. Cloud Hypervisor can hit an equivalent `/proc` scan race, and duplicate socket inode resolution can fail the same cleanup path.

The persisted process start time confirms that the stored PID still refers to the original hypervisor without depending on the socket. A recycled PID has a different start time and remains fail-closed. Pre-upgrade metadata without `HypervisorStartTime` deserializes to zero and retains the existing socket-confirmation behavior.

This is intended to be folded into #363. `HypervisorProcessExists` and API orphan cleanup retain their existing fail-open behavior and are out of scope.

## tests

Passed:

- `go build ./...`
- `go vet ./lib/instances/...`
- `go vet ./...`
- `go test ./lib/instances/ -run 'TestResolveLiveHypervisorPID|TestKillHypervisor|TestForceKillHypervisor|TestRefreshHypervisorPID|TestVGPUAssignmentClaimed|TestHypervisorProcessExists|TestProcessStartTime' -v -count=2`
- 215 non-integration `lib/instances` top-level tests selected from unit-test files

Environment-limited:

- `go test ./lib/instances/ -short` does not exclude this package's integration tests; the run failed on missing `mkfs.erofs`, image preparation, and unprivileged network/hypervisor requirements. The non-integration subset above passed after separating those tests.
- `sudo go test ./lib/instances/ -run 'TestQEMUStandbyAndRestore' -v -timeout 30m` reached QEMU startup but failed because `/dev/vhost-vsock` is unavailable. The focused missing-socket kill-path test passes deterministically; CI can exercise the full lifecycle.
- `TestCreateInstanceWithNetwork` was not rerun because the same host lacks `/dev/vhost-vsock`.

### DIFF
--- a/lib/instances/admission_allocations.go
+++ b/lib/instances/admission_allocations.go
@@ -108,6 +108,7 @@ func (m *manager) rollbackAdmissionAllocationActive(stored *StoredMetadata) {
 	// allocation marked active. Clear the in-memory PID first so any later sync
 	// from this metadata view also treats the instance as inactive.
 	stored.HypervisorPID = nil
+	stored.HypervisorStartTime = 0
 	m.setAdmissionAllocationActive(stored, false)
 }
 
--- a/lib/instances/create.go
+++ b/lib/instances/create.go
@@ -849,6 +849,7 @@ func (m *manager) startAndBootVM(
 
 	// Store the PID for later cleanup
 	stored.HypervisorPID = &pid
+	stored.HypervisorStartTime = processStartTime(pid)
 	log.DebugContext(ctx, "VM started", "instance_id", stored.Id, "pid", pid)
 
 	// Optional: Expand memory to max if hotplug configured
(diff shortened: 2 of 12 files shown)


yummybomb:
closing — fix to be folded into #363 by the stack owner

PR #387, opened 2026-08-10
Prevent stale Created state after restore

sjmiller609:
## Summary

- publish the running hypervisor state after a successful restore resume
- prevent concurrent restore callers from observing a stale cached `Created` state and receiving a 409
- add a regression test covering two concurrent restores with a transitional state query

## Tests

- `go test -race ./lib/instances -run '^(TestConcurrentRestoreClearsTransitionalHypervisorStateCache|TestLifecycleNoopTransitionsReturnCurrentInstanceWithoutEvent)$' -count=1`
- `go vet ./lib/instances`
- `go test ./lib/instances -count=1` *(integration tests requiring `mkfs.erofs` and completed local image builds could not run in this environment)*

---

> [!NOTE]
> **Low Risk**
> Small, targeted ordering fix on the restore success path plus a unit test; no auth or data-model changes.
> 
> **Overview**
> After a successful standby restore, **`restoreInstance` now writes `hypervisor.StateRunning` into the hypervisor state cache** immediately after metadata is saved and before building the API response. That clears a transitional **`Created`** value that can be left when a state query hits the VMM during `RestoreVM` (before resume finishes), which could otherwise make a overlapping restore see the wrong cached state and fail with a conflict.
> 
> A new regression test **`TestConcurrentRestoreClearsTransitionalHypervisorStateCache`** runs two concurrent `RestoreInstance` calls while the fake starter pins **`Created`** in the cache mid-restore; both completes succeed and only one restore runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc9f608f8f6f4e6b30d9d95554777cbc7cfdf036. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/instances/restore.go
+++ b/lib/instances/restore.go
@@ -389,6 +389,10 @@ func (m *manager) restoreInstance(
 		log.WarnContext(ctx, "failed to update metadata after restore", "instance_id", id, "error", err)
 	}
 
+	// A state query during RestoreVM may have cached Created. Resume succeeded,
+	// so publish Running before deriving the response and releasing the lock.
+	m.storeCachedHypervisorState(id, hypervisor.StateRunning)
+
 	// Return instance state from current metadata without forcing a log scan.
 	finalInst := m.toInstanceWithoutHydration(ctx, meta)
 	// Record metrics
--- a/lib/instances/restore_state_cache_test.go
+++ b/lib/instances/restore_state_cache_test.go
@@ -1,0 +1,96 @@
+package instances
+
+import (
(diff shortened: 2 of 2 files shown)


yummybomb (approved):
Looks good! 

publishing Running before deriving the response correctly prevents the stale cache value from propagating into the lifecycle event and subsequent restore call. the focused concurrency test covers the regression well.

PR #391, opened 2026-08-11
images: mirror digest-pinned indexes byte for byte

mpuig:
## Problem

A Dockerfile that pins its base image by digest almost always pins the **multi-arch index digest** — it's the digest `docker pull` and Docker Hub print. `MirrorBaseImage` resolves that reference to the platform image and pushes it under the index digest, which the local registry correctly refuses:

```
push to local registry: PUT …/v2/library/python/manifests/sha256:9b4929a7…:
unexpected status code 400 Bad Request: digest mismatch:
expected sha256:9b4929a7…, got sha256:1e58d36e…
```

Because the mirror failure is only a WARN, the build proceeds and fails later with an unrelated error (`create builder instance: image is required` on v0.3.0), so the root cause is invisible in the API response. Net effect: **every digest-pinned `FROM` is unbuildable** — and supply-chain-pinned Dockerfiles are exactly the ones that hit it.

Reproduced on v0.3.0 on a GitHub-hosted ubuntu runner with `FROM python:3.13-alpine@sha256:9b4929a7…`; journal evidence above.

## Fix

An index named by digest is mirrored **as the index itself** (`remote.WriteIndex` — children first, then the index manifest), so the destination stores content whose digest is exactly the one the caller pinned.

Unchanged behavior:
- tag references still mirror only the platform-resolved image (the storage-saving path);
- a digest reference naming a plain (single-platform) manifest already round-trips and is untouched.

The push logic moves into `pushMirrored` so the property is testable against in-memory registries (`pkg/registry`, as in `imagepush/manager_test.go`): the new tests pin a random two-manifest index by digest, mirror it, and prove the pinned digest resolves at the destination — which is the assertion the old code could not pass.

## Testing

- `go test ./lib/images/` green (new + existing tests)
- `go build ./...` green

Found while building [barista](https://github.com/mpuig/barista.sh) on hypeman; happy to adjust if you'd rather handle index-pinned refs differently (e.g. refusing them loudly instead of mirroring all platforms).

---

> [!NOTE]
> **Medium Risk**
> Changes the base-image mirror path used during builds; behavior differs for digest-pinned multi-arch refs (full index push vs single platform), though tag mirroring stays the same and new tests cover the push seam.
> 
> **Overview**
> Fixes **digest-pinned `FROM` lines** that pin the multi-arch **index** digest (typical for supply-chain pinned Dockerfiles). Mirroring used to resolve the reference to a single-platform image and push it under the index digest, which registries reject with **digest mismatch** and broke those builds.
> 
> **`MirrorBaseImage`** now fetches a **`remote.Descriptor`** first and delegates push to new **`pushMirrored`**: when the source is a **digest reference to an index**, it mirrors the **whole index** via **`remote.WriteIndex`** so the local registry stores content under the **exact pinned digest**. **Tag references** are unchanged—they still mirror only the **platform-resolved** image to save storage, including picking the requested platform from a multi-arch tag.
> 
> Adds **httptest/registry** tests for index digest preservation, tag-only image mirroring, and multi-platform tag resolution to **`linux/arm64`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13c870c8279abe32f14d70707b7c60210330f6fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/images/mirror_test.go
```diff
@@ -89,3 +94,82 @@ func TestStripScheme(t *testing.T) {
...
+	// the destination under the pinned digest, i.e. it verified on push.
+	mirrored, err := remote.Get(dstRef)
+	require.NoError(t, err, "the pinned digest must resolve at the local registry")
+	assert.Equal(t, indexDigest, mirrored.Digest)
+	assert.True(t, mirrored.MediaType.IsIndex(), "an index must arrive as an index")
+}
+
+// The storage-saving path is unchanged: a tag reference mirrors only the
+// (platform-resolved) image, not the whole index.
+func TestPushMirroredByTagStillMirrorsImage(t *testing.T) {
```
chruffins, line 151: the tag test uses a single-platform image, so it doesn’t check that tagged multi-platform indexes resolve to just the requested platform image

chruffins (changes requested):
thanks for putting up this PR! it's just missing some test coverage but looks good otherwise

mpuig:
Thanks! Added test coverage in `eb23dca8`: `TestPushMirroredByTagResolvesMultiPlatformIndexToRequestedPlatform` builds a two-platform index (linux/amd64 + linux/arm64), fetches the tag with `WithPlatform(linux/arm64)`, and asserts the mirror arrives as that arm64 image — not the whole index, and not the amd64 sibling. The prior single-platform test couldn't distinguish those cases.

mpuig:
@chruffins the requested test coverage is in `eb23dca8` and the review thread is resolved — CI is green. Could you take another look and re-approve when you get a chance? I can't re-request the review myself from the fork. Thanks!

PR #353, opened 2026-08-05
Add pushes API for exporting images to remote registries

chruffins:
Layer 3 of remote registry push support (stacked on #348, #350).

## What

The HTTP surface for outbound pushes, docker-aligned credentials included.

### API (openapi.yaml → generated)
- `POST /pushes` → 202 + Push job. Body: `image` (hypeman image, must be `ready`), `target` (full remote ref), optional `insecure`, optional `credentials`
- `GET /pushes` (newest first), `GET /pushes/{id}`
- Push object: `queued → pushing → pushed/failed` with queue position, error, layers/bytes, timestamps
- errors: 400 invalid_name/invalid_target, 404 image not found, 409 image_not_ready, 500 otherwise

### Credentials — borrow, don't store
- request `credentials` (username/password/registry_token, mirroring docker config.json fields) map to an `authn.AuthConfig` and are borrowed for that single push only — never persisted (asserted in layer 2) or logged
- omitted → falls back to the server's default resolution (Docker keychain)
- scopes reuse `image:write` (POST) / `image:read` (GET) — no new scope plumbing, existing tokens work

### Wiring
- `ProvidePushManager` (resolver = image manager, concurrency from `limits.max_concurrent_pushes`, default 2)
- `ApiService.PushManager` + wire regen
- **GC roots composition**: `compositeOCICacheRoots` in main.go feeds the OCI cache GC both the registry's BuildKit cache tags and the push manager's in-flight digests (`imagepush.Manager` gains `LiveCacheManifestDigests` to satisfy `ocicachegc.RootsProvider`)

### Tests
Handler-level with a fake push manager (hermetic, follows the images_test pattern): request + credentials mapping, nil-credentials fallback, full error-status mapping table, get-not-found, list empty/all with layers/bytes mapping.

## Notes
- `lib/oapi/oapi.go` regenerated via `make oapi-generate` (pinned oapi-codegen v2.5.1); the embedded-spec blob diff includes the pre-existing compression drift any regen under the current Go toolchain produces on main too
- `wire_gen.go` regenerated with wire v0.6.0
- SSE/events for pushes not included — clients poll `GET /pushes/{id}` (builds-style events can be a follow-up)
- the 409 for not-ready images is deliberate: the image exists but can't be pushed yet

## Verification
- `go build -tags containers_image_openpgp ./...` — full tree builds
- new handler tests + `lib/scopes`, `lib/providers`, `cmd/api/config`, and all push/image packages pass

---

> [!NOTE]
> **Medium Risk**
[...]

PR #398, opened 2026-08-11
Add polling headers to push creation

chruffins:
## summary
- document `Location` and `Retry-After` on `POST /pushes` responses
- populate both headers with the created job URL and a two-second polling interval
- restore the in-process `WaitForPush` manager method for future callers; the HTTP API does not consume it and continues polling
- add coverage for the response headers and `WaitForPush`

## testing
- `go test ./lib/oapi`
- `go test ./lib/imagepush -count=1`
- targeted push API tests pass with temporary empty embedded-binary placeholders
- the full `cmd/api/api` package still has unrelated environment-dependent failures because `mkfs.erofs` and network bridge setup are unavailable


chruffins:
Moved these commits onto PR #353 so the polling headers and in-process WaitForPush restoration are part of the original change.

PR #399, opened 2026-08-11
Restore standby after failed Firecracker restore

sjmiller609:
## Summary

- reap Firecracker when a snapshot restore fails
- remove the stale API socket so the retained snapshot derives as Standby and can be retried
- cover failed-restore process and socket cleanup

## Testing

- `go test ./lib/hypervisor/firecracker`
- `go test -race ./lib/hypervisor/firecracker`
- `go vet ./...`
- `make build`

The broader `go test -race ./lib/instances` run was blocked by unauthenticated Docker Hub rate limits and existing asynchronous image-build race reports.

---

> [!NOTE]
> **Medium Risk**
> Touches VM restore failure paths and process/socket lifecycle; behavior change is localized to failed restore cleanup with a direct test.
> 
> **Overview**
> When **Firecracker snapshot restore** fails partway through `RestoreVM`, the deferred cleanup now runs **`cleanupFailedRestore`** instead of only sending **SIGKILL** to the VMM.
> 
> That helper **reaps the process** with `Wait4`, **deletes the API socket** at `socketPath`, and is covered by a test that asserts the socket is gone and the process is no longer running. The intent is to avoid leaving a live VMM and a stale `fc.sock` so the retained snapshot can be treated as **Standby** and restore can be retried.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a0e7b78207144a4f2b4b157b62723a4ff6572b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/hypervisor/firecracker/process.go
+++ b/lib/hypervisor/firecracker/process.go
@@ -131,7 +131,7 @@ func (s *Starter) RestoreVM(ctx context.Context, p *paths.Paths, version string,
 	}
 
 	cu := cleanup.Make(func() {
-		_ = syscall.Kill(pid, syscall.SIGKILL)
+		cleanupFailedRestore(pid, socketPath)
 	})
 	defer cu.Clean()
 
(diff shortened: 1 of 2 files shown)


chruffins (approved):
lgtm

PR #403, opened 2026-08-12
ci: allow manual test workflow runs

chruffins:
## Summary

- add a `workflow_dispatch` trigger to the test workflow
- keep the existing push trigger unchanged

## Testing

- `git diff --check`
- verified the workflow trigger declaration locally

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ name: Test
 
 on:
   push: {}
+  workflow_dispatch:
 
 jobs:
   test:


sjmiller609 (approved):

sjmiller609 (approved):
if this is for reviewing public PRs, sometimes it can work like you have a slash command you run as a maintainer to trigger tests for contributors to run, but I like this too

chruffins:
> if this is for reviewing public PRs, sometimes it can work like you have a slash command you run as a maintainer to trigger tests for contributors to run, but I like this too

I didn't know that was an option; just noticed that test CI never ran (so could never be restarted) for a public PR

PR #405, opened 2026-08-12
Run tests for public pull requests

chruffins:
## summary
- trigger the Test workflow for pull requests targeting main
- add maintainer-only `/test` and `/run-tests` commands
- dispatch tests against the exact current fork commit
- skip Docker Hub and TLS credentials for fork PRs

## usage
A maintainer, collaborator, or organization member comments `/test` on an open PR targeting `main`. The workflow resolves the PR head SHA, dispatches the existing Test workflow from `main`, and comments back with the workflow URL.

## safety
The command workflow does not execute PR code. It only reads PR metadata on a hosted runner and dispatches the existing privileged test workflow after an authorized maintainer command. Fork runs receive no repository secrets, use an empty Docker config, and check out the exact fork commit with credentials disabled.

## testing
- `git diff --check`
- actionlint passed; the only ignored diagnostic is the repository-specific `kvm` self-hosted runner label
- workflow-only change; GitHub Actions is running validation on this PR.

### DIFF
--- a/.github/test-command-maintainers
+++ b/.github/test-command-maintainers
@@ -1,0 +1,7 @@
+# SECURITY: REVIEW PUBLIC CONTRIBUTIONS BEFORE APPROVING THEM
+#
+# Everyone listed below is expected to verify that a public contribution is safe
+# to run on Kernel CI before approving it with the /test slash command.
+# One GitHub username per line.
+sjmiller609
+chruffins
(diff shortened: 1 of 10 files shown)


sjmiller609 (approved):
We decided that it's OK to allow running public contributor's changes in our testing infra, but we have to be careful in reviewing their change doesn't introduce malicious code before running the slash test command on a PR

PR #410, opened 2026-08-12
Borrow registry credentials for image pulls

chruffins:
## summary

Allows Hypeman to pull private registry images using temporary, per-request credentials without persisting them.

- accept optional Docker-style credentials on `POST /images`
- use borrowed credentials for manifest resolution and asynchronous image pulling without persisting credential material
- fail interrupted credentialed pulls after restart so callers can retry with fresh credentials
- preserve the server keychain path when credentials are omitted or empty

## tests

- `go test ./lib/images -run 'Test(BorrowedCredentialsAuthenticateResolveAndPull|CreateImageRequestCredentialsAreNotPersisted|RecoverInterruptedCredentialedPullFailsForFreshRetry|RecoverInterruptedBuildsCapturedFixture)' -count=1`\n- `go test ./cmd/api/api -run 'TestCreateImage_(MapsBorrowedCredentials|EmptyCredentialsUseServerKeychain|ErrorStatusMapping)|TestCreatePush_(NoCredentialsStaysNil|EmptyCredentialsFallsBackToDefault)' -count=1`\n- `make oapi-generate` and verified the generated file is clean\n\n## notes\n\n- a broader `go test ./cmd/api/api -count=1` was attempted but unrelated integration tests could not download a kernel release because GitHub returned 503


chruffins:
fixed both review findings in `2ed977b`:

- in-flight pulls now keep a SHA-256 credential fingerprint and reject anonymous/credentialed or different-credential dedup attempts with `409 credential_conflict`, instead of silently inheriting the first request's login
- borrowed credentials now have a 30-minute lease starting when the pull is queued; the queue closure does not capture the secret, expired queued pulls fail explicitly, active registry work receives the same deadline, and the completion hook clears the lease
- added regression coverage for credential conflicts, queued credential expiry, and the API's 409 mapping

validated with:
- `go vet ./lib/images ./cmd/api/api ./lib/imagepush ./lib/registrypush ./lib/queue`
- focused credential tests under `go test -race`
- API status-mapping tests
- `go test ./lib/imagepush ./lib/registrypush ./lib/queue`

I also attempted the broader image/API suites; their Docker Hub integration cases are currently blocked by the unauthenticated pull-rate limit, while the focused local-registry and unit coverage passes.

sjmiller609 (approved):
these two findings seem worthwhile to me, but if the second one gets too complicated, maybe skip it.

**lib/images/manager.go:181-200** — in-flight dedup ignores credential differences, so fresh credentials may silently inherit the original request’s authentication. fingerprint and reject mismatches, as push does.
**lib/images/manager.go:495-503** — restart recovery also fails builds in converting, although registry access has finished. these should recover from the local OCI cache without credentials.

chruffins:
addressed the remaining approval note in `2846289`: an interrupted credentialed image in `converting` now resumes without credentials when its OCI image is already cached locally. `pending`/`pulling` jobs—and `converting` jobs missing the cache—still fail and require fresh credentials, so recovery never falls back to an unauthenticated remote pull.

added a cached-conversion restart regression test; focused recovery tests pass under `go test -race`, and `go vet ./lib/images` passes.

PR #411, opened 2026-08-13
Fix public test workflow coordination

sjmiller609:
## Summary

- fetch `main` into `FETCH_HEAD` so persistent self-hosted workspaces cannot reject the UFFD version check
- isolate macOS concurrency groups by the tested public commit so slash-command runs do not cancel normal branch CI
- allow the slash-command controller to comment on pull requests without marking a successful dispatch failed if GitHub rejects the confirmation comment

## Why

The first run after #405 merged failed because a persistent runner already had a divergent `upstream/main` ref from a prior fork test. The end-to-end slash-command run also dispatched successfully but was reported as failed when its confirmation comment received a 403.

## Testing

- parsed both workflow files as YAML
- ran actionlint (only the repository-specific `kvm` runner label warning remains)
- ran the UFFD version check against `FETCH_HEAD`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Workflow-only changes with no application runtime impact; minor risk of overlapping macOS jobs if concurrency grouping behaves unexpectedly.
> 
> **Overview**
> **Hardens public `/test` dispatch and self-hosted CI** so fork/slash-command runs do not fight normal branch jobs or fail for benign GitHub API issues.
> 
> The Linux job’s UFFD version check now shallow-fetches upstream `main` into `FETCH_HEAD` and compares against that, avoiding stale or divergent `upstream/main` refs on persistent runners. macOS `test-darwin` and `e2e-install` concurrency groups key off `inputs.pr_head_sha` when present (else `github.ref`), so slash-command test runs no longer share a group with push CI and cancel each other.
> 
> The test-command **Report dispatch** step no longer fails the job when posting the confirmation PR comment fails (e.g. 403); it logs the API response, emits a workflow warning, and records the outcome in the job summary while leaving the successful dispatch intact.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbf376a0bade2ba66931eb52df33b5b7d53408ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #412, opened 2026-08-13
Fix persistent CI main ref fetch

chruffins:
## summary
- force-update the persistent upstream/main ref before checking the UFFD pager version
- prevent self-hosted runners from failing on stale remote-tracking refs

## testing
- git diff --check


chruffins:
superseded by PR #411

PR #417, opened 2026-08-17
Remove the command-line fallback from socket owner resolution

yummybomb:
## Summary

Stacked on #363. Deletes `pidByCmdline` and the `confirmed` flag from socket owner resolution, collapsing `classifyResolvedHypervisorOwner` from an 8-branch matrix to 4 cases.

A command-line-only match could never authorize teardown — destructive paths already refused unconfirmed PIDs — so the fallback's only effect was distinguishing provable death from ambiguity. That distinction only matters if the fd scan can miss a live owner, which requires the caller to lack `CAP_SYS_PTRACE`. Hypeman runs as root (dev) or with a full capability set (prod), so the listener scan is authoritative: a confirmed owner is returned, a missing listener is proof the hypervisor is gone, and only a failed scan fails closed.

The fallback was also actively harmful: a debug client carrying the socket path in its argv (`ch-remote`, `socat`) resolved as a live unconfirmed match and wedged stop/delete with a 500 until the client exited. That case now correctly classifies as provable death and the bystander is never signaled.

**Assumption made explicit:** resolution correctness now depends on hypeman retaining ptrace-equivalent capabilities (root or `CAP_SYS_PTRACE`). This is documented on `ResolveProcessPID`. If the service is ever de-privileged below that, the fd scan can miss owners and this needs revisiting.

## Testing

- `go vet` clean; `go test -race ./lib/hypervisor/ ./lib/instances/` targeted identity/kill/shutdown/backfill suites pass locally (`TestSocketCacheKeyChangesWhenSocketIsRecreated` fails in this sandbox on the unmodified base too — inode reuse in /tmp)
- Same suites cross-compiled and run as root on a dev host with real `/proc`: all pass
- Live validation on that host: a real daemon control socket resolves to its owning PID; a dead socket path held in a live process's argv returns `ErrNoOwningProcess` instead of failing closed
- New/updated coverage: `TestResolveProcessPIDIgnoresCommandLineBystander`, `TestKillHypervisorIgnoresCommandLineBystander`, `TestShutdownHypervisorIgnoresCommandLineBystander`, reworked `TestClassifyResolvedHypervisorOwner` table

### DIFF
--- a/lib/hypervisor/socket_pid_linux.go
+++ b/lib/hypervisor/socket_pid_linux.go
@@ -20,43 +20,31 @@ var procDir = "/proc"
 const soAcceptcon = 0x10000
 
 // ResolveProcessPID finds the process currently holding the listening Unix
-// socket for the given hypervisor control path. confirmed reports whether the
-// PID was found through socket ownership rather than its command line.
-func ResolveProcessPID(socketPath string) (pid int, confirmed bool, err error) {
+// socket for the given hypervisor control path, via the socket inode in
+// /proc/net/unix and each process's fd table. The fd scan requires the
+// caller to hold CAP_SYS_PTRACE (or run as root) so no live owner is missed;
+// an ErrNoOwningProcess result is proof the listener is gone.
+func ResolveProcessPID(socketPath string) (pid int, err error) {
 	return resolveProcessPID(socketPath, 0)
 }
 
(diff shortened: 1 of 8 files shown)


yummybomb:
folding this into #363 directly — same review scope

PR #419, opened 2026-08-17
api: report the actual fork mode on Instance

mpuig:
## What

Fork already decides copy-on-write vs full copy — `fork.go` computes
`shareMemFile := HypervisorType == Firecracker && source.State == Standby` — but
that decision was a local variable. Callers had no way to know which mode a fork
actually used and had to infer it from the hypervisor. This surfaces it.

- `StoredMetadata.ForkMode` records the measured mode at fork time
  (`ForkModeShared` / `ForkModeCopied`), mirroring the existing `shareMemFile`
  decision exactly. Persisted, so `GetInstance` reports it too. Empty for
  instances not created by a fork.
- `Instance.fork_mode` (read-only enum: `shared` | `copied`) added to the
  OpenAPI schema, regenerated (`make oapi-generate`), and mapped in the instance
  API response.

`shared` = copy-on-write via a shared mem-file (the source is frozen only if it
was Running at fork time); `copied` = a full private copy (stopped-source forks,
or hypervisors without shared-memory fork).

## Why

A consumer that forks needs to record and report the *actual* mode a fork used
(and whether a large source was frozen), rather than guessing from the backend.
Exposing the mode the server already computed makes that honest.

## Compatibility

Additive and backward compatible: existing metadata reads back with an empty
`fork_mode`, and the field is `omitempty` on the wire.

## Test

`TestForkInstance_VZStoppedSourceSupported` asserts a VZ stopped-source fork
reports `ForkModeCopied`.

---

> [!NOTE]
> **Medium Risk**
> Changes fork and snapshot-fork metadata and API surface; behavior is additive but touches core instance lifecycle paths where incorrect metadata could mislead consumers.
> 
> **Overview**
> Forked instances now expose **how memory was duplicated** instead of forcing callers to infer it from hypervisor and source state.
> 
> **`fork_mode` on Instance** (`shared` | `copied`) is persisted in metadata when an instance is created via instance fork or snapshot fork, matching the existing `shareMemFile` decision (Firecracker + standby → shared; otherwise copied). The read-only field is on the OpenAPI `Instance` schema and returned from `GetInstance` / fork responses via `instanceToOAPI`.
> 
> **Fork snapshot** accepts optional **`tags`** on `ForkSnapshotRequest`; request keys override tags cloned from the snapshot source while other source tags remain. The snapshots API maps tags into the domain request, and `validateForkSnapshotRequest` runs the same tag validation as create paths.
> 
> Tests cover VZ stopped-source forks reporting `ForkModeCopied` and invalid fork-snapshot tags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1dab359c602433e830f874bf9a06f2fa30ae0bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mpuig:
Rebased onto latest `main` (resolved the conflict with #434's instance-expiration controls: `openapi.yaml` merged cleanly with both `fork_mode` and `expires_at`, and `lib/oapi/oapi.go` was regenerated via `make oapi-generate` rather than hand-merging the embedded spec; `cmd/api/api/instances.go` keeps both the ForkMode and ExpiresAt mappers).

Also addressed the Bugbot review:
- **Fork snapshot skips tag validation** (`9f4e208d`): `validateForkSnapshotRequest` now calls `tags.Validate(req.Tags)`, mirroring create-instance and create-snapshot, so caller-supplied tags can't reach instance metadata unvalidated on this direct manager path. Added `TestValidateForkSnapshotRequestValidatesTags`.
- **Snapshot fork omits ForkMode**: already handled — `forkSnapshot` computes `shareMemFile` and writes `forkMeta.ForkMode` (shared/copied), same as the instance-fork path.

`go vet` clean; fork/snapshot/tags tests green.

PR #428, opened 2026-08-19
Add Windows hypervisor primitives

sjmiller609:
## summary

- add direct-kernel and UEFI boot modes with raw and qcow2 disks
- support Secure Boot OVMF and per-instance swtpm lifecycle on standard QEMU
- advertise and validate UEFI/TPM requirements through runtime capabilities
- manage QEMU and swtpm through one cleanup stack while preserving required socket ordering
- expose software TPM output through the instance logs API
- document firmware, TPM ownership, restart, and cleanup semantics in the hypervisor READMEs

## targeted tests

- `go test ./lib/hypervisor/qemu ./lib/hypervisor ./lib/paths ./lib/instances ./cmd/api/api -run ... -count=1`

Long-running KVM and full CI suites were not run for this review pass.

---

> [!NOTE]
> **Medium Risk**
> Touches VM boot orchestration (swtpm/QEMU ordering, process reconciliation) and shared config validation used by all hypervisor starters; mistakes could break Linux direct-kernel paths or leave orphaned TPM processes.
> 
> **Overview**
> Adds **Windows-oriented hypervisor primitives** on standard amd64 QEMU: **UEFI firmware boot** (OVMF code/vars, optional Secure Boot with `smm=on`), **qcow2 disks**, and a **per-instance software TPM 2.0** via `swtpm` started before QEMU on a shared cleanup stack.
> 
> **`VMConfig`** gains `BootMode`, `Firmware`, `TPM`, and disk `Format`, with shared **`ValidateBootConfig`** / **`ValidateDirectRawConfig`** so Cloud Hypervisor, Firecracker, and vz reject UEFI, TPM, and non-raw disks. QEMU profiles advertise **`SupportsUEFIBoot`** / **`SupportsTPM`** on q35 only; microvm stays direct-kernel/raw.
> 
> **Observability & API:** instance logs accept **`source=swtpm`** (OpenAPI + API handler); README documents `hypeman logs --source swtpm`.
> 
> **CI** installs `ovmf`/`swtpm`, exports OVMF paths, and runs **`TestWindowsConfigIntegration`** (KVM) with retries. QMP socket wait extended to 30s; fork/snapshot path rewriting covers firmware and TPM paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4b75cbc4c42f9e4843cb1e2c22cc4c3b9681558. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #435, opened 2026-08-20
Detect wedged vGPU VFs from the guest and report to the health store

yummybomb:
## Summary

Top half of the wedged-VF work, stacked on #462 (the VF health store, placement exclusion, admission, and `/resources` fields). This PR adds the detection path that feeds that store.

- The guest agent watches `/dev/kmsg` for kernel-facility `NVRM: ... RmInitAdapter failed!` records, probes driver init at boot with `nvidia-smi -L` when the image has it, and exposes the current GPU initialization state over `GetGPUInitStatus` via vsock.
- A host sentinel controller polls vendor-VFIO instances with up to 64 concurrent checks and reports failures and successes into the VF health store from #462.
- Only instances with a live VMM (control socket present) are polled; a stopped or standby instance whose failed release left its claim in metadata is skipped, since QEMU vsock dials by guest CID alone and a stale CID could be reused by an unrelated instance.
- Reports are keyed on `GPUClaimedAt` from #462, the identity a claim receives when it is persisted.

## Rebuilt on the claim-first allocator

This branch was rebuilt from the new #462 head after #321 merged as the claim-first rewrite. The guest agent, guest RPC, sentinel controller, wiring, metrics, and docs are the previously reviewed code, squashed into one commit. What changed in the port:

- `GPUAssignedAt` is gone from `main`; the sentinel keys reports on `GPUClaimedAt`, which #462 sets in the same metadata save as the claim.
- The `GPURetainedForCleanup` skip is gone with the retention-record design. The existing control-socket check already excludes instances without a VMM.
- The "allocate the vGPU immediately before metadata persistence" change to `create.go` is dropped. Claim-first already persists the claim before the VF is touched.
- The reconcile liveness change is re-hooked into `main`'s two `hypervisorMayBeAlive` call sites; reconcile and the create/start cleanup guard share one wrapper that logs the resolution error and increments the liveness counter when it fails closed.
- The retention-record tests that the old branch adjusted no longer exist and were dropped with them.

Rebased onto the current #462 head (`cc6b625`, itself rebased onto `main` after #428): the store now takes the threshold in `InitVFHealth`, retries a failed persist on read, and passes the quarantine set into the selector struct. The four review-round commits were squashed into the detection commit for the rebase; the review follow-ups (sentinel keyed on `devices.FormatVFAssignedAt`, throttled repair warning, lock comment on `RepairVFHealthStore`) are a separate commit on top.

## Safety and failure handling

- GPU initialization success is terminal in the guest agent, so a delayed failure watcher cannot replace a confirmed success.
- The host reads GPU init state only from the guest-agent RPC; workload output on the shared serial console cannot influence the tally, and userspace writes to `/dev/kmsg` are rejected by the kernel-facility check. The guest is still the reporter, so a root workload that replaces the agent can report a failure; the per-assignment threshold and randomized VF selection bound how fast that drains capacity.
- A slow `nvidia-smi` attempt is killed after 30 seconds. The guest agent waits for that process to be reaped before retrying, so an attempt stuck in uninterruptible I/O cannot accumulate concurrent probes; the independent kmsg watcher still reports the underlying init failure.
- Unreachable guests do not serialize a host-wide scan; checks run with a fixed concurrency limit.
- Repeated polls and controller restarts do not double-count an assignment. The sentinel revalidates the VF assignment before reporting, and the health store deduplicates reports by assignment.
- Reconciliation preserves claims when hypervisor liveness is uncertain and records that condition in logs and metrics.

## Observability

- `hypeman_instances_vgpu_sentinel_init_failures_total`
- `hypeman_instances_vgpu_sentinel_quarantines_total`
- `hypeman_instances_vgpu_sentinel_checks_total` by `result` (`ok`, `failed`, `unknown`, `rpc_error`, `unsupported_agent`, or `list_error`)
- `hypeman_instances_vgpu_quarantined_vfs`
- `hypeman_instances_vgpu_vf_health_store_unavailable`
- `hypeman_instances_vgpu_liveness_uncertain_total`

`lib/devices/GPU.md` gains the detection and sentinel documentation.

## Known gap

The `nvidia-smi -L` probe is the only source of an OK state. An image without `nvidia-smi`, or a driver that takes longer than the 10 minute probe window to initialize, stays UNKNOWN: its failures are still detected and tallied, but its assignments can never clear a tally or rescind a quarantine. Fleets running such images recover VFs only through the manual runbook in `lib/devices/GPU.md`.

## Out of scope

An operator force-cycle endpoint. Recovery remains the documented manual DCGM quiesce, SR-IOV cycle, state edit, restart, and verification flow.

## Testing

Passed locally:

```bash
go vet ./lib/instances ./lib/devices ./lib/guest ./lib/system/guest_agent ./lib/providers ./cmd/api
go test -race ./lib/devices ./lib/guest ./lib/system/guest_agent ./lib/resources ./lib/providers
go test -race ./lib/instances -run 'VGPU|Sentinel|Reconcile|Metric|HypervisorMayBeAlive|ResolveLiveHypervisor|Select|Quarantin|StoredVGPU|Cleanup|ReleaseStored|Snapshot.*GPU|Fork.*GPU'
```

`cmd/api/wire_gen.go` was regenerated with `wire` and matches. `lib/guest/guest.pb.go` and `guest_grpc.pb.go` are carried over from the previous head; `guest.proto` has not changed on `main` since, and the header records the same `protoc` and `protoc-gen-go` versions as the committed files.

The full `lib/instances` suite was not run here; tests that boot real VM images need a host this environment does not provide.

The underlying wedge signal and manual recovery sequence were previously validated on L40S hardware. A live end-to-end run of the guest watcher and host controller is still required before merge, including two failed assignments and the success path.

---

> [!NOTE]
> **High Risk**
> Changes vGPU capacity management (quarantine), runs continuous vsock polling against live VMs, and tightens when stale VF claims are released—mistakes could mis-quarantine hosts or leave wedged VFs in rotation.
> 
> **Overview**
> Adds an end-to-end path to **detect guest NVIDIA `RmInitAdapter` failures** on vendor VFIO vGPUs and feed the existing VF health store (quarantine / placement exclusion).
> 
> The **guest agent** watches kernel `/dev/kmsg` for NVRM init-failure lines, optionally probes with `nvidia-smi -L`, and exposes **`GetGPUInitStatus`** over vsock. Shared `scanKmsg` replaces ad-hoc kmsg reading in the clock keeper.
> 
> The **vGPU sentinel controller** runs in the API process (wire-injected), polls only **running** vendor-VFIO instances (control socket present), and records failures/successes keyed on **`GPUClaimedAt`**. It repairs an unavailable health store once per poll and emits sentinel-specific metrics; instance metrics add quarantine and store-unavailable gauges plus **`hypeman_instances_vgpu_liveness_uncertain_total`** when reconcile/cleanup cannot prove the hypervisor is dead.
> 
> **VF health store** changes: `RepairVFHealthStore`, stricter persist behavior (no-op reports succeed while persist is failed; mutations fail closed), and a fast path so routine OK reports skip `vendorVFIOMu` when the VF has no tallies.
> 
> **Reconcile** preserves claims on ambiguous hypervisor liveness via `vgpuHypervisorMayBeAlive` (replacing the removed package-level helper). **`GPU.md`** documents detection, limits (OK only from `nvidia-smi`), and observability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b56faa60c4438b552718ef2b07386d1e66f8a44b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/devices/vf_health.go
+++ b/lib/devices/vf_health.go
@@ -312,9 +312,65 @@ func ReportVFInitFailure(report VFInitFailureReport) (VFReportResult, error) {
 // ReportVFInitSuccess clears failures through an exactly matched successful
 // assignment. A quarantine is rescinded only when that assignment triggered it.
 func ReportVFInitSuccess(report VFInitSuccessReport) (VFSuccessResult, error) {
+	if !vfHealth.hasFailures(report.VFAddress) {
+		return VFSuccessResult{}, nil
+	}
 	vendorVFIOMu.Lock()
 	defer vendorVFIOMu.Unlock()
 	return vfHealth.reportSuccess(report)
+}
+
+// hasFailures reports whether a success for address could clear anything. It
+// takes only s.mu so the sentinel's routine OK reports for healthy VFs skip
+// vendorVFIOMu. Load and address errors report true so reportSuccess
+// surfaces them.
+func (s *vfHealthStore) hasFailures(address string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.loadErr != nil || !vfHealthAddressPattern.MatchString(address) {
+		return true
+	}
+	record, ok := s.records[address]
+	return ok && len(record.Failures) > 0
+}
+
+// RepairVFHealthStore retries a failed load or persist. It serializes with
+// vendor-VFIO configure, destroy, and health mutations, and returns without
+// taking those locks while the store is healthy.
+func RepairVFHealthStore() error {
+	if !VFHealthStoreUnavailable() {
+		return nil
+	}
+	vendorVFIOMu.Lock()
+	defer vendorVFIOMu.Unlock()
+
+	vfHealth.mu.Lock()
+	defer vfHealth.mu.Unlock()
+	if err := vfHealth.ensureLoadedLocked(); err != nil {
+		return err
+	}
+	return vfHealth.retryPersistLocked()
+}
+
+// VFHealthStoreUnavailable reports whether persisted state failed to load or
+// the last write failed.
+func VFHealthStoreUnavailable() bool {
+	vfHealth.mu.Lock()
+	defer vfHealth.mu.Unlock()
+	return vfHealth.loadErr != nil || vfHealth.persistErr != nil
+}
+
+// TotalQuarantinedVFs returns the number of quarantined VFs in persisted state.
+func TotalQuarantinedVFs() int {
+	vfHealth.mu.Lock()
+	defer vfHealth.mu.Unlock()
+	count := 0
+	for _, record := range vfHealth.records {
+		if record.QuarantinedAt != nil {
+			count++
+		}
+	}
+	return count
 }
 
 func (s *vfHealthStore) sortedRecordsLocked() []vfHealthRecord {
@@ -335,9 +391,6 @@ func (s *vfHealthStore) reportFailure(report VFInitFailureReport) (VFReportResul
 	if !vfHealthAddressPattern.MatchString(report.VFAddress) {
 		return VFReportResult{}, fmt.Errorf("invalid VF address %q", report.VFAddress)
 	}
-	if err := s.retryPersistLocked(); err != nil {
-		return VFReportResult{}, err
-	}
 
 	previous, existed := s.records[report.VFAddress]
 	result := VFReportResult{Failures: len(previous.Failures), Threshold: s.threshold}
@@ -348,6 +401,9 @@ func (s *vfHealthStore) reportFailure(report VFInitFailureReport) (VFReportResul
 		if sameVFAssignment(failure, report.InstanceID, report.AssignedAt) {
 			return result, nil
 		}
+	}
+	if s.persistErr != nil {
+		return VFReportResult{}, fmt.Errorf("VF health state unavailable: last write failed: %w", s.persistErr)
 	}
 
 	record := vfHealthRecord{
@@ -393,9 +449,6 @@ func (s *vfHealthStore) reportSuccess(report VFInitSuccessReport) (VFSuccessResu
 	if !vfHealthAddressPattern.MatchString(report.VFAddress) {
 		return VFSuccessResult{}, fmt.Errorf("invalid VF address %q", report.VFAddress)
 	}
-	if err := s.retryPersistLocked(); err != nil {
-		return VFSuccessResult{}, err
-	}
 	previous, ok := s.records[report.VFAddress]
 	if !ok || len(previous.Failures) == 0 {
 		return VFSuccessResult{}, nil
@@ -410,6 +463,9 @@ func (s *vfHealthStore) reportSuccess(report VFInitSuccessReport) (VFSuccessResu
 	}
 	if match < 0 {
 		return VFSuccessResult{}, nil
+	}
+	if s.persistErr != nil {
+		return VFSuccessResult{}, fmt.Errorf("VF health state unavailable: last write failed: %w", s.persistErr)
 	}
 
 	// Only the newest failure can rescind a quarantine; see VFInitSuccessReport.
--- a/lib/devices/vf_health.go
+++ b/lib/devices/vf_health.go
@@ -334,9 +334,13 @@ func (s *vfHealthStore) hasFailures(address string) bool {
 	return ok && len(record.Failures) > 0
 }
 
-// RepairVFHealthStore retries a failed load or persist. It serializes with
-// vendor-VFIO configure, destroy, and health mutations, and returns without
-// taking those locks while the store is healthy.
+// RepairVFHealthStore retries a failed load or persist so the store recovers
+// without waiting for a placement or /resources read to do it. It takes
+// vendorVFIOMu like the report paths, so a reload that changes the quarantine
+// set cannot interleave with a vendor-VFIO configure. The same retry also
+// runs from checkedAddresses under only s.mu; there, configure's own re-check
+// under vendorVFIOMu covers the selection-to-configure window instead. It
+// returns without taking either lock while the store is healthy.
 func RepairVFHealthStore() error {
 	if !VFHealthStoreUnavailable() {
 		return nil
--- a/lib/instances/vgpu_sentinel.go
+++ b/lib/instances/vgpu_sentinel.go
@@ -59,6 +59,10 @@ type VGPUSentinelController struct {
 	initFailures       metric.Int64Counter
 	quarantines        metric.Int64Counter
 	checks             metric.Int64Counter
+
+	// repairErrLoggedAt throttles the repair warning while the store stays
+	// unavailable; the gauge carries the condition between log lines.
+	repairErrLoggedAt time.Time
 }
 
 func NewVGPUSentinelController(manager Manager, meter metric.Meter, log *slog.Logger) (*VGPUSentinelController, error) {
@@ -169,8 +173,9 @@ func (c *VGPUSentinelController) probeVendorVFIO() (bool, error) {
 }
 
 func (c *VGPUSentinelController) pollOnce(ctx context.Context) {
-	if err := c.repairHealthStore(); err != nil {
+	if err := c.repairHealthStore(); err != nil && time.Since(c.repairErrLoggedAt) >= time.Minute {
 		c.log.WarnContext(ctx, "vGPU sentinel failed to repair VF health state", "error", err)
+		c.repairErrLoggedAt = time.Now()
 	}
 	targets, err := c.store.listVGPUSentinelTargets(ctx)
 	if err != nil {
@@ -345,7 +350,7 @@ func (m *manager) getVGPUSentinelTarget(_ context.Context, instanceID string) (v
 	}
 	assignedAt := ""
 	if meta.GPUClaimedAt != nil {
-		assignedAt = meta.GPUClaimedAt.UTC().Format(time.RFC3339Nano)
+		assignedAt = devices.FormatVFAssignedAt(*meta.GPUClaimedAt)
 	}
 	return vgpuSentinelTarget{
 		instanceID:     instanceID,
--- a/lib/instances/vgpu_sentinel.go
+++ b/lib/instances/vgpu_sentinel.go
@@ -1,0 +1,358 @@
+package instances
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/kernel/hypeman/lib/devices"
+	"github.com/kernel/hypeman/lib/guest"
+	"github.com/kernel/hypeman/lib/hypervisor"
+	"github.com/kernel/hypeman/lib/logger"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	vgpuSentinelPollInterval       = 5 * time.Second
+	vgpuSentinelPollTimeout        = 5 * time.Second
+	vgpuSentinelMaxConcurrentPolls = 64
+)
+
+type vgpuSentinelTarget struct {
+	instanceID     string
+	vfAddress      string
+	assignedAt     string
+	hypervisorType hypervisor.Type
+	vsockSocket    string
+	vsockCID       int64
+}
+
+type vgpuSentinelStore interface {
+	listVGPUSentinelTargets(ctx context.Context) ([]vgpuSentinelTarget, error)
+	getVGPUSentinelTarget(ctx context.Context, instanceID string) (vgpuSentinelTarget, bool, error)
+}
+
+var _ vgpuSentinelStore = (*manager)(nil)
+
+// VGPUSentinelController quarantines VFs whose guest reports a wedged driver
+// init. It polls each vendor-VFIO instance's guest agent over vsock; the
+// serial console is shared with workload output, so nothing read from logs is
+// trusted. The health store deduplicates repeated reports per assignment, so
+// polling is idempotent and a failed persist retries once on the next tick.
+type VGPUSentinelController struct {
+	store              vgpuSentinelStore
+	log                *slog.Logger
+	interval           time.Duration
+	repairHealthStore  func() error
+	reportFailure      func(devices.VFInitFailureReport) (devices.VFReportResult, error)
+	reportSuccess      func(devices.VFInitSuccessReport) (devices.VFSuccessResult, error)
+	guestGPUInitStatus func(ctx context.Context, target vgpuSentinelTarget) (guest.GPUInitState, string, error)
+	discoverFramework  func() (devices.VGPUFramework, []devices.VirtualFunction, error)
+	initFailures       metric.Int64Counter
+	quarantines        metric.Int64Counter
+	checks             metric.Int64Counter
+}
+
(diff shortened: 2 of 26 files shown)


lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,395 @@
...
+	if err != nil {
+		c.log.WarnContext(ctx, "vGPU sentinel could not confirm assignment before conviction",
+			"vf", target.vfAddress, "instance_id", target.instanceID, "error", err)
+		return false
+	}
+	if !ok || current.vfAddress != target.vfAddress || current.assignedAt != target.assignedAt {
+		c.log.InfoContext(ctx, "vGPU sentinel skipping conviction: assignment changed during scan",
+			"vf", target.vfAddress, "instance_id", target.instanceID)
+		return false
+	}
```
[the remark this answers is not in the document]
yummybomb: true — fixed in b3bf925. the recheck now skips only when a *different* assignment exists; a released assignment (stop/delete after the marker was read) convicts. the archive-before-persist ordering in start ties a marker read under an epoch's snapshot to that epoch's VF, and releasing the assignment can't un-wedge it, so the conviction is valid.

one note on severity: the common version of "operator stops the failing instance" isn't this race — once stop clears the assignment before a scan pass, the instance drops out of the target list entirely and the marker is never read (the v1 coverage bound; detection defers to the next victim boot). the recheck only added a drop in the sub-second window between the metadata snapshot and the recheck, which is what this fixes.

lib/devices/vf_health.go
```diff
@@ -0,0 +1,203 @@
...
+		VFAddress:     q.VFAddress,
+		InstanceID:    q.InstanceID,
+		SentinelLine:  q.SentinelLine,
+		WedgeCount:    1,
+		QuarantinedAt: time.Now().UTC(),
+	}
+	s.records[q.VFAddress] = record
+	if err := s.persistLocked(); err != nil {
+		return record, false, err
+	}
```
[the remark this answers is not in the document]
yummybomb: outdated — this was true at the commit reviewed but fixed in 38249dc: quarantine() now rolls the record back from memory when persist fails and returns the error, the sentinel leaves the tail open on a failed quarantine, and the guest agent re-emits the marker every 30s, so the next report retries the write (covered by TestVGPUSentinelControllerRetriesFailedQuarantine). IsVFQuarantined no longer exists.

lib/devices/vf_health.go
```diff
@@ -0,0 +1,213 @@
...
+	defer s.mu.Unlock()
+	if err := s.ensureLoadedLocked(); err != nil {
+		return false, err
+	}
+	if _, ok := s.records[vfAddress]; !ok {
+		return false, nil
+	}
+	delete(s.records, vfAddress)
+	return true, s.persistLocked()
+}
```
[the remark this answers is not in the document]
yummybomb: outdated — ClearVFQuarantine was removed in 44b8a12 (no callers). the only clear path is the documented runbook: edit vf-health.json and restart hypeman immediately, so memory and disk can't diverge.

sjmiller609 (commented):
Overall looks good, a few small feedbacks and improvements we should consider.

- **Versioning:** use a versioned envelope for future counters and recovery state:

```json
{"version": 1, "records": [...]}
```

- **Failure threshold:** currently 1 detection quarantines a VF. Should this be configurable? e.g. quarantine after failures from N distinct VM assignments, default 2. Also we could have a sucess reset: successful GPU-initialization sentinel that clears the VF’s failure count. I think we should do this if there's any case where this error happens even when the VF is not wedged.

- **Resources API:** expose allocatable and quarantined slots so reported capacity matches admission:

```diff
 "total_slots": 64,
 "used_slots": 5,
+"allocatable_slots": 57,
+"quarantined_slots": 2
```

```diff
 "profiles": [{
   "name": "NVIDIA L40S-12Q",
-  "available": 59
+  "available": 57
 }]
```

- **GPU-less hosts:** exit the controller when no GPU framework exists; continuous five-second discovery is unnecessary.

- **Terminology:** replace “convict/convictions” terminology, that seems confusing.

- **Cleanup follow-up:** acknowledged that automated GPU reset and cleanup are intentionally deferred; this PR handles detection and quarantine only.

yummybomb:
Addressed the review feedback (commits be500500, 7b248553):

- **Versioning**: `vf-health.json` is now `{"version": 1, "records": [...]}`; unknown versions fail closed like any other invalid state.
- **Failure threshold**: quarantine now requires failures from `gpu.vf_quarantine_threshold` distinct instance assignments, default 2, deduped per assignment so rescans/restarts can't double-count. Success reset is in too: the guest agent probes driver init at boot with `nvidia-smi -L` (the device open runs RmInitAdapter, so on a wedged VF the probe also forces detection instead of waiting for the workload) and emits `HYPEMAN-GPU-INIT-OK` on success, which clears the VF's failure tally. It never clears a quarantine — that stays manual.
- **Resources API**: `allocatable_slots` and `quarantined_slots` are now in the `/resources` GPU status; profile `available` already excluded quarantined VFs.
- **GPU-less hosts**: the controller exits when discovery finds no vGPU framework instead of re-probing every 5s (still retries on discovery *errors* so a transient sysfs failure doesn't kill it).
- **Terminology**: convict/conviction is gone; the metric split into `vgpu_sentinel_init_failures_total` (below-threshold reports, the early signal for systemic non-wedge failures) and `vgpu_sentinel_quarantines_total`.
- **Cleanup follow-up**: unchanged, force-cycle endpoint stays a follow-up.

Still pending before merge: the live end-to-end run on the dev GPU host, which should now cover the threshold path (two victim boots) and the OK probe.

lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,418 @@
...
+	if err != nil {
+		c.log.WarnContext(ctx, "vGPU sentinel scan failed to read app log",
+			"instance_id", target.instanceID, "error", err)
+		return
+	}
+	switch event {
+	case VGPUSentinelEventInitFailed:
+		tail.done = c.handleFailure(ctx, target, line)
+	case VGPUSentinelEventInitOK:
+		c.handleSuccess(ctx, target)
```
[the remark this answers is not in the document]
yummybomb: fixed in 9d2f3fc7 — a below-threshold recorded failure no longer sets the tail done, so the scan keeps reading and a later `HYPEMAN-GPU-INIT-OK` on the same boot clears the tally. the store's per-assignment dedup makes the continued rescans report nothing. covered by `TestVGPUSentinelControllerClearsTallyOnOKAfterRecordedFailure`.

lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,418 @@
...
+		return
+	}
+	if !unchanged {
+		return
+	}
+	cleared, err := c.reportSuccess(target.vfAddress)
+	if err != nil {
+		c.log.WarnContext(ctx, "vGPU sentinel failed to clear recorded init failures",
+			"vf", target.vfAddress, "instance_id", target.instanceID, "error", err)
+		return
```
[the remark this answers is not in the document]
yummybomb: fixed in 9d2f3fc7 — a clear whose persist fails is remembered as `pendingSuccess` on the tail and retried at the start of every subsequent scan, so the once-per-boot OK marker can't be lost to an advanced offset. covered by `TestVGPUSentinelControllerRetriesFailedTallyClear`.

yummybomb (commented):
found three issues that should be addressed before merge: two false-positive quarantine paths and the unbounded `nvidia-smi` wait. the targeted vGPU/store/sentinel tests pass; the current linux CI failure is the unrelated `TestWaitForState_ChannelClosedOnDelete` timing failure.

yummybomb (commented):
revising severity: these are technically reproducible edge cases, but neither should block this PR.

- `lib/instances/vgpu_sentinel.go:244-248` — non-blocking follow-up: a confirmed `OK` retry can be deleted only if the VF-health write fails *and* the same instance is reassigned between adjacent repeated `OK` lines. reusing an existing pending entry would close that resilience edge, but this is a compounded failure rather than a normal-path correctness issue.

- `lib/instances/vgpu_sentinel.go:343-388`, `lib/instances/logs.go:137-190` — non-blocking follow-up: copytruncate can move a marker to `app.log.1` between sentinel scans. for this to affect the one-shot `OK`, the active log must already exceed the 50MB rotation threshold and the successful probe must land in the narrow pre-rotation window. worth making rotation-aware, but not a merge blocker for the initial quarantine path.

both repros prove the edge cases exist; they do not justify the blocker severity I originally assigned. the core failure detection, assignment revalidation, success ordering/rescission, and bounded probe paths look sound.

yummybomb (commented):
one blocking test-quality issue. the sentinel/store boundaries, fail-closed placement behavior, assignment revalidation, and persistence rollback paths otherwise look coherent in context. targeted race tests and go vet pass locally; the current linux CI also exposes the changed-test failure below.

lib/system/guest_agent/gpu_watch_test.go
```diff
@@ -0,0 +1,217 @@
...
+		"[ -f \"$COUNT_PATH\" ] && count=$(/bin/cat \"$COUNT_PATH\")\n" +
+		"count=$((count + 1))\n" +
+		"echo $count > \"$COUNT_PATH\"\n" +
+		"[ $count -eq 1 ] && exec /bin/sleep 10\n" +
+		"exit 0\n"
+	nvidiaSMI := filepath.Join(binDir, "nvidia-smi")
+	require.NoError(t, os.WriteFile(nvidiaSMI, []byte(script), 0o755))
+	t.Setenv("COUNT_PATH", countPath)
+
+	probeGPUInitUntil(&gpuInitReporter{}, nvidiaSMI, time.Now().Add(time.Second), 10*time.Millisecond, 0)
```
yummybomb: **blocker:** this 10ms deadline is shorter than normal shell startup under race/loaded CI, so the nominally successful second attempt can time out too. The current Linux job failed here with `actual: 3`, and `go test -race ... -count=100` reproduces it locally. Please make the attempt runner/clock injectable and use a deterministic fake, or otherwise give the success attempt independent scheduling margin; asserting exactly two real process launches with this deadline is flaky.

yummybomb: fixed in 88f3c3d6 by injecting the probe attempt into the retry loop and testing the timeout/success sequence directly. `go test -race ./lib/system/guest_agent -run TestProbeGPUInitRetriesAfterAttemptTimeout -count=100`, the package race suite, and `go vet` pass.

lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,414 @@
...
+		if _, ok := alive[id]; !ok {
+			delete(c.tails, id)
+		}
+	}
+}
+
+func (c *VGPUSentinelController) scanTarget(ctx context.Context, target vgpuSentinelTarget) {
+	tail := c.tails[target.instanceID]
+	if tail == nil || tail.vfAddress != target.vfAddress || tail.assignedAt != target.assignedAt {
+		tail = &vgpuSentinelTail{vfAddress: target.vfAddress, assignedAt: target.assignedAt}
```
yummybomb: The tail offset is process-local, but a successful report removes the only persistent dedup record. After `FAILED -> OK`, restarting hypeman starts at offset 0, records the historical failure again, increments `init_failures_total`, then clears it again on the historical OK. With `vf_quarantine_threshold: 1`, replay also emits a fresh quarantine/rescind and increments both counters on every restart; if that replayed clear fails, a healthy VF stays quarantined. Please retain a bounded last-resolved assignment/watermark per VF (or coalesce a replay before mutating state) and add a restart regression test.

lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,414 @@
...
+	if info.Size() < tail.offset {
+		tail.offset = 0
+		tail.skippingLongLine = false
+	}
+	if _, err := f.Seek(tail.offset, io.SeekStart); err != nil {
+		return err
+	}
+
+	reader := bufio.NewReaderSize(f, vgpuSentinelMaxLineBytes)
+	for {
```
yummybomb: non-blocking: please thread `ctx` into `scanSentinelLog` and check cancellation between reads (and target scans). On the first pass after controller/process start every assignment begins at byte 0, so a host with many logs near the 50 MB rotation limit can keep `Run` inside this loop and hold the process-level `errgroup.Wait` well after shutdown was requested.

yummybomb (commented):
two code-quality findings. the store/placement boundaries, persistence rollback, assignment revalidation, and tests otherwise look coherent in repo context. the new functions peak at cyclomatic complexity 16 (`loadLocked` / `selectLeastLoadedVF`); those paths remain cohesive and well-covered, so I don't see mechanical decomposition as a merge requirement.

lib/instances/vgpu_sentinel.go
```diff
@@ -0,0 +1,414 @@
...
+)
+
+const (
+	vgpuSentinelScanInterval = 5 * time.Second
+	vgpuSentinelMaxLineBytes = 64 * 1024
+)
+
+// Require a standalone guest-agent line so echoed marker text cannot count against a VF.
+var (
+	vgpuSentinelFailedPattern = regexp.MustCompile(`^\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2} \[guest-agent\] (HYPEMAN-GPU-INIT-FAILED ts=\S+ nvrm="NVRM: [^"\r\n]*RmInitAdapter failed![^"\r\n]*")\r?\n?$`)
```
yummybomb: **blocker:** a standalone line does not establish that the guest agent wrote it. In exec mode the customer entrypoint's stdout/stderr is wired directly to the same serial console (`lib/system/init/mode_exec.go:132-133`), so an ordinary workload can print a complete line matching this regex without root or `/dev/kmsg` access. Repeating that across assignments can quarantine healthy VFs, and the runbook's statement that only a root guest can forge this is therefore incorrect. Please carry the report over a channel whose source the host can distinguish (for example a dedicated guest-agent RPC/transport), or otherwise authenticate the source; matching the log shape is not sufficient here.

lib/system/guest_agent/gpu_watch.go
```diff
@@ -0,0 +1,217 @@
...
+	done := make(chan error, 1)
+	go func() { done <- cmd.Wait() }()
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+	select {
+	case err := <-done:
+		return err
+	case <-timer.C:
+		_ = cmd.Process.Kill()
```
yummybomb: this bounds how long the caller waits, but not the attempt's lifetime. If `nvidia-smi` is stuck in the uninterruptible ioctl described above, `Kill` does not make `cmd.Wait` return; the process and waiter goroutine remain live while `probeGPUInitUntil` launches another attempt after 15s. The 10-minute loop can therefore accumulate many concurrent stuck probes on the VF. Can we retain the in-flight attempt and avoid launching another until it is reaped (or stop retrying once an attempt cannot be killed), so there is at most one outstanding `nvidia-smi`?

yummybomb (commented):
reviewed the change against `hypeship/vendor-vfio-vgpu` and traced the sentinel through guest output, host log scanning, durable health state, placement, and admission. the persistence rollback and assignment-identity handling are coherent, but the serial marker is not actually source-authenticated, and timed-out probes can overlap in the exact uninterruptible-ioctl case this code targets.

leaving the first finding as a blocker; github does not allow this account to request changes on its own PR.

yummybomb (commented):
a

sjmiller609 (approved):

PR #436, opened 2026-08-20
Secure mediated VM egress

sjmiller609:
## Summary
- enforce direct TCP blocking by instance source address instead of TAP interface name
- reject proxy requests from unregistered VM source addresses
- resolve and dial only public upstream addresses through the built-in proxy

## Why
Bridge forwarding does not preserve the TAP device as the logical input interface, so interface-matched enforcement could be bypassed. The proxy also accepted unregistered VM sources and could dial private or special-purpose addresses.

## Testing
- `go test -race ./lib/egressproxy`
- `go test ./lib/instances -run '^(TestValidateCreateRequest_|TestRestoreEgress)' -count=1`\n- `go vet ./lib/egressproxy ./lib/instances`\n\nThe full instances integration suite was not run locally because it requires root networking and host image-conversion dependencies; CI covers it.

---

> [!NOTE]
> **High Risk**
> Changes host iptables enforcement and proxy SSRF/source-auth behavior for VM egress, which is security-critical networking. A rule or resolver bug could leak host/private access or drop legitimate traffic.
> 
> **Overview**
> Hardens mediated VM egress so guests cannot bypass the proxy or reach private/special-purpose addresses.
> 
> Linux iptables rules now match each instance **source IP** instead of TAP interface name (which is lost after bridge forwarding). They also drop spoofed TAP traffic, reject new TCP to the host except DNS and the proxy port in `all` mode, and use conntrack NEW state.
> 
> The proxy itself **403s unregistered source IPs** and dials only public global-unicast destinations (`publicDialContext`), preferring IPv4 when both families resolve. Mixed public/private DNS answers are treated as blocked.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e69387c18b701aae5ff0ddda07da472373a31035. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/egressproxy/README.md
+++ b/lib/egressproxy/README.md
@@ -63,7 +63,9 @@ Operationally, this is intended for key rotation, revocation/reissue flows, and 
 
 - Real secret values are persisted in the normal instance `env` metadata, which is already host-side state.
 - TLS interception requires guest trust of the proxy CA; hypeman installs this CA in the guest when proxy mode is enabled.
-- Egress enforcement is applied per instance TAP device and removed when the instance stops/standbys/deletes.
+- Proxy requests are accepted only from source addresses registered to running instances.
+- Upstream resolution rejects private, loopback, link-local, carrier-grade NAT, documentation, benchmark, multicast, and reserved addresses.
+- Egress enforcement matches each instance source address and is removed when the instance stops/standbys/deletes.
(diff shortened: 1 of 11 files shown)


sjmiller609:
Closing because Quickbox does not need a second egress-enforcement boundary for the employee beta. We will rely on Hypeman’s existing VM isolation and revisit mediated egress only if outbound policy or credential injection becomes a product requirement.

PR #456, opened 2026-08-26
Materialize per-layer artifacts with explicit whiteout handling

chruffins:
## tldr

Content-addressed per-layer artifact store at `images/layers/<layer-blob-digest>/`. Pulls and composition (next PR in the stack) share it. Native layer artifacts are Linux-root-only; rootless Linux and macOS use the merged-rootfs path instead, with no image-content change.

## what this adds

1. **Materialization** (`materializeLayerArtifact`) — unpacks a layer blob from the existing shared OCI cache with umoci, converts to erofs, installs `layer.erofs` + an `artifact.json` record atomically. Same layer shared across images materializes once; concurrent callers share one build (singleflight), with a manager-level concurrency bound.
2. **Whiteouts as overlayfs** — on supported Linux hosts, `.wh.<name>` becomes a `0:0` char device and `.wh..wh..opq` becomes a `trusted.overlay.opaque` xattr. This is the standard representation required by later overlayfs stacking; it is not attempted where the host cannot create it.
3. **Platform fallback contract** — a startup probe verifies Linux `mknod` and trusted overlay xattr support on the layer-store filesystem. Rootless Linux, macOS, and unsupported filesystems skip per-layer artifact materialization and compose OCI layers into one rootfs with `DirRootfs`, which applies whiteouts in userspace. This preserves image semantics while giving up per-layer sharing on those hosts.
4. **Hardened extraction** — umoci confines entries, blob digest + diff ID verified, stream capped at 100 GiB, detached builds bounded by a 1h deadline.
5. **Disk accounting** — the layer store, including records and in-progress extraction/install trees, is counted in `TotalOCICacheBytes` (conservative: blob + artifact of the same layer may transiently coexist). Active builds keep totals uncached, and disk-usage scans retry if invalidated while a scan is in progress.

## review status: done

Two independent review passes (correctness + code-quality). Findings and how they are handled:

- Hung blob read could wedge the singleflight key forever → 1h deadline on detached builds
- Cache blob integrity was never verified → unpack now checks the compressed digest
- Rootless/macOS whiteout extraction could not produce valid overlayfs artifacts → unsupported hosts now use the merged-rootfs fallback
- Disk-usage invalidation could return stale totals → scans retry after a concurrent invalidation
- Same-size artifact corruption was reusable → records now include and verify an artifact checksum
- Detached layer builds were unbounded across distinct digests → a manager-level semaphore limits them
- Compressed blob verification could miss decoder-unconsumed suffixes → the raw blob is hashed independently
- Layer-build cleanup could race disk accounting → cleanup completes before the build is marked inactive
- Conflicting descriptor encodings could share one artifact path → records store the normalized encoding and reject mismatches
- Dead/simplification cleanup: `composeOnDiskFormat` + `whiteoutPrefix` moved to tests, `multiCloser` deleted, blob path via `paths.OCICacheBlob`

Verified empirically on a dev host: `mkfs.erofs` 1.8.10 and `mkfs.ext4` both preserve whiteout char devices and opaque xattrs through `ExportRootfs` on the supported Linux-root path. Depends on host erofs-utils version; check it in the pull-integration PR.

## reviewer focus (3 things)

1. `layerBuildTimeout` = 1h is a judgment call — flag if hypeman has a convention for bounding background work
2. Unpack now hard-fails on blob digest mismatch — new error if any path intentionally writes unverified blobs
3. PR `#458` wires the store into production. It skips layer materialization on unsupported hosts and retains blob-based merged-rootfs composition as the fallback.

## validation

Full `lib/images` + `lib/paths` green on a rootless host; capability-gated whiteout/device tests pass or skip based on an actual `mknod` + trusted-xattr probe. Tests cover: materialize/reuse/rebuild, corrupt-record recovery, same-size artifact corruption, invalid digest rejection (`sha256:..`, `sha256:a/b`), whiteout/opaque output and cross-layer semantics, symlink confinement and replacement, hardlinks, cancellation, zstd + docker media types, concurrent singleflight (8 callers, one build), trailing-padding diff IDs. The exported erofs whiteout round trip moves to the composition PR, which is where the host erofs-utils version gets checked.

CI note: the `test` workflow is flaky on this branch — failures are unrelated integration tests (`TestEgressProxyRewritesHTTPSHeaders` timeout, docker/network tests needing `nginx:alpine`), failing before the latest commits too. Layer-artifact tests pass consistently.

## next step

Review the focus items above, then approve. PR `#458` is the production caller and lifecycle follow-up.

---

> [!NOTE]
> **Medium Risk**
> New layer unpack and filesystem materialization touches tar extraction and external mkfs tools; incorrect whiteout or path handling could affect future image composition, though pull behavior is unchanged until wired in.
> 
> **Overview**
> Introduces a **content-addressed per-layer artifact store** under `images/layers/<blob-digest>/`, keyed by compressed layer digest plus format (erofs/ext4). **`materializeLayerArtifact`** reads blobs from the existing OCI cache, unpacks with umoci into a temp dir, converts to the default disk format, and installs `layer.*` plus an `artifact.*.json` record atomically; duplicate work is deduped with **singleflight** and valid on-disk records are reused.
> 
> **Whiteouts** are modeled explicitly: artifact extraction uses umoci **`OverlayfsRootfs`** (character-device whiteouts and opaque xattrs) so stacked layers can be mounted later; compose-oriented unpacking uses **`DirRootfs`**. Layer unpack supports gzip/zstd, verifies diff IDs, caps unpacked size at 100 GiB, and honors **context cancellation** via a `contextReader` and **`CommandContext`** for `mkfs.erofs`.
> 
> **Disk accounting** now walks `layer.*` files (skipping `.unpack-*` temps) and folds that into **`TotalOCICacheBytes`** / resource admission alongside OCI blobs. New path helpers live in `lib/paths`. The materialization API is implemented and tested; production wiring is in follow-up PR `#458`, which skips it on unsupported hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4ed10542efd0080115dac657b639703fc029953. Configure [here](https://cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/images/README.md
+++ b/lib/images/README.md
@@ -65,6 +65,10 @@ Content-addressable storage with tag symlinks (similar to Docker/Unikraft):
         rootfs.erofs
       latest -> abc123def456...   # Tag symlink to digest
       3.18 -> def456abc123...     # Another tag
+    layers/                       # Shared materialized layer artifacts
+      abc123def456.../
+        layer.erofs
+        artifact.erofs.json
   system/
     oci-cache/              # Shared OCI layout for all images
       index.json            # Manifest index with digest-based tags
@@ -82,6 +86,7 @@ Content-addressable storage with tag symlinks (similar to Docker/Unikraft):
 - Natural hierarchy: All versions of an image grouped under repository
 - Easy inspection: Clear which digest belongs to which image
 - Layer caching: All images share the same blob storage, layers deduplicated automatically
+- Materialized layer artifacts are reference-protected and reconciled by the layer lifecycle manager; stale temporary trees are age-gated before removal.
 
 **Design:**
 - Images stored by manifest digest (content hash)
@@ -92,6 +97,7 @@ Content-addressable storage with tag symlinks (similar to Docker/Unikraft):
 - Shared blob storage enables automatic layer deduplication across all images
 - Orphaned digests are automatically deleted when the last tag referencing them is removed
 - Symlinks only created after successful build (status: ready)
+- Disk accounting uses logical file sizes, matching image metadata and storage admission rather than filesystem block allocation.
 
 ## Reference Handling (reference.go)
 
--- a/lib/images/disk.go
+++ b/lib/images/disk.go
@@ -1,7 +1,9 @@
 package images
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -29,15 +31,20 @@ var DefaultImageFormat = func() ExportFormat {
 	return FormatErofs
 }()
 
-// ExportRootfs exports rootfs directory in specified format (public for system manager)
+// ExportRootfs exports rootfs directory in specified format (public for system manager).
 func ExportRootfs(rootfsDir, outputPath string, format ExportFormat) (int64, error) {
+	return ExportRootfsWithContext(context.Background(), rootfsDir, outputPath, format)
+}
+
+// ExportRootfsWithContext exports rootfs directory and cancels external formatters with ctx.
+func ExportRootfsWithContext(ctx context.Context, rootfsDir, outputPath string, format ExportFormat) (int64, error) {
 	switch format {
 	case FormatExt4:
-		return convertToExt4(rootfsDir, outputPath)
+		return convertToExt4(ctx, rootfsDir, outputPath)
 	case FormatErofs:
-		return convertToErofs(rootfsDir, outputPath)
+		return convertToErofs(ctx, rootfsDir, outputPath)
 	case FormatCpio:
-		return convertToCpio(rootfsDir, outputPath)
+		return convertToCpio(ctx, rootfsDir, outputPath)
 	default:
 		return 0, fmt.Errorf("unsupported export format: %s", format)
 	}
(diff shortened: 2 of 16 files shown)


chruffins:
@cursor review

sjmiller609 (approved):
**lib/images/layer_artifact.go:90–99** — what about OCI nondistributable gzip/zstd media types? they currently fall through to raw tar and fail extraction, although the existing unpacker supports them.
**GC follow-up:** fine to leave GC out of this PR, but important to land it after merging. otherwise unused artifacts and crash-leftover temporary directories persist indefinitely.
**benchmarking:** the materialization path isn’t active in production yet. before enabling it, will need some benchmarking.

chruffins:
<p>ran it on a test machine as root, using:</p>
<ul>
<li>4 gzip OCI layers</li>
<li>16 MiB uncompressed per layer, 64 MiB total</li>
<li><code>/mnt/data</code> storage</li>
<li>3 cold runs plus warm-cache runs</li>
</ul>

results
- current merged-rootfs baseline | 1.73s
- cold per-layer materialization | 3.05s
- warm materialization | 51.9ms

</div>
<p>cold materialization is ~1.8× slower for one image because it creates four separate EROFS artifacts instead of one merged filesystem. The expected win appears when subsequent images reuse those layers: warm reuse was ~52ms for all four layers.</p> seems like there's some juice to squeeze here, but will add it in a follow-up!

PR #458, opened 2026-08-26
Shared-layer lifecycle: accounting, eviction, recovery, materialization

chruffins:
## summary

Adds the runtime consumer for shared image layers while preserving Hypeman's two-disk VM contract.

- **Shared base disks**: images with multiple layers compose every layer except the final layer into a content-addressed read-only base disk under `images/bases/`.
- **Per-instance final layer**: the final layer is materialized into the VM's writable overlay disk under `upper/`. The VM still receives one read-only rootfs disk and one writable overlay.
- **Build fallback**: unsupported hosts, single-layer images, corrupt artifacts, and failed materialization use the existing flattened rootfs path.
- **Artifact lifecycle**: final-layer artifacts and shared bases use in-flight leases, keyed singleflight, grace-period cleanup, reference scanning, startup recovery, and disk accounting.
- **Overlay capacity**: image data is added to the physical overlay allocation while the caller-facing writable capacity remains unchanged. Admission reservations use the physical size.
- **Promotion and recovery**: layered rootfs links are recreated relative to their destination during legacy-to-content promotion, and runtime manifest lookup follows the authoritative image layout.
- **Observability**: image build phases distinguish shared-base export from flattened filesystem export.
- **CI reliability**: EROFS artifacts are extracted with `fsck.erofs` instead of loop-mounted, layered overlay roots retain searchable `0755` permissions, and interrupted netlink dumps retry under parallel test load.

## validation

- Full CI is green: Linux tests, Darwin tests, and install E2E pass.
- Targeted Linux end-to-end validation confirms nginx content is reachable through the layered rootfs and ingress path.
- Image package tests pass on a root-capable Linux host with EROFS support.
- Rootless local package tests cover manifest validation, shared-base hashing, lifecycle accounting, and legacy promotion; native mount tests require host privileges.
- Storage and disk-I/O benchmarks show the composition prerequisite does not increase the flattened path's disk footprint or I/O. Representative shared-base runtime benchmarks remain required before production rollout.

## rollout

Do not enable this path in production until storage and disk-I/O benchmarks are run against representative multi-layer images. The flattened path remains the fallback for hosts without native layer-artifact support.

- **Host-storage alert fix**: runtime overlay preparation no longer creates transient read-only loop mounts for EROFS layer artifacts, avoiding false full-filesystem alerts on the host.


chruffins:
closing as superseded. the next implementation will target content-addressed artifacts for every OCI layer and compose them as multiple read-only OverlayFS lower layers, rather than exporting flattened shared prefixes. benchmark coverage will be added before enabling the new runtime path.

PR #465, opened 2026-08-28
Deslop wedged-VF detection: dedupe sentinel report guard, trim redundant tests

yummybomb:
Cleanup pass on #435, stacked on `hypeship/vgpu-wedge-quarantine`. No behavior changes to the detection path; net -130 lines, mostly tests.

## Production code

- `vgpu_sentinel.go`: `handleFailure`/`handleSuccess` duplicated the confirm-assignment preamble (confirm → warn on error → skip on change). Folded the logging into `confirmAssignment`, which now returns a bool and tags its log lines with an `action` attribute. Logging-only difference: the success path now also logs at info when it skips a changed assignment (previously silent).
- `gpu_watch.go`: moved the `guestServer.GetGPUInitStatus` RPC method out from between two `gpuInitReporter` methods so the reporter methods are contiguous. Pure move.
- `process_identity.go` / `storage.go`: trimmed two comments (one explained the diff rather than the code, one enumerated its callers).

## Tests

- `vgpu_sentinel_test.go` (-177/+... net ~-120):
  - The fake `reportFailure` re-implemented the health store's per-assignment dedup, and `ReportsFailureOnce` then asserted that fake's own behavior. Dedup is the store's contract and is covered in `lib/devices`; the fake is now a plain recorder and the test (renamed `ReportsFailure`) keeps the UNKNOWN-not-reported and report-fields coverage, absorbing the NVRM log assertions from the deleted `LogsNVRMMessageOnQuarantine`.
  - Dropped `PollsTargetsConcurrently` — it exercised `errgroup.SetLimit`, not sentinel logic.
  - `RepairsHealthStoreOncePerPoll` proved once-per-poll with 64 targets; 2 targets prove the same thing.
  - Merged the three changed/released-assignment tests into table-driven `ConfirmsAssignmentBeforeReporting` covering all four state×assignment combinations (the failed×changed case is new).
  - Removed a brittle log-content assertion from `RetriesFailedTallyClear`.
- `vf_health_test.go`: the no-op-success test spun 64 goroutines, but nothing is contended — the assertion (no persist retry on a failed store) holds with one call. Kept its unique repair assertion (exactly one persist, two dir syncs).
- `gpu_watch_test.go`: folded success-is-terminal into `TestGPUInitReporterState` and dropped `ProbeGPUInitSkipsWithoutNvidiaSMI` (asserted only an empty log buffer on a trivial early return).

## Deliberately not changed

- Constructor nil-checks and the `vgpuSentinelStore` downcast in `NewVGPUSentinelController` — same pattern as `NewHealthCheckController`, so they match package idiom (kept `TestNewVGPUSentinelControllerRejectsUnsupportedManager` with them).
- `create.go` setting `GPUProfile` outside `setStoredVGPUDevice` — deliberate: `clearStoredVGPUDevice` keeps the profile for the next start.
- All fail-closed semantics (terminal success, assignment revalidation, no per-report persist retry, uncertain-liveness preservation).

## Testing

```
go test -race ./lib/system/guest_agent ./lib/guest ./lib/devices -count=1
go test -race ./lib/instances -run 'Test(VGPUSentinel|ListVGPUSentinel|ProbeGPUInit|RunGPUProbeAttempt|GPUInitReporter)' -count=1
go test ./lib/devices ./lib/providers
go vet ./lib/system/guest_agent ./lib/guest ./lib/instances ./lib/devices ./lib/providers
```

All pass locally (embedded VMM/agent binaries stubbed for compilation, same constraint as the parent PR).

### DIFF
--- a/lib/devices/vf_health_test.go
+++ b/lib/devices/vf_health_test.go
@@ -3,10 +3,8 @@ package devices
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -328,7 +326,7 @@ func TestReportVFInitSuccessWithoutMatchingFailureClearsNothing(t *testing.T) {
 	assert.False(t, result.Rescinded)
 }
 
-func TestReportVFInitSuccessNoopFanoutDoesNotRetryFailedPersist(t *testing.T) {
+func TestReportVFInitSuccessNoopDoesNotRetryFailedPersist(t *testing.T) {
 	resetVFHealthStore(t)
 	var syncCalls int
 	vfHealth.mu.Lock()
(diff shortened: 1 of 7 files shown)


yummybomb:
folding these changes directly into #435 instead

PR #466, opened 2026-08-28
Unify vGPU retention wrapping and dedupe test fixtures

yummybomb:
## Summary

Cleanup pass on the vendor VFIO vGPU lifecycle layer (#321), no behavior change:

- **One retention mechanism in start.** The start-path rollback for a failed vGPU create previously hand-built `VGPUCleanupPendingError` in two places instead of going through `vgpuRetention` like create does. The branch now records retention via `markRetained` and lets the existing deferred `wrapPending` produce the error, so the pending-error construction and the retained/unretained metric each live in one place. Error text, `Retained` flag, metric values, and wrapping order are unchanged.
- **Inlined `deferWrapPending`.** It was a one-line adapter around `wrapPending`; the ordering constraint it documented belongs at the two call sites, which now carry it.
- **Deduped test fixtures.** The identical 5-field vendor VFIO `VGPUDevice` literal appeared three times in `vgpu_test.go`; extracted `testVendorVFIODevice`.

Reviewed the rest of the diff for dead branches and redundant tests and deliberately left it alone: the per-callsite retention-stub rejection tests match the four independent guards, the reconcile tests each cover a distinct fail-closed path, and the `interface{ StartVGPUReconciler }` assertion in `cmd/api/main.go` follows the file's existing idiom for optional manager methods.

## Testing

- `go build ./...`, `go vet ./...` clean
- `go test -race` passes for `lib/devices`, `lib/hypervisor/qemu`, and the `lib/instances` vGPU/retention/reconcile/lifecycle suites
- `cmd/api/api` and `lib/instances` integration failures in this environment are identical on the unmodified base branch (network/hypervisor privileges), per-test diff verified

### DIFF
--- a/lib/instances/create.go
+++ b/lib/instances/create.go
@@ -280,7 +280,8 @@ func (m *manager) createInstance(
 	var gpuAssignedAt *time.Time
 	retention := vgpuRetention{instanceID: id}
 
-	defer retention.deferWrapPending(&retErr)
+	// Deferred before cu.Clean so rollback records retention before this wraps the error.
+	defer func() { retErr = retention.wrapPending(retErr) }()
 	cu := cleanup.Make(func() {
 		log.DebugContext(ctx, "cleaning up instance on error", "instance_id", id)
 		m.persistVGPURetention(ctx, &retention)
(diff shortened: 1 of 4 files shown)


yummybomb:
Closing because this cleanup commit was pushed directly onto #321.

PR #478, opened 2026-09-14
Move image rootfs onto / in guest init instead of chroot

sjmiller609:
## summary

- `lib/system/init`: replace the `chroot(2)` into `/overlay/newroot` with a `switchRoot` that moves the overlay mount onto `/` (`MS_MOVE`, as `switch_root(8)` does) and then chroots into it, in both exec and systemd mode
- give the async kernel-headers worker its own mount namespace so its pre-switch paths (`/kernel-headers.tar.gz`, `/overlay/newroot/...`) keep resolving after init moves the overlay
- add `TestNestedDockerExecRoot`: boots `docker:28.5.2-dind` with `dockerd` as the entrypoint, runs a container from a busybox-only image imported from the dind rootfs (no registry access), and checks `docker exec` sees the container rootfs; adds the image to the test prewarm list

## why

`setns(2)` into a mount namespace resets the caller's root to that namespace's root (`mntns_install` in `fs/namespace.c`). Guest init only `chroot(2)`ed into the image overlay, so the mount namespace root stayed the initrd. Any runc-based runtime inside the guest inherits that: `docker run` works (PID 1 gets its root via `pivot_root`), but `docker exec` joins the container's mount namespace with `setns` and lands in the initrd instead of the container rootfs.

Making the image overlay the namespace root fixes this for every image without patching the runtime inside the image. Only newly booted instances pick it up; snapshots taken with the old init keep the old layout until re-created.

## testing

on a KVM dev box, patched init:

- `TestNestedDockerExecRoot` passes
- `TestSystemdMode` (systemd PID 1, guest agent service, journal, serial log rotation) passes
- exec, cp, standby, and stop/start API tests pass when run individually (a parallel group run hit GitHub 504s downloading the guest kernel, unrelated)
- ad-hoc check on a debian instance: `/` is the overlay in `/proc/self/mountinfo`, `/overlay` is no longer visible, `/run/hypeman/kernel-headers.status` reaches `ready`, `/usr/src/linux-headers-$(uname -r)` and the `/lib/modules/.../build` symlink exist, and the console log has `HYPEMAN-HEADERS-READY`
- `go test ./lib/system/init/` passes

negative control with main's init, same test:

```
+ docker exec nested /bin/busybox cat /runtime-created
cat: can't open '/runtime-created': No such file or directory
--- FAIL: TestNestedDockerExecRoot (35.40s)
```

---

> [!NOTE]
> **High Risk**
> Changes guest PID 1 boot and mount namespace layout for all new instances; misbehavior would break guest startup, nested containers, or kernel-headers install.
> 
> **Overview**
> Guest init now **moves the image overlay onto `/`** (`switchRoot` via `MS_MOVE`, like `switch_root(8)`) in both exec and systemd modes instead of only `chroot(2)` into `/overlay/newroot`. That makes the container rootfs the mount namespace root so nested runc **`docker exec`** joins the right rootfs instead of the initrd.
> 
> The async **kernel-headers worker** is started with **`CLONE_NEWNS`** so initrd and pre-switch overlay paths still resolve after the root switch.
> 
> **Tests and CI:** shared helpers assert `/` is overlay in `mountinfo` and that headers install completes; **`TestNestedDockerExecRoot`** boots `docker:28.5.2-dind` and validates exec rootfs behavior (KVM-only, prewarmed with **`LinuxOnly`**). Systemd integration gains a **RootIsOverlay** subtest. Builder cache test waits up to **5 minutes** for `ReadyForBuilds`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 546f5022b36b27bcc337bd219dc76ada7980202b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chruffins (approved):
approved — reviewed at 6ed9128. reverted the unrelated builder readiness timeout relaxation in integration/builder_cache_test.go.

no correctness bugs found. remaining non-blocking findings:

- lib/system/init/headers.go:54-57 — the async headers worker maintains separate initrd and guest path contracts through a mount-namespace workaround.
- lib/system/init/mode_exec.go:31-35 and lib/system/init/mode_systemd.go:44-48 — root-switch and fatal-error handling are duplicated across modes.
- integration/nested_docker_test.go:50 — coverage exercises exec mode but does not directly assert systemd root switching or headers-worker namespace isolation.
- integration/nested_docker_test.go:62-113 — shared integration helpers and VM fixture setup are spread across unrelated test files.
- cmd/test-prewarm/main.go:49 and integration/nested_docker_test.go:21 — the Docker image reference is duplicated across packages.
- integration/nested_docker_test.go:111 — the Eventually failure message captures lastOutput before the polling closure runs, so it is always empty.
- lib/system/init/mode_systemd.go:17 and related docs/comments — several references still describe the old plain-chroot behavior.

open question: the headers-worker isolation relies on the initramfs mount tree remaining private. That is true for the current custom Alpine initrd path, but the invariant is implicit.

validation: gofmt and git diff --check passed; repository tests were not run. no blocking findings.

chruffins (approved):
correction to the prior approval: the builder readiness timeout change is intentional and is not a finding. the 5-minute timeout has been restored.

approved — no correctness bugs found. remaining non-blocking findings:

- lib/system/init/headers.go:54-57 — the async headers worker maintains separate initrd and guest path contracts through a mount-namespace workaround.
- lib/system/init/mode_exec.go:31-35 and lib/system/init/mode_systemd.go:44-48 — root-switch and fatal-error handling are duplicated across modes.
- integration/nested_docker_test.go:50 — coverage exercises exec mode but does not directly assert systemd root switching or headers-worker namespace isolation.
- integration/nested_docker_test.go:62-113 — shared integration helpers and VM fixture setup are spread across unrelated test files.
- cmd/test-prewarm/main.go:49 and integration/nested_docker_test.go:21 — the Docker image reference is duplicated across packages.
- integration/nested_docker_test.go:111 — the Eventually failure message captures lastOutput before the polling closure runs, so it is always empty.
- lib/system/init/mode_systemd.go:17 and related docs/comments — several references still describe the old plain-chroot behavior.

open question: the headers-worker isolation relies on the initramfs mount tree remaining private. that is true for the current custom alpine initrd path, but the invariant is implicit.

validation: gofmt and git diff --check passed; repository tests were not run.

rgarcia (approved):
approved — the root-switch implementation and nested Docker regression coverage look sound. a few non-blocking follow-ups:

### Question

- `integration/nested_docker_test.go:104` — could this also wait for `/run/hypeman/kernel-headers.status` to become `ready`? The new `CLONE_NEWNS` behavior otherwise lacks automated end-to-end coverage.

### Nits

- `lib/system/init/mode_systemd.go:17` — update “Uses chroot to switch” to describe the mount move plus chroot.
- `cmd/test-prewarm/main.go:49` — DIND is prewarmed on macOS even though this test always skips without `/dev/kvm`; consider making it Linux-only.

sjmiller609:
thanks — addressed in 546f502:
- nested docker test now waits for `/run/hypeman/kernel-headers.status` to reach `ready` and checks the headers dir and `build` symlink landed, so the headers-worker namespace path has end-to-end coverage
- dind prewarm entry is Linux-only (new `LinuxOnly` flag on `prewarmImage`)
- the `mode_systemd.go` comment was already fixed in bf07096

sjmiller609:
thanks both.

**chris** — went through your list:

- `Eventually` message capturing `lastOutput` early: real bug, fixed in `bf07096` (now `EventuallyWithT`, reports the last `docker info` output).
- stale chroot wording in `mode_systemd.go` / `network.go` / README: fixed in `bf07096` and `f7b49ab`.
- open question on the initramfs tree staying private: made explicit in `bf07096`. `switchRoot` now does `mount("", "/", MS_PRIVATE|MS_REC)` before the `MS_MOVE`, so it's enforced rather than assumed.
- headers-worker unshare, duplicated call sites, shared fixtures, image ref in two places: intentionally left. I prototyped the full version in #479 (single switch in `main.go`, both modes staging the headers tarball into the image and running the worker post-switch, shared test fixture). It came out at roughly +120/-140 non-test lines and changes guest-visible behavior in exec mode for no bug fixed, so I closed it. Happy to revisit as its own PR if we want the headers path unified.

**raf** — addressed in `546f502`:

- nested docker test now waits for `/run/hypeman/kernel-headers.status` to reach `ready` and checks the headers dir and `build` symlink landed, so the `CLONE_NEWNS` path has end-to-end coverage.
- dind prewarm entry is Linux-only via a `LinuxOnly` flag on `prewarmImage`.
- the `mode_systemd.go` comment was already fixed in `bf07096`.

both tests also assert the guest `/` is the image overlay in `/proc/self/mountinfo` now, in exec and systemd mode.

one thing I noticed while babysitting CI: four red runs on this branch were GitHub 500/504s downloading the guest kernel in `EnsureSystemFiles`. Separate PR, but a retry there would remove most of the flake we're seeing.

PR #481, opened 2026-09-15
Add deduplicated fsmerge rootfs support

chruffins:
## tldr

Adds the Hypeman-side implementation for deduplicated OCI layer -> rootfs storage. This uses the single device approach to EROFS's native sub-filesystem merging [shown here](https://erofs.docs.kernel.org/en/latest/merging.html). Hypeman assembles the metadata and shared layer artifacts into single read-only block devices using `dm-linear`. The existing flattened image path is now a fallback when fsmerge generation or host/platform support is unavailable.

This path requires a guest kernel configuration change to enable `CONFIG_EROFS_FS`.

## what changed

- Materializes OCI layers as content-addressed EROFS artifacts during image builds.
- Builds metadata-only fsmerge EROFS rootfs images for multi-layer Linux images.
- Repairs the external EROFS device-slot mappings and superblock checksum emitted by erofs-utils 1.8.10.
- Assembles the metadata image and layer artifacts into one shared read-only dm-linear device through read-only loop devices.
- Selects the assembled device in the common VM disk configuration path, covering Firecracker, QEMU, and Cloud Hypervisor on Linux.
- Ensures standby restore prepares the same image rootfs device before restoring the snapshot.
- Adopts existing image devices after restart and reconciles stale mappings; failed setup only cleans resources owned by that attempt.
- Keeps non-Linux platforms and unsupported hosts on the existing flattened-rootfs path.
- Persists the runtime rootfs type in the image manifest model.
- Serializes setup only per image digest, allowing unrelated image devices to initialize concurrently.

## guest-kernel dependency

The fsmerge runtime requires a separately released `kernel/linux` guest artifact with EROFS flattened-device support enabled. This PR contains the full Hypeman-side path but does not change the guest kernel repository. Until that artifact is deployed, fsmerge image execution must remain rollout-gated by the kernel release.

No Firecracker change is required: the assembled host device is passed as the existing read-only root drive.

## validation

- `GOCACHE=/tmp/hypeman-go-cache go test ./lib/images -run '^$'`
- `GOCACHE=/tmp/hypeman-go-cache go test ./lib/images -run 'Test(PatchFsmerge|ValidateDMName|ManifestModel)' -count=1`
- `GOCACHE=/tmp/hypeman-go-cache go vet ./lib/images`
- Separate Firecracker v1.14.2 validation on a development host confirmed the dm-linear layout boots, mounts EROFS, reads all layers, and passes whiteout/opaque-directory assertions.

## benchmark

Synthetic Firecracker benchmark using a 64 MiB shared base layer, two small unique layers, and ten image variants with a 1 MiB unique layer each:

| measure | flattened | fsmerge |
| --- | ---: | ---: |
| ten-variant storage | 681.6 MiB | 77.7 MiB |
| storage savings | — | 603.9 MiB (88.6%) |
| rootfs readiness median (10 runs) | 661.5 ms | 565.0 ms |
| rootfs readiness p95 | 1089 ms | 611 ms |
| direct 67 MiB read median (5 runs) | 224 ms | 116 ms |
| VMM RSS at readiness median (5 runs) | 4860 KiB | 4816 KiB |

Ten corrected fsmerge boots completed with 10/10 rootfs mounts. The benchmark shows no meaningful performance regression, but the sequential layout samples are not a controlled comparison and do not establish a speedup. Lazy dm setup on first VM creation, including read-only loop attachment and size discovery, was variable at roughly 1.3 s median; the resulting device is shared per image and reused by subsequent VMs.

The benchmark covers rootfs storage, lazy device assembly during VM creation, and guest mount readiness with a synthetic fixture. It does not measure a Chromium workload, the HTTP image-pull path, or snapshot/restore latency.


chruffins:
@cursor review

sjmiller609 (changes requested):
**summary:** There are a few concerns here to look into. Also keep in mind we don't have to go with this storage optimization at all if the complication of it is not worth the benefit.

the overall approach seems reasonable: share immutable layer files while keeping a reusable read-only disk for guest creation. Generally concerned on overhead for tons of loop devices and potentially-change-blocking guest disk IO performance.

### cleanup / GC
- `lib/images/dm_linear_linux.go:151–157` — successful loop detachments remain in `d.loops` after a partial failure. a later retry can detach a device number that has been reassigned. remove released resources from tracked state immediately.
- `lib/images/fsmerge_runtime.go:121–126` — cleanup treats an empty cache as no device. after a daemon restart, deleting an image before its next use leaves surviving mappings/loops behind.
- `lib/images/tag.go:79–96` — tag-replacement cleanup removes unreferenced content without closing its device. centralize reclamation across image-removal paths.
- `lib/images/fsmerge_runtime.go:110–132` — add startup reconciliation and periodic GC for owned mappings/orphan loops, including interrupted setup and manual directory deletion while stopped. use verifiable ownership, coordinate with create/restore, and retry busy devices without forcing removal. also reclaim unused materialized layers while preserving references from retained images, builds, guests, and snapshots.
- `lib/images/dm_linear_linux.go:60–65` — conditional hardening: failed setup calls `Close()` before owning the named mapping. separate managers racing on the same image could remove the winner’s device. clean up only resources this attempt owns.

### tests
- `lib/images/fsmerge_test.go:20–41` — the fixture occupies two blocks but declares one; expected offsets don’t match whole-file concatenation.
- add **one comprehensive real-tool integration test** covering layer contents, overwrites/whiteouts/opaque directories, assembly, restore preparation, and restart/cleanup recovery. make it deterministic, concurrent-run compatible, self-cleaning on failure, with bounded waits. **run it in Linux CI; missing prerequisites must fail, not skip.**

### potential performance issues
- `lib/images/fsmerge_runtime.go:40–41` — the global mutex serializes setup across unrelated images. consider per-image deduplication and bounded concurrency with safe loop allocation.
- `lib/images/dm_linear_linux.go:68–108` — validate device counts at 1,000 active unique images: exporter cardinality/scrape cost, enumeration, startup/fork latency, and read overhead versus flattened files.

### rollout note
guest-kernel compatibility is a deployment prerequisite, not a blocker assuming compatible kernel versions are enforced.

hiroTamada (commented):
reviewed at 9c75702b. the shared-layer approach makes sense, but i'd address the rollout and lifecycle issues before enabling it.

### rollout
- **P1 — `lib/images/manager.go:525–530`:** fsmerge is enabled automatically from host capabilities, while the PR requires a separately released compatible guest kernel. please add a default-off rollout/compatibility gate until that dependency is guaranteed. successful fsmerge generation retains no flat backup, so guest mount failures cannot use the build-time fallback.

### runtime correctness
- **P2 — `lib/images/fsmerge_runtime.go:190–210`:** device cleanup holds the global map mutex across external commands, including `dmsetup remove --retry`. a slow cleanup blocks cached root-disk lookups for unrelated images. reproduced with mocked commands. use per-digest lifecycle serialization and keep the global map lock short.
- **P2 — `lib/images/fsmerge_runtime.go:67–116`:** synchronous `singleflight.Do` ignores waiting callers' cancellation, while the initiating caller's context controls the shared setup. reproduced a waiter exceeding its deadline and returning success only after setup was released. use `DoChan`, per-caller cancellation, and a separately bounded shared context.
- **P2 — `lib/images/fsmerge_runtime.go:159–172`:** if mapper removal succeeds but loop detachment fails, the remaining loops are never retried by reconciliation because it only enumerates live mapper devices. reproduced with mocked commands. reconcile pending cleanup state as well, including failed setup cleanup.
- **P2 — `lib/images/fsmerge_runtime.go:22–31`:** logical deletion removes tags and files before attempting device cleanup. an in-use mapper then returns busy, making the API return 500 although the image is gone; lookup and delete retry return not found. reproduced through DeleteImage with mocked device commands. defer busy-device cleanup after successful logical deletion, or reject in-use deletion before mutations.

### smaller fixes and coverage
- `lib/images/fsmerge.go:111` — the table bounds check adds another 72 bytes after accounting for every full 128-byte slot. this rejects an in-bounds table ending at the image boundary. reproduced at helper level; reachability with real mkfs output not established. compare the actual table end against the image length.
- `lib/images/fsmerge_test.go:21–25,40–43` — the metadata fixture occupies two 4 KiB blocks but declares one, so expected layer offsets 1/4 disagree with the later concatenation's 2/5. fix the fixture and consider validating primary-device size against actual file size.
- please commit coverage for real fsmerge replacement/deletion/opaque-directory semantics, create/restore after host restart, legacy flat images, and the lifecycle failures above. current new tests primarily cover helper behavior.

### question
- `lib/images/dm_linear_linux.go:29–49` — should adopting an existing device verify its backing paths, table layout, and read-only state rather than only the number of loops? no exploit reproduced; this is a hardening question.

### validation
- targeted image-package tests and go vet ./lib/images passed.
- the image-package tests cross-compile for darwin/arm64 (compile only).
- temporary tests reproduced the findings above; no source changes were made.
- instance tests could not compile because required embedded hypervisor/guest-agent binaries were absent. no real fsmerge boot or snapshot/restore test was run.

